PR #447, opened 2013-08-05
Extended the AbstractInstrumentedFilter to support group aggregations on...

BGehrels:
... HTTP status codes.

When using Metrics for monitoring, you want to find situations, where your application is somehow misbehaving. The HTTP spec groups HTTP status codes in 5 Blocks: 1xx (Informational), 2xx (Success), 3xx (Redirection), 4xx (Client Error) and 5xx (Server Error). Having counts for each of these groups is pretty useful. A misbehaving application will probably throw one of the 5xx codes. A crawler running amok will probably trigger a lot of 4xx errors. if all those run into the "others" bucket, it gives me nearly no information.

 This commit changes AbstractInstrumentedFilters behaviour in the following way:
 There are now three Types of meters
- The user defined Meters for individual status codes
- The 5 group meters for the status code groups mentioned above
- The group named "other"
  
  Each request triggers either the group meters, or the "other" meter if there is an invalid status code. _In addition_, the user defined Meters will be triggered, if one of them matches the status code.

### DIFF
--- a/metrics-servlet/src/main/java/com/codahale/metrics/servlet/AbstractInstrumentedFilter.java
+++ b/metrics-servlet/src/main/java/com/codahale/metrics/servlet/AbstractInstrumentedFilter.java
@@ -27,6 +27,7 @@ public abstract class AbstractInstrumentedFilter implements Filter {
 
     // initialized after call of init method
     private ConcurrentMap<Integer, Meter> metersByStatusCode;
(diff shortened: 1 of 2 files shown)


BGehrels:
Wow, it's been over a year now, i would still love to see this in Metrics :-) Is there anything more i can do?

Issue #446, opened 2013-08-02
Meter with Configurable Windows


thobbs:
We're looking to use a Meter inside of Cassandra, but we really need EWMAs with time windows besides the 1, 5, and 15m defaults.  I would like to contribute a configurable version of the Meter class, but I want to get your opinion on the design first.

The Meter class is very hardcoded to use 1, 5 and 15m.  The Metered Interface even list specific methods for each of those.  I'm guessing you don't want to break that API, so a new class that does not implement Metered would be required.

The main points of the new interface would be:
- A constructor that takes a list (or vararg) of EWMA window sizes
- A `getRate(window)` method
- `mark()`, `getCount()`, and `getMeanRate()` would, of course, also be implemented

Thoughts?

PR #467, opened 2013-09-24
Add ConfigurableMeter Class

thobbs:
As mentioned in #446, the purpose of this is to add a version of the Meter class which allows custom EWMA windows to be specified.  The new ``ConfigurableMeter`` class does that and now ``Meter`` extends that class.

Because the ``Metered`` interface explicitly specifies 1, 5, and 15 minute rates, the API for this is a bit of an awkward fit.  If there are any changes I can make to resolve this, please let me know.

Thanks, and I look forward to feedback!


ryantenney:
Something like this will be supported in v4

Issue #515, opened 2014-01-16
SlidingTimeWindowReservoir getTick() can overflow and break


omerkudat:
The getTick() function of SlidingTimeWindowReservoir uses a collision buffer to allow for measurements which are happening within the same "tick". The default Clock tick is provided by System.nanoTime().

Under certain circumstances, nanoTime() returns a long sufficiently large that when multiplied by the collision buffer, it overflow into a negative number. The next line in the function will then check whether the new tick (nanos \* buffer) is larger than the old and, if not, it will return oldTick + 1. Because of this, the getTick() function returns ticks which are always incresing by one, therefore distorting the window.

To fix this, I added an absolute offset value to the class. The offset value can be a final field which is initialized when the class is constructured or a static. Following each call to Clock.getTick(), we would reduce the return value but this absolute amount.

ryantenney:
@victor-prp I've just committed a possible fix, as well as your test: d49489663c43aa29593a363a08d2dbb7dd2e3582, however the test now times out, instead of resulting in an assertion error.

martinsmatthews:
We've seen exactly this in production, using > 3GB memory when we get an overflow (there are lots of metrics in there ;)). 

@ryantenney agree with your logic, I've tested your fix and it looks good to me - test case is here https://gist.github.com/martinsmatthews/c2f87b8ec90652b7e3a1

Let me know what you think.

ryantenney:
@martinsmatthews Great test, thanks! I'll get that committed to the branch and merge it in.

martinsmatthews:
Cool, we are going to monkey patch this in our production environments, but would be good to get a bona fide release at some point!

mozotom:
Here's what I've done at my place that fixed it https://github.com/dropwizard/metrics/pull/931
We had this problem when physical server uptime was over 417 days - rebooting server, would restart up time for another 417 days, but we wanted a permanent fix, so I created this one.

mozotom:
My original solution got shut down, so here's alternative: https://github.com/dropwizard/metrics/pull/932 If you want to keep your service running longer without OOM, consider smaller collisionBitwidth (like 4 instead of default 8 will give you 16 years, not one)

Zvezdochet:
[...]

PR #523, opened 2014-02-12
Add offset to Clock.UserTimeClock against overflow

omerkudat:
UserTimeClock returns elapsed time from an arbitrary starting time, not
epoch. This means value could be negative or arbitrarily large (not in
line with System.currentTimeMillis.) Other metrics classes, such as
SlidingTimeWindowReservoir multiply the Clock.getTicks with some value
to create a collision buffer. If nanoTime is large enough, this can
overflow and break the window.

To work around this, this commit modifies Clock.UserTimeClock to return
elapsed time since the UserTimeClock class was classloaded. It is still
elapsed time and increasing, but it will be a smaller value and therefore
less likely to overflow. Unless the JVM has a really long uptime. Hmm...

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/Clock.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Clock.java
@@ -37,12 +37,17 @@ public abstract class Clock {
     }
 
     /**
-     * A clock implementation which returns the current time in epoch nanoseconds.
+     * A clock implementation which returns the elapsed time in nanoseconds.
      */
     public static class UserTimeClock extends Clock {
+        // System.nanoTime can be arbitrary, so offset from a starting point
+        // so the values are not very large (or negative.) All we want is
+        // elapsed time.
+        private static final long OFFSET = System.nanoTime();
+
         @Override
         public long getTick() {
-            return System.nanoTime();
+            return System.nanoTime() - OFFSET;
         }
     }
 
--- a/metrics-core/src/test/java/com/codahale/metrics/ClockTest.java
+++ b/metrics-core/src/test/java/com/codahale/metrics/ClockTest.java
@@ -29,9 +29,13 @@ public class ClockTest {
                 .isEqualTo(System.currentTimeMillis(),
                            offset(100.0));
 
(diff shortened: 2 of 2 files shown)


ryantenney:
This isn't guaranteed to fix the issue, I've reopened #515.

omerkudat:
I don't know about the leak but this patch does fix the overflow issue. What do you think is missing?

ryantenney:
It is possible that the class is loaded when `long ns = System.nanoTime(); ns >= 0 && ns < Long.MAX_VALUE / COLLISION_BUFFER` thus the overflow behavior is entirely unchanged. Also, javadocs advise us that nanoTime **may** be negative, causing further weirdness, including but possibly not limited to underflow.

omerkudat:
I don't quite get the class loading issue. However, the negative nanos shouldn't be a problem as long as the difference is used; values from subsequent invocations will always be larger numbers.

ryantenney:
Having thought through it, the answer, if we're going to do `getTick() * 256` in SlidingTimeWindowReservoir, is to change all instance of that to (something like) `(getTick() % (Long.MAX_VALUE / COLLISION_BUFFER)) * COLLISION_BUFFER`.

augustocallejas:
@ryantenney - Can you explain your comment to @omerkudat regarding:
```long ns = System.nanoTime(); ns >= 0 && ns < Long.MAX_VALUE / ((COLLISION_BUFFER - 1) / COLLISION_BUFFER)```

Providing only offsets from an initial value ```System.nanoTime()```obtained when the ```UserTimeClock``` class is loaded seems okay, although I do agree that overflow could eventually happen over a long period of time when multiplying by ```COLLISION_BUFFER```.

martinsmatthews:
This will still cause an overflow after ~417 days uptime by my reckoning

mozotom:
Here's what I've done at my place that fixed it https://github.com/dropwizard/metrics/pull/931
We had this problem when physical server uptime was over 417 days - rebooting server, would restart up time for another 417 days, but we wanted a permanent fix, so I created this one.

PR #534, opened 2014-03-06
New Metric: Sliding-window, Increment-only counter (aka Spiral)

sschwell:
I've found this Metric to be extremely useful. It was inspired by <a>https://github.com/boundary/folsom#spiral-meter</a>


ryantenney:
I haven't seen any support for this, so I'm afraid I can't merge it.  User defined metric types should be supported by v4.

PR #561, opened 2014-04-08
Introduce MetricName type to encapsulate more information about metrics.

udoprog:
**The point of this pull request** is not to have it merged. I appreciate the fact that it is a big undertaking to incorporate something like this (third party libraries especially will be a blast).

I am hoping to start the discussion, and I see no better way of doing it that through code.

##### Commit message
This is a large refactoring meant to change the way metric names are
represented to better suit a wide variety of backends.

Quite a few metric backend systems are starting to use a general concept of
'tags' which is a step back from having too encode information in the (string)
names of the metrics*.

\* KairosDB (http://code.google.com/p/kairosdb/), OpenTSDB
(http://opentsdb.net/overview.html)

Tags are meant to allow for natural grouping of similar timeseries allowing for
the ability to easily correlate freely between the timeseries that incorporate
them.

For example, take the following two timeseries.

* example.com.**interface.traffic**.eth0.in
* example.com.**interface.traffic**.eth0.out

Using tags, these would be represented as;

* **interface.traffic** {host=example.com, interface=eth0, direction=in}
* **interface.traffic** {host=example.com, interface=eth0, direction=out}

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
@@ -165,18 +165,18 @@ public class ConsoleReporter extends ScheduledReporter {
     }
 
     @Override
-    public void report(SortedMap<String, Gauge> gauges,
-                       SortedMap<String, Counter> counters,
-                       SortedMap<String, Histogram> histograms,
-                       SortedMap<String, Meter> meters,
-                       SortedMap<String, Timer> timers) {
+    public void report(SortedMap<MetricName, Gauge> gauges,
+                       SortedMap<MetricName, Counter> counters,
+                       SortedMap<MetricName, Histogram> histograms,
+                       SortedMap<MetricName, Meter> meters,
+                       SortedMap<MetricName, Timer> timers) {
         final String dateTime = dateFormat.format(new Date(clock.getTime()));
         printWithBanner(dateTime, '=');
         output.println();
 
         if (!gauges.isEmpty()) {
             printWithBanner("-- Gauges", '-');
-            for (Map.Entry<String, Gauge> entry : gauges.entrySet()) {
+            for (Map.Entry<MetricName, Gauge> entry : gauges.entrySet()) {
                 output.println(entry.getKey());
                 printGauge(entry);
             }
@@ -185,7 +185,7 @@ public class ConsoleReporter extends ScheduledReporter {
 
         if (!counters.isEmpty()) {
             printWithBanner("-- Counters", '-');
-            for (Map.Entry<String, Counter> entry : counters.entrySet()) {
+            for (Map.Entry<MetricName, Counter> entry : counters.entrySet()) {
                 output.println(entry.getKey());
                 printCounter(entry);
             }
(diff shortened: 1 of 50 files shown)


salilsurendran:
In this particular PR the name has been replaced with a class MetricName. Instead of this wouldn't it be simpler to have tags supplied when one creates a counter, gauge or timer etc via the MetricsRegistry.counter, .gauge() etc.

ryantenney:
I'm confident that this MetricName approach will be the most flexible and user friendly way to implement tags.

salilsurendran:
Could you please let me know why a MetricName would be friendly way to implement tags? Suppose if I create a counter like this:

MetricRegistry.counter("name", new String[]{"tags"});
counter.increment();

Are there any underlying functionality that would prevent an approach I mentioned above?

ryantenney:
There's nothing that would prevent us from doing that. This is simply a better API, in my opinion:

```java
MetricName counterName = new MetricName("counter").tagged("tag1", "value1");
MetricRegistry.counter(counterName)  // counter { tag1=value1 }
MetricRegistry.counter(counterName.tagged("tag2", "value2"));  // counter { tag1=value1, tag2=value2 }
```

salilsurendran:
But doing this results in 50 file changes in your PR. Whereas if you go with what I suggest you will only have to change the MetricRegistory methods to accept tags. I was wondering if there is a reason related to functionality that requires the use of MetricName?

ryantenney:
Committed to branch. Thanks @udoprog!

udoprog:
@ryantenney Wow, must admit I was not expecting this. kudos for looking at it again!

At my company we've implemented something similar (but we went with the name MetricId), what we ended up doing was replace MetricRegistry and start from scratch. We'll be releasing the library pretty soon.

These are a few things we've learned along the way;

* Most of the existing MetricSet's did not structure the things that we wanted them to, so we started from scratch and re-implemented the ones we wanted with tags in mind. Strong conventions are a must (e.g. every metric must have a distinct ```key``` and a ```unit``` tag that makes sense).
* Immutability of MetricId is really neat, but expensive, copying the underlying Map degrades performance in the scenarios where you want to have a dynamic field and lots of measurements. We had to implement a ```MetricIdCache<K, MetricId>``` to cope with this.
* It is very beneficial for ```MetricName``` and ```MetricRegistry``` to be interfaces that are distributred separately from the core, it makes it easier for frameworks to integrate with metrics. It also makes it easier to version everything (new packages when breaking backwards compatibility) and providing shims for frameworks which are slow adopters.

@salilsurendran
Replacing the identifier of a time series with a concrete type has a few benefits.

Refactoring probably does not have to happen again. E.g. the behavior of a MetricName and its underlying implementation can change freely without having to modify signatures and method invocations all over the place.

Concretely about what you suggest, if you encapsulate the tags of a metric in the API, every method that interacts with metrics would have to support the _new_ way of doing things, and in order to make use of it, all these 50 files would have to be modified at one point or another.

String's are in my opinion poor metric identifiers, they require encoding to encapsulate non-trivial information (like tags). And they have very strong pre-defined behaviour that can not (should not) be extended or modified (final). Two things i'd want for a metric identifier in a framework.
- - -
I hope any of this will be useful, I'd be thrilled to help out on this if possible.

davidkarlsen:
Really looking forward to this - what is the timeframe for a v4 release?

PR #617, opened 2014-07-30
Allow to configure format of Double and Float on GraphiteReporter

nakedpony:
Doesn't brake anything. By default uses old format.


nakedpony:
up :)

arteam:
It should be now implemented in the form of #1073.

PR #652, opened 2014-09-08
Intercept HttpServletResponseWrapper#setStatus(int, String) to fix response status tracking

noslowerdna:
We are using version 2.2.0 of the metrics library and noticed that the WebappMetricsFilter (renamed in v3 to AbstractInstrumentedFilter) was recording a high number of 'responseCodes.other' responses for our services, and 0 for most other response types, include 'responseCodes.ok'. After a little debugging, found that this was due to HttpServletResponseWrapper#setStatus(int, String) being called to set the response status, and this method not getting intercepted in the WebappMetricFilter's StatusExposingServletResponse. In metrics-web 2.2.0 the StatusExposingServletResponse's httpStatus variable is not initialized (that was changed in v3 to default to 200), so the response status appears to the filter to be 0, associating it to the 'other' meter.

For what it's worth, here are a few examples of how we are setting the status in our service code.
- return Response.ok(builder.toString()).build();
- return Response.status(Status.NOT_FOUND).build();
- return Response.status(Status.SERVICE_UNAVAILABLE.getStatusCode()).entity(builder.toString()).build();

### DIFF
--- a/metrics-servlet/src/main/java/com/codahale/metrics/servlet/AbstractInstrumentedFilter.java
+++ b/metrics-servlet/src/main/java/com/codahale/metrics/servlet/AbstractInstrumentedFilter.java
@@ -144,6 +144,12 @@ public abstract class AbstractInstrumentedFilter implements Filter {
             super.setStatus(sc);
         }
 
+        @Override
+        public void setStatus(int sc, String sm) {
+            httpStatus = sc;
+            super.setStatus(sc, sm);
+        }
+
         public int getStatus() {
             return httpStatus;
         }


mcucchiara:
Unfortunately we're using 2.2.0 and, at this stage, migrating to 3.x is not an easy option for us.

Would it be ok applying the same change to 2.2.x and release a patched version?

Please let me know, happy to submit a PR

ryantenney:
I'd accept the pull request, but I wouldn't be able to release it to maven. It's no longer possible to push a new release to the com.yammer.metrics groupId, and it wouldn't make sense to release it under the new groupId.

noslowerdna:
@mcucchiara For 2.2.x we simply forked the filter class into our code with the bug fix, configuring our service to use that copy instead. [Link](https://gist.github.com/noslowerdna/6572c1113a1c770c95a7) to sample code.

PR #654, opened 2014-09-10
Override toString in Metric implementations

### LINKED ISSUE #566: Override `toString()` in Meters, Histogram, Timer...
For debugging in the IDE and other things it would be cool if all Metered classes would overwrite the `toString()` method. Currently they use just the uninformative `toString()` from `Object`.

srvaroa:
Keeping the basic output for Counter and Clock in the same line as the existing RatioGauge::toString, but for Histogram etc. tries a human friendly yet parseable output. Happy to amend to better formats anyway.

Fixes #566

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/Histogram.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/Histogram.java
@@ -1,5 +1,10 @@
 package io.dropwizard.metrics;
 
+import java.io.ByteArrayOutputStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
 /**
  * A metric which calculates the distribution of a value.
  *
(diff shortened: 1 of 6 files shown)


srvaroa:
@ryankennedy @jplock I rebased this, let me know if it needs any changes.

ryankennedy:
Guessing you meant @ryantenney. 😄

srvaroa:
Oops sorry, naughty autocomplete! :)

metrics-core/src/main/java/io/dropwizard/metrics/Histogram.java
```diff
@@ -53,4 +58,15 @@ public long getCount() {
     public Snapshot getSnapshot() {
         return reservoir.getSnapshot();
     }
+
+    @Override
+    public String toString() {
+        final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        this.getSnapshot().dump(out);
+        try {
+            return out.toString(Charset.defaultCharset().name());
```
jplock: should we just use `StandardCharsets.UTF_8` here to avoid having to deal with `UnsupportedEncodingException`?

srvaroa:
Thx for taking a look @jplock, done.

ryantenney:
Thanks!

Issue #178, opened 2012-03-02
Allow for fine-grained filtering of metrics


codahale:
Currently reporters' outputs can be filtered by `MetricName`, but it would be preferable to have more fine-grained filters than that, ideally down to the measurement level.

Pull request #91 by @cburroughs attempted to address this, but it's still an open issue.

ryantenney:
Lots of demand for this, will aim to get this in v4.

ryantenney:
Going to follow the approach outlined in #837

bentatham:
Great!  So I'll throw away the PR #836 in favour of this approach?

ryantenney:
@bentatham Sounds good.

bentatham:
Is this still planned for 4.0.0?  My idea in #837 is targeted for 5.0.0.  Would love to have something in place.  

I'm happy to attempt a PR, if I have hope it will be looked at for 4.0.0.

PR #659, opened 2014-09-12
Added support for filtering metric values out of GraphiteReporter

jlw4:
I implemented these changes to avoid flooding our graphite server with too many data points.

The other reporters could use the MetricValueFilter as well, if they want to reduce the volume of data published.


ryantenney:
This has become an extremely common request, and I've reopened #178 to track it. This is now on the roadmap for v4 as the proper way to do it will introduce breaking changes. The upshot is that each individual reporter will not need to support sub-metric filtering.

arteam:
Implemented as #1048

PR #685, opened 2014-10-10
684 Add SocketReporter to send metrics to Socket

Sree2k:
Adding socketReporter allows the metrics to be sent over socket for consumption which is commonly used scenario in Apache Chukwa. We need this functionality to integrate with chukwa.


ryantenney:
I'm afraid I haven't seen any other requests for this functionality, so I'm unable to merge it.  I would strongly recommend you separate this module into its own project, publish it to Maven Central, and add a link to the project in the file `third-party.rst`.

PR #695, opened 2014-10-24
add some logs when Graphite Reporter is in error

twillouer:


twillouer:
any idea if this can interess you ?

PR #696, opened 2014-10-24
remove Error if reportError = false

twillouer:


ryantenney:
Disabling the logger for this class would be the preferred method to suppress log messages.

PR #699, opened 2014-10-28
Add getLong and getDouble to JmxGaugeBean. Add logging of exceptions to ScheduledReporter

taer:
Some JMX visualizers(like jmc) will only plot primitives on their
dashboards. This adds 2 primitive fields to the JmxGauge that will unbox
the value if it's a boxed primitive.

For Scheduled reporter, log exceptions and do not propagate.


taer:
Closing this. Submitting JMX change against master

PR #700, opened 2014-11-03
Add size field to the histogram json representation

jirkah:
Having the size field included in the json is interesting for some use cases. Not sure if it would be interesting for Timers. Thanks for your comments.


ryantenney:
Size isn't reported by other reporters, and I haven't seen any other requests for this functionality, so I'm afraid I'm unable to merge it.

PR #714, opened 2014-12-03
Add support for Pickled Graphite over UDP

philwill-nap:
Pulled the logic to create the Pickled data format into its own class (GraphitePickler).
Adjusted PickledGraphite to use the GraphitePickler class.
Created PickledGraphiteUDP class which is to PickledGraphite what GraphiteUDP is to Graphite.


arteam:
Superseded by #1035.

PR #715, opened 2014-12-04
Fix for the monitoring jetty9 async servlet requests.

tempredirect:
Async requests are monitored via the addition of an async listener.
The bug was that `onStartAsync(AsyncEvent)` is called when the handler
invokes `startAsync()` but the `InstrumentedHandler` only added the
listener after the call to the delegate handler, thus would miss
the all onStartAsync event.

The net result of the bug was that the start time of all async requests
was reported as 0 so the calculated elapsed times where then rather
large (i.e. currentTimeMillis()).

### DIFF
--- a/metrics-jetty9/src/main/java/io/dropwizard/metrics/jetty9/InstrumentedHandler.java
+++ b/metrics-jetty9/src/main/java/io/dropwizard/metrics/jetty9/InstrumentedHandler.java
@@ -230,6 +230,7 @@ public class InstrumentedHandler extends HandlerWrapper {
             // new request
             activeRequests.inc();
             start = request.getTimeStamp();
+            state.addListener(listener);
         } else {
             // resumed request
             start = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


tempredirect:
Any chance somebody can have a quick look at this please?

tempredirect:
Have rebased this request against master.

tempredirect:
Hi, anybody fancy having a look at this??

ryantenney:
@tempredirect I'm afraid I'm not terribly knowledgable about Jetty. I've been hoping that someone who is more knowledgable would come along and sign off on this PR!

uwemaurer:
I can confirm that InstrumentedHandler is currently not working with async servlets and this change fixes it.
I suggest to merge this fix.

This will also solve #914

jonathanmgrimm:
Can also confirm this fixes the same problem as others are seeing

mziemba:
I have the same issue, are there any plans on merging this PR?

ryantenney:
@jonathanmgrimm have you had a chance to look into this any further? I was going to merge it, but your comment concerns me.

ryantenney:
Merging to 3.2 in #1074

PR #725, opened 2014-12-18
Added possibility to define a custom size of snapshot

rzukow:
In our project we need a bit more precise quantiles (like p9999) but this is not possible without increasing size of snaphot.
This PR introduces a possibility to create MetricRegistry object with custom size of created metrics.


ryantenney:
I intend to include support for customization of metrics, but it's going to be a bigger change than this. As far as specifying size on the reservoir, that's already possible with the `public ExponentiallyDecayingReservoir(int size, double alpha)` constructor.

rzukow:
Is it possible to specify size of reservoir when creating MetricRegistry?
I think the problem here is that the only constructor used by MetricRegistry is the default one, and it doesn't really matter that there is ExponentiallyDecayingReservoir(int size, double alpha) constructor.

ryantenney:
Only if you create the timer manually: new Timer(new EzponentiallyDecayingReservoir(...)) and then register it with the metric registry.

V4 will permit you to set a default metric factory which allow you to customize the default behavior.

rzukow:
Ok, should be enough for  now. Thanks.

Issue #637, opened 2014-08-31
Graphite metrics should clean metrics names


JensRantil:
Graphite does not allow metrics/sensors with names containing `'`, but Metrics happily submits them. This makes Graphite log a bunch of errors not being able to create the whisper files yada yada.

I see three different solutions to this:
 * Patch Metrics to not submit these metrics if they have the wrong naming, logging the metrics with the bad naming.
 * Automagically clean the sensor names (removing weird characters etc.) on submission to Graphite/carbon.
 * Keep things the way they are. Possibly adding a static helper method that can be used to clean sensor names from garbage. I am currently wrapping a bunch of sensor names in:

```java
public class MetricsUtils {

    /**
     * Clean graphite metrics names.
     * <p>
     * This was created because I was seeing a lot of stacktraces in Carbon log due to broken metric names.
     * 
     * @param proposal
     *            the proposed metrics' name.
     * @return cleaned metric's name
     */
    public static String cleanMetricName(String proposal) {
        return proposal.replace("'", "");
    }

}
```

Input welcome.

JensRantil:
I just ran into a bug(?) triggered in Graphite when a metric had a parenthesis in it. Filename was `/whisper_root/package/path(something)/1_mean.wsp`. And I was plotting `aliasByNode(package.*.1_mean, 1)`. We are now internally stripping all parenthesis.

felixbarny:
Graphite only supports alphanumeric characters + !#$%&"'*+-.:;<=>?@[\]^_`|~

JensRantil:
@felixbarny Thanks! Do you have a reference for this list? As stated in the initial description, `'` is broke my Graphite.

felixbarny:
I've found it somewhere in Graphite's source code. Can't remember where though...

ryantenney:
I've gone with the following regex: [^a-z0-9!#\$%&"'\*\+\-.:;<=>\?@\[\]\^_`\|~]+

If there are no objections then I will be releasing 3.1.1 today.

felixbarny:
I think the regex is missing the backslash character (`\\`)

ryantenney:
So it is. Thanks!

ryantenney:
Forward slash is legal then?

felixbarny:
No, that's one of the most illegal chars ;)
[...]

PR #731, opened 2014-12-31
Sanitize names sent to graphite

thattolleyguy:
Enhancing the sanitizer to remove other characters that aren't allowed. Any illegal character is replaced with a dash. @ryantenney

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
@@ -13,7 +13,8 @@ import java.util.regex.Pattern;
  * A client to a Carbon server via TCP.
  */
 public class Graphite implements GraphiteSender {
-    private static final Pattern WHITESPACE = Pattern.compile("[\\s]+");
+    public static final Pattern DISALLOWED_CHARS = Pattern.compile("[^a-zA-Z0-9!#\\$%&\"'\\*\\+\\-:;<=>\\?@\\[\\\\\\]\\^_`\\|~.]");
+
     // this may be optimistic about Carbon/Graphite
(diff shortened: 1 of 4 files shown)


nickbabcock:
Why the trailing `.` at the end of the regex, which makes it different than the regex in https://github.com/dropwizard/metrics/issues/637#issuecomment-63964756? Your solution won't change `foo@bar.com` (the period is significant to graphite!). Dropping the period from the regex will change the previous example to `foo@bar-com`

nickbabcock:
With #936 merged, should this be closed?

Grundlefleck:
Hi, 

I just came across this issue (happened to try 3.2.2 and 4.0.2) and it has not been addressed by #936. At first glance it looks like extending the disallowed characters, from only '-' to also exclude apostrophes would have avoided the issue. 

How this manifested for me is that from the metrics-jvm module, using Java 9, the MemoryUsageGaugeSet emits metrics prefixed with: 
 - `jvm.memory.pools.CodeHeap-'non-nmethods'.`
 - `jvm.memory.pools.CodeHeap-'non-profiled-nmethods'.`
 - `jvm.memory.pools.CodeHeap-'profiled-nmethods'.`

I have a workaround for this particular issue[0] but a generic replacement of any invalid characters would have been great.

Is there any chance of getting this PR revived, or should I submit something similar based off more recent code?

---
[0] these are the only offending metrics I've noticed out of the modules we use, and I have chosen to drop the metrics like so:
```
registry.register("jvm.memory", new MemoryUsageGaugeSet(
  ManagementFactory.getMemoryMXBean(),
  ManagementFactory.getMemoryPoolMXBeans().stream().filter(bean -> !bean.getName().contains("'")).collect(toList())));
```

maxmil:
Referencing the same issue https://github.com/dropwizard/metrics/issues/1322

PR #733, opened 2015-01-08
Introduce a backend for Instrumental

bvarner:
[Instrumental](http://instrumentalapp.com) is a cloud-based metric graphing / dashboard type service that uses a protocol similar to graphite, but with a few changes (handshake, gauge semantics, etc.)

This PR includes full tests cases to make sure the handshake works, and the core function is largely based (but different from) the graphite reporter.

I've also included a change exposing a private method as public for registering MetricSets with string prefixes... this seemed like a really natrual thing to do, especially since not having MetricSets with prefixes was getting very unwieldy in the Instrumental front-end.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -381,7 +381,7 @@ public class MetricRegistry implements MetricSet {
         }
     }
 
-    private void registerAll(String prefix, MetricSet metrics) throws IllegalArgumentException {
+    public void registerAll(String prefix, MetricSet metrics) throws IllegalArgumentException {
         for (Map.Entry<String, Metric> entry : metrics.getMetrics().entrySet()) {
             if (entry.getValue() instanceof MetricSet) {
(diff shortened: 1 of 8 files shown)


ryantenney:
Hi Bryan, I'm sorry to say that we can't accept new modules into this project at this time.  It has become difficult to maintain a large collection of modules, many of which we didn't author.  I'd strongly encourage you to publish this as an independent project, and submit a pull request to https://github.com/dropwizard/metrics/blob/master/docs/source/manual/third-party.rst with a link to your project.

As for changing the visibility on `void registerAll(String name, MetricSet metricSet)`, as `MetricSet` implements `Metric` you can pass your MetricSet to `<T extends Metric> T register(String name, T metric)`, which first checks if the metric is a `MetricSet` and calls `registerAll` (see: https://github.com/egineering-llc/metrics/blob/69fc970151f509a0dc6ccf34ef7c7e9beed663aa/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L84).  This is a frequent source of confusion, so I have opened issue #734 to improve the documentation.

bvarner:
Thanks, i missed the instanceof. :-)

PR #736, opened 2015-01-09
Deprecate buildMap function from MetricRegistry

mehrdad-hassanabadi:
This is due to following the wrong pattern of calling an overridable method
  from within the constructor of the superclass.
  Read more here:
  http://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors


ryantenney:
Thanks!

PR #737, opened 2015-01-09
Added a reference to the metrics-instrumental 3rd party lib.

bvarner:
As suggested by ryantenney in https://github.com/dropwizard/metrics/pull/733,

We've pushed our metrics-instrumental project to maven central, and here's a PR regarding an update to the third-party documentation.

Thanks!


skamille:
+1 lgtm

PR #738, opened 2015-01-18
Including Ekstazi (ekstazi.org) profile to optimize execution of the tests

andzaytsev:
Ekstazi allows to speed up regression testing by only executing tests that were affected by the changes that were made since the last test run. To run the tests with Ekstazi with my patch to pom.xml, one should use ```mvn test -Pekstazi```.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,35 @@
                 </plugins>
             </build>
         </profile>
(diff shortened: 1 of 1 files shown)


ryantenney:
Thanks for the patch, but the test duration isn't a concern for me at this time.

PR #744, opened 2015-01-28
Fix InstrumentedHandler timings for asynchronous requests

rossabaker:
In asynchronous contexts, all request times are being calculated against start time 0.
1. startTime is never initialized, because the listener is installed after the async context has started.
2. startTime makes the listener not thread safe, but it is shared by all requests.

The timings are correct after reverting 92f7d920deac3c0065ae504784a3b271d6b46fe7.


rossabaker:
The spirit of the reverted commit could be preserved by passing the start time to a listener created on each suspended request.  Since this handler is already coupled to Jetty, I don't see the point, and it's a little bit lighter weight with the shared listener.

PR #752, opened 2015-02-20
Ping response without new line

vilhelmpersson:
When testing it is easier to test for ping only.
It is not obvious that there is an newline in the response when testing in the browser before setting up the surveillance using the ping.
Removed the new line from the ping response.


ryantenney:
I have concerns about this breaking the /ping endpoint for existing users, so I'm going to leave it as is.

PR #755, opened 2015-02-22
Add getLong and getDouble to JmxGaugeBean

taer:
Some JMX visualizers(like jmc) will only plot primitives on their
dashboards. This adds 2 primitive fields to the JmxGauge that will return
the primitive if it's a boxed primitive.


taer:
Build fail seems to be transient. Only 1 of the 2 failed

PR #760, opened 2015-03-02
Added a concurrent instances count to Timer and with max concurrent, along with updating Reporters.

Jukkorsis:
Timer now collects how many concurrent events of a Timer there are. When collecting metrics on a HTTP call or DB operation, it is often useful to know if calls are being made in parallel/concurrently, along with the rate and latency. For example, knowing the concurrency gives you insight into how thread or connection pools are being using.

Fixed Reporter tests.


marshallpierce:
Seems like this is a separate concept from a Timer, no? It would be nice to have a way to not incur this overhead for users of Timer that don't need this info. It's more analogous to a Counter anyway.

jmason:
Agreed.  The pattern I tend to use for this is an AtomicInteger variable to measure concurrency, recording into a Histogram -- this gives min/max/mean and, importantly, percentiles.

Jukkorsis:
I originally considered creating a new Timer-like class that captured concurrence metrics, either like the Histogram or using a Histogram. Then users could use the simple/current Timer or a new advanced Timer. However I turn away from that idea for a few reasons:

1. I could not come up with a simple name for such a class
2. I did not think the library in general needed a Timer class and an "AdvanceTimer" class
3. A lot of what a Histogram offers I did not really need

So I decided that simple was better and my solution does not add much overhead.

I did consider not having the "max concurrent" metric, as I can get that information from Graphite the metrics repository I use. So I would consider removing "max concurrent" to reduce the "overhead" if that was a issue.

ryantenney:
Agreeing with @marshallpierce on this, it is a separate concept from a timer.

PR #762, opened 2015-03-12
Add /package_version admin servlet.

shtratos:
This servlet outputs jar package version as plain text.
It can be useful to find out which package version is actually running.
The version is derived from jar package Manifest file via built-in Java mechanism.


ryantenney:
Forgive me if I'm mistake, but won't this simply return the version of the Metrics jar?

shtratos:
Common pattern with Dropwizard services is to merge all jars in one über-jar via maven-shade plugin. In that case the endpoint will return version of that jar.

ryantenney:
Since the usefulness of this is predicated on having an uberjar, I can't merge it.  Dropwizard might be open to doing something like this however.

Issue #764, opened 2015-03-15
Make TICK_INTERVAL configurable in Meter


doom369:
Hi there,

I decided to use metrics in order to restrict user message request limit. But currently Meter implementation is uses hardcoded value of 5 seconds for rate update. 
```
private static final long TICK_INTERVAL = TimeUnit.SECONDS.toNanos(5);
```
So user could easly do as many request as he wants within this 5 seconds. I propose to add one more constructor for that.
```
    public Meter(int seconds) {
        this.clock = Clock.defaultClock();
        this.startTime = this.clock.getTick();
        this.lastTick = new AtomicLong(startTime);
        this.TICK_INTERVAL = TimeUnit.SECONDS.toNanos(seconds);
    }
```

What do you think?

ryantenney:
I'm afraid I don't quite understand the application. Could you please elaborate?

PR #765, opened 2015-03-15
Make TICK_INTERVAL configurable in Meter via constructors

doom369:
Pull request for [Make TICK_INTERVAL configurable in Meter via constructors] (https://github.com/dropwizard/metrics/issues/764).


ryantenney:
Changing the tick interval on a meter doesn't make sense, since the 1, 5 & 15 minute exponentially weighted moving averages expect a 5 second tick.

PR #768, opened 2015-03-17
Change Graphite client to use isConnected()

### LINKED ISSUE #767: Graphite instance of GraphiteSender doesn't reconnect
On disconnect from carbon server the Graphite client doesn't reconnect. This appears to be due to the way connect() does it s checks. Instead of checking isConnected() it checks for socket != null. However on disconnected socket is not set to null. 

This does not effect the PickelGraphite client as it uses the isConnected() function instead.

nerdynick:
Resolves #767

On disconnect from carbon server the Graphite client doesn't reconnect. This appears to be due to the way connect() does it s checks. Instead of checking isConnected() it checks for socket != null. However on disconnected socket is not set to null. 

This does not effect the PickelGraphite client as it uses the isConnected() function instead.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
@@ -104,7 +104,7 @@ public class Graphite implements GraphiteSender {
 
     @Override
     public void connect() throws IllegalStateException, IOException {
-        if (socket != null) {
+        if (isConnected()) {
             throw new IllegalStateException("Already connected");
         }
         InetSocketAddress address = this.address;


ryantenney:
Thanks for bringing this to my attention. A test started failing due to this change, so I made this change and updated the tests. Resolved in 70559816f1fc3a0a0122b5263d5478ff07396991 and dd0fca2f4d89497ee57284410c2db0c72a63296a.

adrianmkng:
Any plans to release this fix to maven central? We're experiencing the problem as well.

ryantenney:
I just pushed the 3.1.2 release to Maven Central

PR #770, opened 2015-03-26
Simplify slf4j reporting when logging is disabled

schlosna:
Avoid overhead when logging is disabled for the specified reporting log level or marker.


ryantenney:
Thanks! Also pulled this into 3.1-maintenance.

schlosna:
Thanks for the quick review and merge @ryantenney!

PR #777, opened 2015-04-02
Added hourly and daily metrics

zdanek:
I've created hourly and daily metrics. This pull request is for 4.0.0 (master) branch.


ryantenney:
Support for customizing meters will make it into v4, but I can't force hourly and daily on our users, as I'm already getting lots of complaints about reporting too many metrics.

zdanek:
My client will kill me ;)
When will v4 be available?

Issue #778, opened 2015-04-08
metrics-httpasyncclient only captures submission time


derekbassett:
The current implementation of metrics-httpasyncclient only captures submit time, not very helpful for a usage perspective since submission time is very quick.  It would be more handy if we could see the current active requests in metrics-httpasyncclient and the duration of those requests.

PR #779, opened 2015-04-08
Improves collection of data from metrics-httpasyncclient

derekbassett:
This is directly related to issue #778.  This provides an active count of httpasyncclient requests as well as a duration.


mlaflamm:
I ran into the same issue while trying to instrument my async clients. I was about to modify the instrumentation code myself while I noticed this pull request and I decided to give it a try. It works and I like the addition of the active counter. Unfortunately the current implementation is not robust enough. I got a negative active count after receiving many ```SocketTimeoutException```.

I think that wrapping the ```FutureCallback``` passed to ```CloseableHttpAsyncClient.execute()``` could simplify the code and make it more robust. In the callback wrapper constructor, increment the active counter and create the timer context. Decrement the active counter and stop the timer context when ```FutureCallback.completed()``` or ```FutureCallback.failed()``` are invoked.

Does this make sense or I am missing something important?

derekbassett:
@mlaflamm That makes perfect sense.  I will look into the issue with SocketTimeoutException.  
I originally wrote it using the FutureCallback, and moved it to using the Request/Response because I believe it captured a more accurate time.  Also the active count only increments/decrements when the request is "active."  This is only between when the request is generated but before request is completed, and when the response is received until the response is completed.

Issue #781, opened 2015-04-14
HealthCheckRegistry.unregister() should return the removed HealthCheck


nerdynick:
The HealthCheckRegistry should return the removed HealthCheck when you unregister it. This would allow you for the user to manipulate the check, wrapping it in another class, or in my particular use case. It would allow me to move a check from one registry to another.

ryantenney:
Tracking in #781

PR #782, opened 2015-04-14
#781 - Make HealthCheckRegistry return HealthCheck upon unregister

nerdynick:


nerdynick:
This resolves #781

jplock:
Seems like a simple enough change.  Any reason not to merge @ryantenney ?

ryantenney:
If we do this we should also look at the return type for MetricRegistry.remove, currently a boolean.

PR #785, opened 2015-04-17
Added new metric CounterMeter

nerdynick:
This is a new Metric which allows you to meter a count. Allowing the user to know what the 1min, 5min, 15min, and Mean are of a given counter. 

The use case for this is for when you have a queue, collection, or thread pool that getSize() or similar function is not an option. Currently we use this for measuring the open connections on an Async/NIO connection pool. In this case Jetty. As well as use it for measuring the number of items spooled to an on disk journal(s) for batch sending.


ryantenney:
This is intriguing, but I haven't seen any demand for it.  I'll leave the PR open to gauge support.

ryantenney:
Sorry, didn't mean to close it.

nerdynick:
Sounds good, It was just a usefully metric we wrote and figured we'd publish it in case anyone else wanted it.

We found having a rate vs actually counts for a very high volume counter allowed us to truly see what the real load was. As from tick to tick the counter would look very sporadic. In our case this was Jetty's connection pool on a high transnational server.

marshallpierce:
This uses EWMA, which is only valid for normally distributed values, AFAIK. I'm not enough of a statistician to know what the Right Answer is given the problem you want to solve, but I think EWMA isn't it.

PR #787, opened 2015-04-24
Honour metrics annotation in resources implementing interfaces

remmelt:
When a resource implements an interface, the metrics annotations (@Metered, @Timed, @ExceptionMetered) were not picked up when they were applied to the implementing method. They were picked up when they are on the defining method, in the interface.
The fix looks at the annotation on the interface or defining method first, then on the implementation or handling method after. Now the annotations on implementing methods are picked up.

Our use case is a separate library that holds the API contract, using `@GET` and `@Path` etc. The implementation should be responsible for metering, so the implementing method should have the `@Metered` annotation, not the interface.

The fix is quite simple. I've added a set of tests.

### DIFF
--- a/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
+++ b/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
@@ -194,7 +194,11 @@ public class InstrumentedResourceMethodApplicationListener implements Applicatio
     private void registerTimedAnnotations(final ImmutableMap.Builder<Method, Timer> builder,
                                           final ResourceMethod method) {
         final Method definitionMethod = method.getInvocable().getDefinitionMethod();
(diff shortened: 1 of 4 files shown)


ryantenney:
I'm interested to hear from others as to whether this is a desired change. Please weigh in!

davidkarlsen:
+1

saumitrabhave:
+1

My Rationale Behind It:
We separate out JAX-RS annotations on public interfaces and share with service consumers to generate Proxy Classes [See This](https://jersey.java.net/apidocs/2.9/jersey/org/glassfish/jersey/client/proxy/package-summary.html). 

The Metrics annotations on the other side are completely server Implementation detail and need not be shared publicly.

With this change merged, we can put all metrics annotations on Server Impl and share All JAX-RS Annotations publicly sticking to section 3.6 of JSR-339.

saumitrabhave:
Is this still under consideration?

saumitrabhave:
Is it moved? fixed? rejected?

arteam:
The master branch has been removed and 4.0-development is now upstream. I will take a look at the patch and resubmit if it looks good.

PR #792, opened 2015-05-06
Added null check to prevent NullPointerException when connecting to rabb...

massimo-parente:
I have added a null check in the connect() method of com.codahale.metrics.graphite.GraphiteRabbitMQ.java which prevents the method to fail with NullPointerException


ryantenney:
Have you experienced a null pointer exception? If so, could you send me the stack trace?

massimo-parente:
Hi Ryan, please ignore my pull request.  My mistake, I've had this fix in mind for a while but I forgot that I already fixed this long ago! I just didn't noticed that the null check had been moved in isConnected() method.

Sorry for my wrong pull request

ryantenney:
No problem!

PR #794, opened 2015-05-07
upgrade httpclient to 4.4.1 and fix transitive deps

carlo-rtr:


jplock:
@ryantenney if we submitted a similar PR to the `3.1-maintenance` branch, would it be possible to create a new metrics release with that change?

ryantenney:
Probably not… I'm trying to avoid changing dependency versions in maintenance releases. I'd instead recommend using Maven's Dependency Management to override the versions of these dependencies.

arteam:
The 3.2 branch has been updated to Apache HTTP Client 4.5.2.

Issue #795, opened 2015-05-07
metrics-servlet doesn't record the real processing time and response code of async servlet


mlaflamm:
At the moment, metrics-servlet is only useful for synchronous servlet. It records the time of the initial calling thread and disregard the time spent processing asynchronously in different threads until the request is really completed. We would like to see the complete processing time and real response code for asynchronous servlet.

ryantenney:
Will track in #796

PR #796, opened 2015-05-07
Add support for async servlet to metrics-servlet

mlaflamm:
Proposition for issue #795.

The requests timer and the activeRequests counter now properly observe the entire processing time. Use meters separate from the response code meters for timeouts and errors since the code isn't representative (i.e. 200) for these situations.


ryantenney:
Merged, thanks!

Issue #797, opened 2015-05-12
NullPointerException from MemoryUsageGaugeSet


nerdynick:
I keep getting the following NullPointerException when using the MemoryUsageGaugeSet with the GraphiteReporter. I know this worked with 3.0.2 and before. Not sure about any of the 3.1.* versions.

```
2015-05-11 18:12:14 [ERROR] [metrics-graphite-reporter-1-thread-1] com.codahale.metrics.ScheduledReporter RuntimeException thrown from GraphiteReporter#report. Exception was suppressed.
java.lang.NullPointerException: null
	at com.codahale.metrics.jvm.MemoryUsageGaugeSet$19.getValue(MemoryUsageGaugeSet.java:184) ~[metrics-jvm-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at com.codahale.metrics.jvm.MemoryUsageGaugeSet$19.getValue(MemoryUsageGaugeSet.java:181) ~[metrics-jvm-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at com.codahale.metrics.graphite.GraphiteReporter.reportGauge(GraphiteReporter.java:268) ~[metrics-graphite-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:158) ~[metrics-graphite-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:160) ~[metrics-core-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at com.codahale.metrics.ScheduledReporter$1.run(ScheduledReporter.java:115) ~[metrics-core-4.0.0-SNAPSHOT.jar:4.0.0-SNAPSHOT]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_45]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [na:1.8.0_45]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [na:1.8.0_45]
[...]

PR #799, opened 2015-05-18
Fixed an issue that not all memory pools support collection usage sta…

ryanrupp:
…tistics e.g. Code Cache, Meta Space in Java 8, this was resulting a NullPointerException when trying to access the used-after-gc gauge. Pools that don't support collection usage statistics will skip this metric.


ryanrupp:
Fixes #797

ryantenney:
Thanks!

PR #800, opened 2015-05-20
Provide a way to create a gauge off a CompositeData object in JMX.

larrys:


ryantenney:
I'm going to leave this open and if there proves to be sufficient demand, I'll merge it.

Issue #334, opened 2013-03-13
Figure out how to look up a MetricRegistry by name


codahale:
Right now there's no default registry, so `InstrumentedEhcacheFactory` won't work.

How should something like that or the Jetty connectors figure out which registry to use based on a name? JNDI?

PR #801, opened 2015-05-21
Opt-in default shared registry

ARentz07:
In https://github.com/dropwizard/metrics/issues/334, it was mentioned that the default metric registry has disappeared. While it makes sense to disallow a "default" default registry to prevent having a dumping ground for all metrics, this unfortunately creates a bit more complexity for multilayered services. If there's a shared registry of metrics between them, for instance, each of the pieces that make up a service would need to be cognisant of this default registry's name, even if the name has no semantic meaning. 

In order to reduce complexity, we are aiming to expose a method to set a default registry name and allow consumers to call into SharedMetricRegistries to access this default registry. This enables a single orchestrator (such as a service container) to initialize the default registry with whatever name is suitable, leaving other pieces to just call SharedMetricRegistries.getDefault() in order to access this default registry.

This comes with a couple of caveats obviously:
1. This method should only be called once (hence the IllegalStateException). It should be recommended somewhere in the contract, but as this class has no Javadoc on the methods, I have not included that recommendation yet.
2. The name **must** be set before calling getDefault(). This is akin to the above in which we are trying to guard against consumers making bad assumptions about the state of their metrics registries. The server container, for instance, should set up the default metric registry before spinning up any workers that would access the default registry.
3. Currently, clear() still clears the default registry. To me this is dubious, since this functionality gives the caller power to blow away the shared default registry that might be needed by other threads, etc. Comments on this are more than welcome as we need to decide what to do in this situation.

Tests are written. I'll add them to the pull request once this implementation code is reviewed.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java
@@ -11,6 +11,8 @@ public class SharedMetricRegistries {
     private static final ConcurrentMap<String, MetricRegistry> REGISTRIES =
             new ConcurrentHashMap<String, MetricRegistry>();
 
+    private static volatile String defaultRegistryName = null;
+
     private SharedMetricRegistries() { /* singleton */ }
 
     public static void clear() {
@@ -40,5 +42,21 @@ public class SharedMetricRegistries {
             return raced;
         }
         return existing;
+    }
+
+    public synchronized static MetricRegistry setDefault(String name) {
+        if (defaultRegistryName == null) {
+            final MetricRegistry registry = getOrCreate(name);
+            defaultRegistryName = name;
+            return registry;
+        }
+        throw new IllegalStateException("Default metric registry name is already set.");
+    }
+
+    public static MetricRegistry getDefault() {
+        if (defaultRegistryName != null) {
+            return getOrCreate(defaultRegistryName);
+        }
+        throw new IllegalStateException("Default registry name has not been set.");
     }
 }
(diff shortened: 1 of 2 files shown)


mattnelson:
+1

This is something that I was sorely missing when attempting to uplift to 3.x. I have quite a few instrumented libraries that I do not have the option of injecting the default registry without considerable refactoring.

ARentz07:
I've added the unit tests for this. Any chance it can get reviewed/merged?

ryantenney:
Thanks! I've reviewed it, and it is under consideration for the next major version (v4, no eta as of yet).

metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java
```diff
@@ -41,4 +43,20 @@ public static MetricRegistry getOrCreate(String name) {
         }
         return existing;
     }
+
+    public synchronized static MetricRegistry setDefault(String name) {
```
mattnelson: Should there be another method for instances when you already have the MetricRegistry? I am using Dropwizard where a metrics registry is created by the framework and I would like to make that instance the default.

As proposed:
```java
SharedMetricRegistries.add("default", bootstrap.getMetricRegistry());
SharedMetricRegistries.setDefault("default");
```

With the overloaded setDefault method.
```java
SharedMetricRegistries.setDefault("default", bootstrap.getMetricRegistry());
```

ARentz07: Probably so, will add that...

ARentz07: Added the new method + tests.

ryantenney:
Merged, thanks!

mattnelson:
@ryantenney It looks like there is going to be a 3.2 release to coincide with dropwizard 0.9. Do you think this feature could be added to the 3.2 milestone?

https://github.com/dropwizard/dropwizard/pull/1137#issuecomment-115736554

Issue #802, opened 2015-05-22
Duplicate strings wasting memory in JmxReporter


bentatham:
When looking at memory usage in our system, there are 424 duplicate ``"events/second"`` strings (30,448 bytes of wasted heap).  Not necessarily a huge deal, but trivial to get rid of.

PR #803, opened 2015-05-22
Use same string instance for rateUnit across all instances

bentatham:
Use same string instance for rateUnit across all instances that have the same TimeUnit.

Fixes https://github.com/dropwizard/metrics/issues/802


ryantenney:
I'm open to this, but it's gotta be thread safe.

How about:

```java
this.rateUnit = ("events/" + calculateRateUnit(rateUnit)).intern();
```

bentatham:
Cool - I learn something new every day - never knew about ``String.intern()``.

ryantenney:
Thanks!

PR #804, opened 2015-05-26
Custom rates for Meter

artnaseef:
Added a feature to allow the application to define custom rates for Meters beyond the built-in 1-min, 5-min and 15-min rates.


ryantenney:
Something like this will make it into v4, but it's going to be a much larger change, to allow for reporters to report the values.

PR #815, opened 2015-06-17
Adding a module to write metrics to InfluxDb

leelasharma:
Adding metrics-influxdb module to enable writing application metrics to [influxdb] (https://influxdb.com/docs/v0.9/introduction/overview.html). 
The latest v0.9 influxdb release was announced recently and it will be great to have metrics report to influxdb as well, similar to graphite.

### DIFF
--- a/metrics-influxdb/src/main/java/com/codehale/metrics/influxdb/InfluxDbHttpSender.java
+++ b/metrics-influxdb/src/main/java/com/codehale/metrics/influxdb/InfluxDbHttpSender.java
@@ -1,0 +1,117 @@
+package com.codehale.metrics.influxdb;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.codec.binary.Base64;
+
+import com.codehale.metrics.influxdb.data.InfluxDbPoint;
+import com.codehale.metrics.influxdb.data.InfluxDbWriteObject;
+import com.codehale.metrics.influxdb.utils.InfluxDbWriteObjectSerializer;
+
+/**
+ * An implementation of InfluxDbSender that writes to InfluxDb via http.
+ */
+public class InfluxDbHttpSender implements InfluxDbSender {
+    private static final Charset UTF_8 = StandardCharsets.UTF_8;
+    private final URL url;
+    // The base64 encoded authString.
+    private final String authStringEncoded;
+    private final InfluxDbWriteObject influxDbWriteObject;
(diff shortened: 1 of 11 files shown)


cemo:
@ryantenney I know you are reluctant for new modules but InfluxDb usage is greatly increased. Adding support for InfluxDB would be nice. 

:+1:

ryantenney:
Given the newfound focus in v4 on tagged metrics, I will be accepting new reporters for backends which support tags.

leelasharma:
@ryantenney That's great to know. This metrics-influxdb module also supports tags.

ryantenney:
@leelasharma merged, thanks!

leelasharma:
@ryantenney @cemo  Actually no real reason, thanks for pointing out. I can submit a pull request to use the same jackson libraries as the project.

metrics-influxdb/src/main/java/com/codehale/metrics/influxdb/InfluxDbHttpSender.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param hostname      the influxDb hostname
+     * @param port          the influxDb http port
+     * @param database      the influxDb database to write to
+     * @param authString    the authorization string to be used to connect to InfluxDb, of format username:password
+     * @param timePrecision the time precision of the metrics
+     * @throws Exception exception while creating the influxDb sender(MalformedURLException)
+     */
+    public InfluxDbHttpSender(final String hostname, final int port, final String database, final String authString,
+                              final TimeUnit timePrecision) throws Exception {
+        this.url = new URL("http", hostname, port, "/write");
```
mwuertinger: We should either add a flag to support HTTPS or we should offer a constructor allowing to pass in the full URL.

ryantenney:
Agreed. I am planning on rewriting this to use the Apache http client. I'll make certain it supports https.

mnuessler:
In general HTTPS support makes perfect sense, but InfluxDB 0.9 does currently not support HTTPS. There is an [open ticket](https://github.com/influxdb/influxdb/issues/2115) and it looks like it will be included in the next minor release (0.9.2).

For the rewrite it could make sense to have a look at the official [InfluxDB Java client|https://github.com/influxdb/influxdb-java] too. Seems that they are close to releasing a new version that supports the new InfluxDB 0.9 API.

jhorwit2:
I'd suggest also looking at line protocol versus json as it's the preferred method and not deprecated like JSON is. With line protocol there are some performance notes to remember, like sorting tags & escaping certain chars.

mnp:
So what's the status on this feature? Is anyone working on it?

arteam:
https://github.com/davidB/metrics-influxdb and https://github.com/iZettle/dropwizard-metrics-influxdb are the most up to date InfluxDB reporters for Metrics 3.2. Unfortunately, for the 4.0 version there is no active maintained reporter under the `io.dropwizard.metrics` package.

mnp:
Okay thanks for the update!

PR #817, opened 2015-06-21
Add non-accumulating garbage collector metric set

gburton1:
The existing GarbageCollectorMetricSet is not ideal for metrics that are reported to graphite because it keeps cumulative counts and times for each garbage collector over the life of the JVM process. It is more interesting to see the counts/times for the elapsed time since the last data point was reported. This NonAccumulatingGarbageCollectorMetricSet wraps the GarbageCollectorMetricSet to provide this functionality. In addition, it provides a GC throughput gauge that is often useful as a summary measurement of application health and capacity.

### DIFF
--- a/metrics-jvm/pom.xml
+++ b/metrics-jvm/pom.xml
@@ -22,5 +22,10 @@
             <artifactId>metrics-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.3.2</version>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


ryantenney:
This is problematic because there can be multiple reporters polling the gauges, each reporting the value since a different reporter polled it. Graphite and other data outputs can compute a derivative which, given a series of cumulative counts and times, will calculate the delta between subsequent data points. See [nonNegativeDerivative](https://graphite.readthedocs.org/en/0.9.12/functions.html#graphite.render.functions.nonNegativeDerivative) in the Graphite docs.

gburton1:
I know what you mean about multiple reporters. That's why I designed it to harvest non-accumulating values in a scheduled background thread, rather than by reporters. In this way, it works with any number of reporters, and the unit tests validate this behavior.

Good point about the Graphite function. I know that we can apply several nested functions (including derivative) to calculate GC throughput from the raw metrics, but we prefer to store the pre-computed value of this important summary metric for the sake of simplicity in our operations (people can see it without figuring out how to slice and dice the raw data from various garbage collectors). It's also nice to have all the other garbage collector metrics in a format that avoids additional calculations on the Graphite side. 

We have been using this for a while at Orbitz, and I thought it might be useful to the broader community as well.

PR #818, opened 2015-06-22
Updating jackson dependencies to match the project

leelasharma:
Getting rid of org.codehaus.jackson dependency and using the same jackson dependencies as the metrics-json module.


ryantenney:
Thanks!

PR #819, opened 2015-06-22
Log and ignore Exceptions from Meters in GraphiteReporter

incubos:
`GraphiteReporter` stops sending metrics to Graphite in case of any non-`IOException` thrown.


ryantenney:
I'd prefer to wrap the call to `reportGauge` in a try, so that a single gauge throwing an exception doesn't prevent the remaining data from being sent to graphite. However, gauges really shouldn't throw exceptions.

incubos:
I agree that `Meter`s shouldn't throw exceptions, but they can and sometimes do (in our use case). Even `Timer` can do that in case of an unexpected system time change. So it is a defensive measure to recover from erroneous `Meter` implementations.

`ignoresExceptionsFromGauges` unit test tests only `Gauge`, but I can add unit tests for other `Meter`s if you think it is necessary.

PR #823, opened 2015-06-27
Update for Jetty 9.3.0

arteam:
Jetty 9.3.0 is the latest update of Jetty with some API changes.

* Update `InstrumentedConnectionFactory` to support instrumentation for Jetty 9.3. 
The update should be backward-compatible with Jetty 9.1 and 9.2, because we only implement a new method.

### DIFF
--- a/metrics-jetty9/src/main/java/io/dropwizard/metrics/jetty9/InstrumentedConnectionFactory.java
+++ b/metrics-jetty9/src/main/java/io/dropwizard/metrics/jetty9/InstrumentedConnectionFactory.java
@@ -8,19 +8,41 @@ import org.eclipse.jetty.util.component.ContainerLifeCycle;
 
 import io.dropwizard.metrics.Timer;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.List;
+
 public class InstrumentedConnectionFactory extends ContainerLifeCycle implements ConnectionFactory {
     private final ConnectionFactory connectionFactory;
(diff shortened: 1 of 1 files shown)


arteam:
The obvious downside is that we can't run tests under Java 7, because Jetty 9.3.0 is compiled against Java 8.

ryantenney:
I'd like to be able to run tests against Jetty 9.2 and Java 7.  Any idea how we can make that happen?

arteam:
What do you think about using reflection to access the `ConnectionFactory#getProtocols` method?
We keep Java 7 compile-time dependency, and in the same time comply with Jetty 9.3.* in runtime.

arteam:
Ping. @ryantenney, could you take a look, please?

drallgood:
Just found this while trying to figure out why my app was failing when I tired to use Jetty 9.3.
Any news when this will be merged?

jplock:
@arteam is this good to merge now?

arteam:
Yes, the patch should be good to merge. Last time I checked, it worked both on Jetty 9.2.* and Jetty 9.3.*.

jplock:
And Java7?

arteam:
I didn't check against Java 7, but the tests passed on Travis CI. I used reflection precisely for the reason to avoid any Java 8 dependency.

arteam:
I've created #1037 to backport this feature to the 3.2 branch. For the 4.0.0 branch we can actually just avoid reflection and just implement a new method.

arteam:
The implementation for 4.0.0 is tracked in #1039.

PR #827, opened 2015-07-06
adds enable/disable feature to MetricRegistry

shivamsharma:
Toggleable interface is implemented by the metrics which support
enable/disable feature. when metric is disabled, all values
collected so far are reset and any calls to update the metrics
do nothing.


ryantenney:
I'm sorry to to say that these features are not on our roadmap.  I appreciate the work you put into this.

shivamsharma:
Ohh ... I didn't know it. But I think it is quite important feature. It can be use for many things.

PR #829, opened 2015-07-11
Code safety in ExponentiallyDecayingReservoir

bwarminski:
While reading the implementation, these two things jumped out at me.

Getting a random double == 0.0 and overflowing a long counter are unlikely, but who knows.


ryantenney:
I'm sold on the random fix, but not on the long overflow. At 1mil event/sec it'd take almost 300,000 years to get there.

ryantenney:
Cherrypicked 172e44b as 258e80b25473184535df4f0c7ec9a80554b21f6a

PR #834, opened 2015-07-23
Handle unexpected exceptions when reporting to graphite

jsravn:
This prevents an unexpected exception from killing the reporter thread
silently. Instead log the exception and close the connection gracefully.

We observed this problem a few times with hystrix (for instance, hystrix 1.4.7 implements its own metrics which have had various bugs - in one case we saw an ArrayOutOfBoundsException when calling #getValue on a gauage). It was difficult to debug as the reporter thread silently dies since the exception is unhandled. In general I think we should be careful to avoid killing the scheduler thread.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -175,13 +175,17 @@ public class GraphiteReporter extends ScheduledReporter {
             }
 
             graphite.flush();
+        } catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


ryantenney:
ScheduledReporter should already make it difficult/impossible to kill the thread: https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java#L111-L120

jsravn:
I missed that. Then I don't know what was killing our reporter thread, but it happened a few times.

ryantenney:
There is a known issue where the Graphite reporter doesn't always reconnect
after disconnecting.

PR #839, opened 2015-07-23
Tagged metric names

ryantenney:
This is ready to go into master, but there is at least one thing I'd like to tackle before this is released:

- [ ] Minimize memory usage by interning MetricNames and/or using a map implementation in MetricName with lower memory requirements than HashMap.


dk14:
Is there any protection from metrics registry overflow in case if a lot of metrics with same name (and tag name), but different tag value was submitted?

Basically I had to do it for my project, because there were some cases where people were submitting requestId as a tagName (and guess how many metricNames were stuck in `MetricsRegistry`), so I gave them a value-count-limit per metric+tagName (and used CAS-based ring to allow more-or-less fare tag-value distribution and quick rolling of such values).

PR #845, opened 2015-08-03
added reconnect flag

rmalchow:
we have the situation that in some places, we might not have graphite available at all. so i would like to be able to tell the graphite reporter to stop reconnecting. 

for this, i added a configuration flag, and if there is an error connecting and this flag is off, i stop the executor. 

the default setting is "true", so it behaves like before.


lazyval:
> we have the situation that in some places, we might not have graphite available at all. so i would like to be able to tell the graphite reporter to stop reconnecting.

If it is not available at all, why bother with disabling reconnection (and making one connection attempt) -- pick no-op reporter instead

rmalchow:
Because it is the same war file. I could build a configuration with a
switch on top ... But why not build the switch into the reporter?
Alternatively, i could try to add a setting for a pause ... The situation
is now that reporting every 10 seconds causes a lot of log file pollution,
even if graphite just disappears for a while. So instead of completely
stopping to reconnect, there could be a setting that says "try again in N
seconds" ~ would you find that more useful?

arteam:
We went with another approach of reconnecting to a Carbon server on every report. #1047.

PR #849, opened 2015-08-06
Support default health check registry. Refactor default metric registry.

mattnelson:
Extension of #801 to support a default registry with SharedHealthCheckRegistries as well.

Also when @ARentz07 and I started to consume the default metric registry, we realized that it did not make much sense to require a name to be associated to the default. Since the default registry changes have not been released yet, I wanted to get these changes in before 4.0 is released.

The default registry test assertions were also refactored into a single test method in order to avoid using reflections to clear the state.


mattnelson:
Will rebase on master to fix build failure once #850 is merged.

metrics-healthchecks/src/test/java/io/dropwizard/metrics/health/SharedHealthCheckRegistriesTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SharedHealthCheckRegistriesTest {
+    @Before
+    public void setUp() throws Exception {
+        SharedHealthCheckRegistries.clear();
+    }
+
+    @Test
+    public void memoizesRegistriesByName() throws Exception {
```
ARentz07: "memorizes"

ARentz07: Nvm, that's correct

mattnelson:
Going through some of my old pulls and noticed this was still open. Is there anything else blocking this from being merged?

jplock:
@ryantenney what do you think about getting this into 4.0.0 and possibly 3.2?

ryantenney:
I've been thinking and what bothers me about this is that we'd end up with a registry that is inaccessible from `getOrCreate`.

mattnelson:
I hadn't considered how this impacted `getOrCreate`. Staying on the current magic string implementation seems like the cleanest option. Sticking with the string option, there are a few things that need to be addressed/documented. The `remove` and `clear` methods are potentially destructive. A caller who holds on to the default registry could be left holding onto a different instance of the default.

jplock:
After upgrading Dropwizard to Metrics 3.2.0 and doing https://github.com/dropwizard/dropwizard/pull/1938, it would have been nice to get this PR into 3.2.0.  Maybe we can merge it into master and cherry-pick it into `3.2-development`?

arteam:
That's a good idea. I'm going to rebase this pull request on the 3.2 branch.

arteam:
On second thought, I think this PR can't be accepted because it changes the public API. The `setDefault(name, registry)` method is removed and some users can depend on it. Let's split this PR in two parts

* fixing race conditions on `defaultRegistryName`
* adding `SharedMetricRegistries`

Issue #859, opened 2015-08-13
@Timed, @Metered @ExceptionHandler on jersey2 inherited resources don't work


vrcar:
My dropwizard app stop working when I migrate it from 0.7.1 to 0.8.X (metrics-jersey to metrics-jersey2). The problem is with inherited resources and metrics. In “metrics/jersey2/InstrumentedResourceMethodApplicationListener.java” timers, meters and exceptionMeters maps use java.lang.reflect.Method as key. So when a child resources try to register annotations it fails due a duplicate key (the parent method is already registered by a different resource). I saw in the previous metrics-jersey module that annotations was registered using the resource class name and the method. I think this approach should work in the jersey2 implementation if there is no side effects changing Method key to String in the maps.

I work in my own fix, could this help?

marshallpierce:
Unfortunately, counting on method name won't always work like it did in jersey 1. Jersey 2 lets you do more complicated ways of applying logic that don't really have a useful method name, e.g.:

```
  final Resource.Builder resourceBuilder = Resource.builder();
        resourceBuilder.path("programmaticResource");

        final ResourceMethod.Builder methodBuilder = resourceBuilder.addMethod("GET");
        // using a lambda makes jersey sad because it can't reflectively access the type
        //noinspection Convert2Lambda
        methodBuilder.produces(MediaType.TEXT_PLAIN_TYPE)
            .handledBy(new Inflector<ContainerRequestContext, Object>() {
                @Override
                public Object apply(ContainerRequestContext containerRequestContext) {
                    return "Such dynamic. Very resource. Wow.";
                }
            });

        final Resource resource = resourceBuilder.build();
        registerResources(resource);
```

That's one of the reasons I started my own jersey2 integration (https://bitbucket.org/marshallpierce/jersey2-metrics).

vrcar:
Ok, I miss that... Then use the http verb + path instead of class name + method name? I can try to arrange it if it's the right approach.

marshallpierce:
This is one of the options my project has; see https://bitbucket.org/marshallpierce/jersey2-metrics/src/9cf28f8e5292486449a17893915980bdee575f31/jersey2-metrics-core/src/test/java/org/mpierce/jersey2/metrics/RequestPathMetricNamerTest.java?at=master#RequestPathMetricNamerTest.java-75 for instance.

[...]

Issue #746, opened 2015-02-05
@Timed annotation on resource super class method throws exception on init


jkytomaki:
If multiple jersey resource classes extend a common base class, and that base class has one resource method with @Timed annotation, Dropwizard's initialization will fail with e.g.

```
java.lang.IllegalArgumentException: duplicate key: public javax.ws.rs.core.Response 
com.codahale.metrics.jersey2.
InstrumentedResourceMethodApplicationListenerTest$ResourceBase.getCommonThings()
```

The same resource classes work with Metrics 2.2.0.

The problem seems to be in the way the map key is selected for registered metrics in InstrumentedResourceMethodApplicationListener's registerTimedAnnotations method (and registerMeteredAnnotations as well). A java.lang.reflect.Method is used as a key, but for the resource method inherited from the base class, it is the same instance for all of the subclasses and thus the initialization will fail on key duplicity if there are more than one subclass.

Here is an example junit test that should replicate the problem (the test method is empty, but the test should still fail on initialization).

``` java
public class InstrumentedResourceMethodApplicationListenerTest extends JerseyTest{

  @Override
  protected Application configure() {
    MetricRegistry registry = new MetricRegistry();
    InstrumentedResourceMethodApplicationListener listener = new
            InstrumentedResourceMethodApplicationListener(registry);

    ResourceConfig rc = new ResourceConfig(FooResource.class, BarResource.class);
    rc.register(listener);
    return rc;
  }

  public abstract class ResourceBase {
    @GET
    @Path("/common-things")
    @Timed
    public Response getCommonThings(){
      return Response.ok().build();
    }
  }

  @Path("/foos")
  public class FooResource extends ResourceBase {
    @GET
    public Response getFoos(){
      return Response.ok().build();
    }
  }

  @Path("/bars")
  public class BarResource extends ResourceBase {
    @GET
    public Response getBars(){
      return Response.ok().build();
    }
  }

  @Test
  public void test() throws InterruptedException {
  }

}
```

marshallpierce:
Could you give my alternate jersey2 integration a try? It should not suffer from this problem. https://bitbucket.org/marshallpierce/jersey2-metrics

OlaviSau:
By the way check out HttpContext for automatic paths

mtakaki:
This is actually an issue on the metrics package and it's being tracked under: https://github.com/dropwizard/metrics/issues/859

[...]

PR #861, opened 2015-09-01
#859 fix inherited timed annotation

mtakaki:
Suggestion for fixing the the issue #859 that prevented having resources inheriting annotated methods.
Also fixing the support for absolute metric names that apparently was not working properly.

### DIFF
--- a/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
+++ b/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
@@ -6,7 +6,6 @@ import com.codahale.metrics.Timer;
 import com.codahale.metrics.annotation.ExceptionMetered;
 import com.codahale.metrics.annotation.Metered;
 import com.codahale.metrics.annotation.Timed;
-
 import org.glassfish.jersey.server.model.Invocable;
 import org.glassfish.jersey.server.model.ModelProcessor;
 import org.glassfish.jersey.server.model.Resource;
@@ -16,11 +15,9 @@ import org.glassfish.jersey.server.monitoring.ApplicationEvent;
 import org.glassfish.jersey.server.monitoring.ApplicationEventListener;
 import org.glassfish.jersey.server.monitoring.RequestEvent;
 import org.glassfish.jersey.server.monitoring.RequestEventListener;
-import org.omg.CORBA.DefinitionKind;
 
 import javax.ws.rs.core.Configuration;
 import javax.ws.rs.ext.Provider;
-
 import java.lang.reflect.Method;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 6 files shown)


rferreira:
hey folks what's the good word on this PR?

mtakaki:
I haven't heard anything regarding my proposed fix for this issue, @rferreira. My suggestion at https://github.com/dropwizard/metrics/issues/746 is still valid. You can import this fix into your project at any time. To make it easier, just add this file: https://raw.githubusercontent.com/mtakaki/metrics/mtakaki_746_fix_inherited_timed_annotation/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java

edwardsb:
@mtakaki I did as you suggested, and got

java.lang.NoSuchMethodException: Could not find a suitable constructor in com.mywebservice.InstrumentedResourceMethodApplicationListener class.

mtakaki:
@edwardsb which version of dropwizard are you using? I noticed it didn't work as soon as I migrated from `0.8.x` to `0.9.x`. So I had to do some changes to it: https://gist.github.com/mtakaki/746dba30edfefc0ffbd5

To be honest, I completely gave up about having my fix merged. So this has been my current approach to get around this issue.

edwardsb:
I am using 0.9.x.

I copied your gist in and its no longer throwing the suitable constructor error.

Though I am now getting null pointers on exception metered methods of sub-resources. For some reason my top level resources that have @Timed, @Metered, and @ExeptionMetered are getting registered on startup, but sub-resource still don't look like they are in the maps of timers, meters, and exception meters.

Ultimately this line is called:
                        final ExceptionMeterMetric metric = this.exceptionMeters.get(chooseName(annotation.name(), annotation.absolute(), handlerClass, definitionMethod));

and since there is nothing in the map with that key, metric is null and then we call a method on it.

Do you register your sub-resources with the Jersey environment? I tried that too and it wasn't working.

ryantenney:
@mtakaki I appreciate the fix, and it will be merged, but not into 3.1. It'd be perfect for 3.2, but I'm trying to focus on v4, and you can expect the fix to be included there.

mtakaki:
@edwardsb I'll take a look tonight. So you have two resources, a parent and a child, and both are being used as resources? I'll try this later today, but I don't think why it would not work. And yes, you need to register all the concrete resource classes. 

@ryantenney sorry, I didn't know you were too busy to look at my PR. Thanks for working on the version 4 :) Really appreciate the good work you're doing with the metrics project.

edwardsb:
@mtakaki Yes here is a gist of what I am working with: https://gist.github.com/edwardsb/0a8d232c66bf54f0b75a

Thanks for the help. Do you mind if I message you on twitter?

mtakaki:
@edwardsb not at all. Feel free to message me on twitter.

leeavital:
This is something we would find useful as well. I'd be glad to help in any way :)

rkalicinski:
Is it possible to merge those commits?

PR #864, opened 2015-09-16
add metrics-atsd

Sild:
We would like to add metrics-atsd module to your project.

### DIFF
--- a/docs/source/manual/atsd.rst
+++ b/docs/source/manual/atsd.rst
@@ -1,0 +1,188 @@
+
+.. _manual-atsd:
+
+#####################
+Reporting To ATSD
+#####################
+
+The ``metrics-atsd`` module implements ``AtsdReporter``, which allows Java applications to stream metrics into `Axibase Time-Series Database`__ .
+
+.. __: https://axibase.com/products/axibase-time-series-database
+
+#####################
+Configure Sender
+#####################
+
+TCP:
+
+.. code-block:: java
+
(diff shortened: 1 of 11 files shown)


ryantenney:
I'm afraid I can't accept new integrations at this time, as maintaining so many integrations that I don't personally use has become extremely onerous.  I'd encourage you to separate this module into its own project, publish it to Maven Central, and open a pull request to add a link to the project in the file `third-party.rst`.

rodionos:
We were actually planning to provide corporate support behind this code, minimizing the maintainers' time. It would useful if you had a note in project documentation saying that master is closed to any integrations outside of maintainers' scope. At least we would have reached out to you before writing code and then seeing it dumped.

ryantenney:
Added this a while ago, though after the PR was opened: https://github.com/dropwizard/metrics/blob/master/CONTRIBUTING.md#important

There's no way I can accept even a fraction of the modules that people have written (see https://github.com/dropwizard/metrics/blob/master/docs/source/manual/third-party.rst)

rodionos:
Makes sense, too bad we missed it. Have you thought about abstracting reporter code under a 'universal' driver/reporter so that concrete implementations can be plugged in simply by specifying class name, not unlike logger class in logback/log4j configuration?

ryantenney:
Oh yea that's already possible, there is no requirement that the reporters exist in this project or even in the same package. Check out any of the reporters listed in third-party.rst (or even metrics-datadog, since I already happened to have this link on my clipboard https://github.com/coursera/metrics-datadog).

rodionos:
Ok, thanks for the info.

PR #868, opened 2015-10-02
Executor did not terminate message to use logger

dannypurcell:
I noticed this message on the console occasionally and figured it could do to have it under the logger control like the others. 
Setting it at warn since there is not much the user could really do about it.


ryantenney:
Thanks!

PR #869, opened 2015-10-06
Refactor influxdb reporter to use Apache HttpClient

davidkarlsen:
A drop-in replacement for the current url-based impl.
However I'd like to "structure" the configuration better, and take username, password parameters instead of an username:pw authstring. WDYT?

### DIFF
--- a/metrics-influxdb/pom.xml
+++ b/metrics-influxdb/pom.xml
@@ -30,7 +30,11 @@
         <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
-            <version>1.10</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.1</version>
         </dependency>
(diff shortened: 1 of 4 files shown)


davidkarlsen:
Also - what do you think about excluding commons-logging from the http client, and use jcl-over-slf4j instead as slf4j seems to be the logging framework in use?

davidkarlsen:
Ping - any thoughts on this?

ryantenney:
Looks pretty good, thanks! I definitely agree about excluding commons-logging.

davidkarlsen:
OK - then I'll fix that - what about splitting the auth string into username and password properties?

ryantenney:
Definitely, and don't worry about constructor binary compatibility.

davidkarlsen:
Fixed - can you pull this and I'll fix the spring-metrics as well?

davidkarlsen:
@ryantenney Maybe this can be pulled now and I can then fix spring-metrics after this as well?

ryantenney:
Thanks! Sorry I missed your comment the other day.

PR #854, opened 2015-08-10
Initial commit of the metrics-collectd plugin.

timezra:
This is a pull request for adding a new metrics reporter that uses the collectd binary protocol over UDP (https://github.com/dropwizard/metrics/issues/853).
I have been in a situation where I funnel many of my application and reporting metrics through a 3rd-party tool (which has an input codec for collectd over UDP along with a bunch of other protocols), and there appears to be a conflict between the UDP listener that gathers collectd data from the standard collectd plugins and both the listener for metrics logs or the graphite listener that pulls in metrics data from the graphite reporter.
This collectd reporter is an attempt to put metrics data and collectd data on even footing when it is pulled into a 3rd-party tool.
Perhaps someone else will find it useful as well.

PR #871, opened 2015-10-10
870: collectd reporter

dernasherbrezon:
Essentially it is the same as graphite udp reporter but implements collectd protocol.


ryankennedy:
It might be good to pull either/both this and #854 as Google Cloud Platform supports collectd with Compute Engine, which is relatively easy to get up and running with Dropwizard. I can try to engage some folks at Google if it would be helpful.

ryantenney:
@ryankennedy I didn't realize, that's good to know, thanks!

ryantenney:
Thanks!

PR #875, opened 2015-10-14
Remove usage of Unsafe

ryantenney:
Thanks to @ikedahi for making me aware of AtomicLongFieldUpdater.

Seems to be a small-ish performance hit, 92,000,000 ops/s now vs 105,000,000 ops/s before.


ikedahi:
That's a subtle performance hit :(
I'm not sure whether that is worth or not.

ryantenney:
I've created PR #876 which selects the best LongAdder implementation

PR #878, opened 2015-10-27
Timer.Context now implements AutoCloseable

kilotaras:
Small improvement that allows Timer.Context to be used in try-with-resources

### DIFF
--- a/metrics-adapter/src/main/java/com/codahale/metrics/Timer.java
+++ b/metrics-adapter/src/main/java/com/codahale/metrics/Timer.java
@@ -11,7 +11,7 @@ import io.dropwizard.metrics.Sampling;
 public class Timer implements Metered, Metric, Sampling {
 	final io.dropwizard.metrics.Timer timer;
 
-	public static class Context implements Closeable {
+	public static class Context implements AutoCloseable {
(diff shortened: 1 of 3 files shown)


jmason:
:+1: from a random watcher here -- we do this in our wrapper for Timer.Context and it's really nice.

ryantenney:
I like it. Can you also remove Closeable, or does that cause problems?

kilotaras:
AutoCloseable doesn't implement Closeable (no idea why though). Removing it would be a breaking change.

https://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html

ryantenney:
Yea, we can live with that. We're working on v4, so breaking changes are welcome.

kilotaras:
Here you go.

marshallpierce:
Closeable extends AutoCloseable, not the other way around, because AutoCloseable allows throwing Exception.

The Closeable imports can be removed now, right?

ryantenney:
I reverted the change to Timer in metrics-adapter, since that module is around for backwards compatibility with 3.x, and we shouldn't do anything that could break it.  Also I removed the Closeable import from Timer in metrics-core.

PR #882, opened 2015-11-02
Allow a pluggable strategy for the name of the CSV files that the CsvReporter uses

wimdeblauwe:
This PR extracts the naming of the CSV file from the CSV reporter itself. This allows users to come up with their own naming strategy. 
Using this code as a base, I have implemented a naming strategy that uses the current date to generate a file per day and allows to keep x days of files. If interested I can also contribute this in a separate pull request.

NOTE: Something went wrong when I tried to [fix my username](https://help.github.com/articles/changing-author-info/) so I had to delete my fork and create a new fork. So https://github.com/dropwizard/metrics/pull/879 is no longer valid. Sorry about that.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/CsvReporter.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/CsvReporter.java
@@ -140,7 +140,7 @@ public class CsvReporter extends ScheduledReporter {
     }
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CsvReporter.class);
-    private static final Charset UTF_8 = Charset.forName( "UTF-8" );
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
 
     private final File directory;
     private final Locale locale;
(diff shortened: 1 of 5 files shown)


metrics-core/src/main/java/io/dropwizard/metrics/CsvReporter.java
```diff
@@ -126,16 +134,18 @@ public CsvReporter build(File directory) {
...
                                    clock,
-                                   filter);
+                                   filter,
+                                   csvFileProvider);
         }
     }
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CsvReporter.class);
-    private static final Charset UTF_8 = Charset.forName("UTF-8");
+    private static final Charset UTF_8 = Charset.forName( "UTF-8" );
```
ryantenney: There are a number of lines like this which haven't changed except for the insertion of some spaces. Could you fix these and make sure that all other method invocations don't contain spaces except after commas?

ryantenney:
Thanks! There's just one issue with formatting, if you can fix that I'll merge this.

wimdeblauwe:
Great! Formatting should be ok now.

ryantenney:
Thanks!

PR #885, opened 2015-11-06
Fix metrics httpasyncclient timing master

juretta:
This PR fixes a problem with the httpasyncclient metrics integration. 

The previous approach did not actually time the request execution but the time to return the `Future` of the response. This was discovered when introducing latency in our code that wasn't reflected at all in the metrics we gathered from the async client library.

Changes:
- The first commit is the minimal change, it does slightly tweak the existing test (using Mockito annotations) as part of introducing another test suite for the timing tests.
- The second commit replaces the `AssertionError` that is being thrown with a `RuntimeException` (this code should not throw an `Error`).
- The last commit refactors and enhances the tests to use a local server instead of making requests to `example.com`.

Happy to change the scope of this PR. Also happy to backport this change to the 3.1 branch if desired.

### DIFF
--- a/metrics-httpasyncclient/src/main/java/io/dropwizard/metrics/httpasyncclient/InstrumentedNHttpClientBuilder.java
+++ b/metrics-httpasyncclient/src/main/java/io/dropwizard/metrics/httpasyncclient/InstrumentedNHttpClientBuilder.java
@@ -68,7 +68,7 @@ public class InstrumentedNHttpClientBuilder extends HttpAsyncClientBuilder {
                 try {
                     timerContext = timer(requestProducer.generateRequest()).time();
                 } catch (IOException | HttpException ex) {
-                    throw new AssertionError(ex);
+                    throw new RuntimeException(ex);
                 }
                 return ac.execute(requestProducer, responseConsumer, context,
                         new TimingFutureCallback<>(callback, timerContext));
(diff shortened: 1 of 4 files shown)


juretta:
There is a race condition in the current approach. 

Materialising the future and calling the callback methods is not an atomic operation. See `org.apache.http.concurrent.BasicFuture` so the timer is being stopped with a potential delay (i.e. the future is already "done" but the timer hasn't stopped yet).
`org.apache.http.concurrent.BasicFuture` does the right thing as you don't want to call out to foreign code from within a synchronised block. 

I don't think it's necessarily a problem (worst case is that the timer slightly overreports) but wanted to point it out.

juretta:
Anything I can do to help getting this merged?

juretta:
Just checking in after another 2 months. Is there anything I can do/help with to get this over the finish line?

Thanks!

robinst:
Any updates? The current implementation of metrics for the async client is non-functional.

The PR's changes to the production code is fairly small, most of the changes are tests that ensure the behavior works now.

marshallpierce:
Why not just move this to an external project with a different lifecycle? It doesn't make sense for metrics to need to release a new version every time AHC changes its API...

juretta:
I don't have any opinion on whether this should be a separate library or not. 

But this is not a change due to any API changes in AHC. This is simply a bug in the the `InstrumentedNHttpClientBuilder` based on (I'm assuming) a misunderstanding on what's going on here. This change could have been in any bugfix/patch release as far as I'm concerned.

marshallpierce:
Yes, but my point is that as it stands, metrics would need to release if AHC changed its API, which doesn't make a lot of sense. This would also probably make it easier for changes to get merged and shipped.

I maintain several integrations with metrics that are external projects and it's been quite straightforward to keep up with the (rare) metrics API changes.

juretta:
This seems to be slated for 4.0.0? As it stands metrics are broken for the async http client. Can this be merged and release in a 3.x release as it really is a bugfix?

arteam:
I had a look and this PR looks good to me. I will merge it to the 4.0.0 branch and try to port to the 3.2 manually. The new 4.0.0 will probably created from scratch, anyway. Thank you for your efforts on this issue.

juretta:
Thanks @arteam ! 

I think https://github.com/juretta/metrics/commits/fix-metrics-httpasyncclient-timing-3.1-maintenance-with-local-server might be helpful. That's where I started before I changed it to target the then current `master` branch.

PR #891, opened 2015-11-25
Modify Timer.Context for Extensibility

vjkoskela:
Change Timer.Context constructor to protected from private so that the class may be extended.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/Timer.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Timer.java
@@ -19,7 +19,12 @@ public class Timer implements Metered, Sampling {
         private final Clock clock;
         private final long startTime;
 
-        private Context(Timer timer, Clock clock) {
+        /**
+         * Protected constructor for extension.
+         * @param timer the {@link Timer} implementation the context should reference
+         * @param clock the {@link Clock} implementation the context should reference
+         */
+        protected Context(Timer timer, Clock clock) {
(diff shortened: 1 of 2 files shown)


ryantenney:
What's your use case for this? I generally avoid expanding visibility if not absolutely critical.

vjkoskela:
I completely agree. We are writing an interop layer between our library and Codahale metrics. Code already using Codahale metrics will be able to output data through our library. We need to create an instance of Timer.Context in order to satisfy the contract of Codahale's Timer in our library where we subclass Timer. Internally, this is being used in an extension of DropWizard to get it to interoperate with existing metrics infrastructure. Happy to discuss this further if you would like more details.

vjkoskela:
@ryantenney any further thoughts on this?

ryantenney:
First class support for custom metrics is one of my priorities for Metrics 4. As you know, as of Metrics 3, custom metric types aren't supported, and understandably you're trying to work around that. I've decided that I don't want to start making changes to things in 3 to allow hacking in custom metric types.

vjkoskela:
Fair enough. Is there a mailing list or issue I can subscribe to for updates about the release of version 4?

ryantenney:
There's the metrics-user group, but there hasn't been much in the way of conversation about v4 yet. I do need to start engaging the group on discussions about v4 priorities. https://groups.google.com/forum/#!forum/metrics-user

PR #894, opened 2015-11-29
Code quality fix - Serializable classes should have a version id

AhmedDaraz:
This pull request is focused on resolving occurrences of Sonar rule squid:S2057 - “Serializable classes should have a version id”. You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS2057

Please let me know if you have any questions.

Ahmed

### DIFF
--- a/metrics-json/src/main/java/io/dropwizard/metrics/json/HealthCheckModule.java
+++ b/metrics-json/src/main/java/io/dropwizard/metrics/json/HealthCheckModule.java
@@ -15,6 +15,7 @@ import java.util.Arrays;
 
 public class HealthCheckModule extends Module {
     private static class HealthCheckResultSerializer extends StdSerializer<HealthCheck.Result> {
+        private static final long serialVersionUID = 7926472295622776147L;
         private HealthCheckResultSerializer() {
             super(HealthCheck.Result.class);
         }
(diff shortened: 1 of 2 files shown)


marshallpierce:
I looked up why those Serializer classes are considered to also be `java.io.Serializable`, and it turns out that Jackson's `StdSerializer` was made `Serializable` in [this commit](https://github.com/FasterXML/jackson-databind/commit/babd8cbe375f2287b5c8dc51e1c3f5b2b5f04acc#diff-7dc7ea522e5b3e588f9152186a2f7bca) with comment "minor javadoc changes". This sounds like it was maybe accidental?

This makes me think that it may not be intentional to have that class be `Serializable`, and therefore we shouldn't give the impression in these subclasses that it is actually supported to use built-in serialization on them.

AhmedDaraz:
Ok, So what do you see with this pull request ?

marshallpierce:
I think we should be sure that we actually intend to support serialization if we add code that would give the reader the impression that it is supported.

I suspect that serializability is not actually particularly useful or relevant for these classes.

ryantenney:
I don't think that change was accidental, as most Jackson classes implement Serializable.  I agree that since we don't support serialization we shouldn't give the impression that we do, but this change is limited to a couple private classes, which implement Serializable whether or not we make this change and which don't hold references to any non-serializable objects. I would agree that these classes are unlikely to be serialized, but this change doesn't hurt and is technically correct.

AhmedDaraz:
Thanks Ryan

PR #896, opened 2015-12-07
added the metrics-newrelic-reporter 3rd party

amirkibbar:
Hi, I've created a simple reporter that can report the metrics into a New Relic account. I'm not a New Relic employee, I just work with their product and thought others might find this reporter useful.


marshallpierce:
You may wish to look at https://github.com/palominolabs/metrics-new-relic (disclaimer: I'm the author), which looks like it already does pretty much the same thing, albeit with more configurability.

amirkibbar:
Hi,

You’re right it’s basically the same thing. Looks like I should’ve googled more before writing my own, would’ve saved me the trouble…

Thanks,
Amir

PR #898, opened 2015-12-14
ability to query individual health checks

osigida:
New feature: ability to query individual health checks

We have few health checks in our application. and one of them is basic dead-lock threads.
When this happens we have no choice than to restart the app.
We would like to automate it but we don't like to query all health checks and then parse the response, because it is useless to call all of them and bring some logic to the monitoring tool. Instead we would like to query individual check and act immediately if its failing.

### DIFF
--- a/metrics-servlets/src/main/java/io/dropwizard/metrics/servlets/AdminServlet.java
+++ b/metrics-servlets/src/main/java/io/dropwizard/metrics/servlets/AdminServlet.java
@@ -19,6 +19,7 @@ public class AdminServlet extends HttpServlet {
     public static final String PING_URI_PARAM_KEY = "ping-uri";
     public static final String THREADS_URI_PARAM_KEY = "threads-uri";
     public static final String HEALTHCHECK_URI_PARAM_KEY = "healthcheck-uri";
+    public static final String HEALTHCHECK_PARAM_ATTRIBUTE_KEY = "healthcheck-param";
     public static final String SERVICE_NAME_PARAM_KEY= "service-name";
 
     private static final String TEMPLATE = String.format(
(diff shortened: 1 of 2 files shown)


ryantenney:
Could you add a test to https://github.com/dropwizard/metrics/blob/master/metrics-servlets/src/test/java/io/dropwizard/metrics/servlets/HealthCheckServletTest.java?

osigida:
@ryantenney I was trying to hack and manage some tests, but the way how the code is organised it is impossible... havy refactoring for health-check servlet is required.

ingwarsw:
Our company also would benefit from such feature..

Any help needed to make it possible?

osigida:
@ingwarsw maybe if you can come up with any kind of unit test? 
we miserably failed to find any way to unit-test this feature.

ingwarsw:
Will try.. 

Do i read correctly that url for test should be in format heatlh/nameOfHealthCheck ?

ingwarsw:
@ryantenney Will it be possible to add it after adding tests and other required stuff or its not desired feature at all?

osigida:
@ingwarsw, right 
there are 2 ways to query it 
`/healthcheck` to query all registered health checks or
`/healthcheck/name_of_healthcheck` to query one of them, example: `/healthcheck/deadlocks`

ingwarsw:
Tests was actually easy :)
I refactored it a bit.. but I dont want to create other PR.

Here it is..
https://github.com/dropwizard/metrics/compare/master...ingwarsw:individual-healthchecks?expand=1

osigida:
the PR is recreated on top of 3.2-development branch 
see https://github.com/dropwizard/metrics/pull/940

Issue #814, opened 2015-06-16
Use a new Socket on each report for the Graphite scheduled reporting


adamdecaf:
We've noticed that after the fix for re-constructing connections when things fail (https://github.com/dropwizard/metrics/pull/675) that it takes 15 minutes (plus extra time based on how often you report) to see graphs again. (Or even see metrics recorded.

We've setup a custom reporter for samza jobs (due to them having their own metrics library with not graphite support). In that library we create a new socket each time, here's a copy/paste of that. (In scala)

``` scala
object Graphite extends Logging {
  def withGraphite(host: String, port: Int)(useGraphite: Graphite => Unit): Unit = {
    debug(s"Graphite connection info: $host:$port")
    var socket: Socket = null
    try {
      socket = SocketFactory.getDefault().createSocket(host, port)
      val writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), "UTF-8"))
      useGraphite(new Graphite {
        def hyphenate(s: String): String = if (s != null) s.replaceAll("""\s+""", "-") else ""
        override def send(name: String, value: String, time: Long): Unit = {
          writer.write(hyphenate(name))
          writer.write(' ')
          writer.write(hyphenate(value))
          writer.write(' ')
          writer.write(time.toString)
          writer.write('\n')
          writer.flush()
        }
      })
    } catch {
      case t: Throwable => error(s"Error writing to Graphite at $host:$port", t)
    } finally {
      if (socket != null) socket.close()
    }
  }
}
```

As you can see we are just creating a new socket each time. This has worked out well for us because since we only report once a minute (no high frequency reporting is needed for us) it's not a worry to re-create a socket on each report.

(The top dots are where we saw metrics, the bottom is samza always reporting.)

![](https://i.imgur.com/ZmaaVvk.png)

So, would it be accepted if we patched dropwizard/metrics to (probably over some threshold, anything under ~30s might want to maintain the existing use case, create a new socket on each send? Or does anyone have other options? We looked at lowering the [`SO_TIMEOUT`](https://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#getSoTimeout%28%29) as well, as right now it seems to be infinite. We have not yet tried that to see what difference it makes though.

calumlean:
Hello all,

We've had a problem with 3.1.1 though an Amazon ELB.
The symptom is that once the JVM starts, we get a few metrics delivered, but then no more.
[...]

PR #899, opened 2015-12-18
#814 always close socket

osigida:
When reporter writes to the socket and it throws an exception, then we loose points. 
So the solution is to close socket every time and open it before we need to write.

Tested in AWS cloud, now an app is able to send data to graphite.
fix for #814

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
@@ -1,5 +1,8 @@
 package com.codahale.metrics.graphite;
 
+import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


akobiakov:
:+1:

kleewho:
:+1:

chernetsov:
Why is this still not merged?

patrox:
@jplock, @ryantenney - can you please share an estimate when this can be merged and released ?
As we have to maintain a separate fork of `dropwizard-metrics` with the above fix,
which is a bit painful.

I can rebase it - as it seems that it got a bit outdated since it has been published,
but we would like to have it merged - as it seems that not only we are affected.

arteam:
Merged as #1036. Thank for your contribution!

PR #900, opened 2015-12-22
Add log4j2 xml-based config support

randomstatistic:
This allows the use of the log4j2 InstrumentedAppender without needing code-based config. ie:

    <?xml version="1.0" encoding="UTF-8"?>
    <Configuration status="INFO" name="log4j2-config" packages="com.codahale.metrics.log4j2">
        <Appenders>
            <MetricsAppender name="metrics" registryName="shared-metrics-registry"/>
        </Appenders>
        <Loggers>
            <Root level="INFO">
                <AppenderRef ref="metrics" />
            </Root>
        </Loggers>
    </Configuration>

For the most part, this was simply a matter of adding the right log4j2 plugin hooks. 

However, I ended up adding a new constructor because the concept of the appender name and the metrics prefix were tightly coupled. This was an irritation because the appender name is used as the logger reference in the XML config.
Also, since the test I added was actually initializing a logging system, I removed the test parallelism for the metrics-log4j2 suite to avoid potential race conditions among the two test files.

Please note that this is a patch against 3.1-maintenance. I wasn't sure where most development is occurring and what the best place to put it would be.

### DIFF
--- a/docs/source/manual/log4j.rst
+++ b/docs/source/manual/log4j.rst
@@ -33,3 +33,19 @@ For log4j 2.x:
     Configuration config = context.getConfiguration();
     config.getLoggerConfig(LogManager.ROOT_LOGGER_NAME).addAppender(appender, level, filter);
     context.updateLoggers(config);
+
+You can also use standard log4j2 configuration, via plugin support:
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="UTF-8"?>
+    <Configuration status="INFO" name="log4j2-config" packages="com.codahale.metrics.log4j2">
+    <Appenders>
+        <MetricsAppender name="metrics" registryName="shared-metrics-registry"/>
+    </Appenders>
+    <Loggers>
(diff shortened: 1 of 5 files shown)


randomstatistic:
Hm, apparently my change to the documentation caused the ExponentiallyDecayingReservoirTest to fail, and only on Java7. That... seems unlikely.
Are the tests flaky on Travis/3.1-maintenance?

ryantenney:
Some tests are timing dependent and can fail randomly. Can you rerun the build? If not I'll rerun it tomorrow.

randomstatistic:
My phone and I can't find how to trigger that without pushing a commit. I can do that, but not until, well, probably tomorrow.

clenart:
would love to see this make it in - it's awkward to mix xml based logger config + InstrumentedAppender

ryantenney:
@randomstatistic Would you mind submitting a pull request to the master branch as well?

PR #901, opened 2015-12-22
Improved javadoc for register(String, T extends Metric) API

aparnachaudhary:


metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
```diff
@@ -88,7 +76,8 @@ protected MetricRegistry(ConcurrentMap<MetricName, Metric> metricsMap) {
     }
 
     /**
-     * Given a {@link Metric}, registers it under the given name.
+     * Given a {@link Metric}, registers it under the given name. If the provided metric is a {@link MetricSet} then all
+     * the metrics hold by MetricSet are registered using {@link #registerAll(MetricName, MetricSet)}.
```
dpursehouse, line 80: s/hold/held/

PR #903, opened 2015-12-30
Organize third-party doc for better readability

raskasa:
Also:

* converted all entries to proper sentences.
* updated a few old links.

### DIFF
--- a/docs/source/manual/third-party.rst
+++ b/docs/source/manual/third-party.rst
@@ -7,36 +7,46 @@ Third Party Libraries
 If you're looking to integrate with something not provided by the main Metrics libraries, check out
 the many third-party libraries which extend Metrics:
 
-* `metrics-signalfx <https://github.com/signalfx/signalfx-java>`_ provides a reporter to send data to `SignalFx <http://www.signalfx.com/>`_
-* `metrics-librato <https://github.com/librato/metrics-librato>`_ provides a reporter for `Librato Metrics <https://metrics.librato.com/>`_, a scalable metric collection, aggregation, monitoring, and alerting service.
-* `metrics-spring <https://github.com/ryantenney/metrics-spring>`_ provides integration with Spring
-* `sematext-metrics-reporter <https://github.com/sematext/sematext-metrics-reporter>`_ provides a reporter for `SPM <http://sematext.com/spm/index.html>`_.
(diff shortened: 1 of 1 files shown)


raskasa:
1. Should we remove `metrics-influxdb` now that there is a first-class module available?
2. There are currently two histogram entries - which one should be deleted?
3. Are `camel-metrics` and `MetricsCatcher` under the correct sections?
3. In general, is this organization good enough?  (i.e. better section names wanted, entries under the incorrect section, etc.)

docs/source/manual/third-party.rst
```diff
@@ -7,36 +7,47 @@ Third Party Libraries
...
-* `metrics-spring <https://github.com/ryantenney/metrics-spring>`_ provides integration with Spring
-* `sematext-metrics-reporter <https://github.com/sematext/sematext-metrics-reporter>`_ provides a reporter for `SPM <http://sematext.com/spm/index.html>`_.
-* `wicket-metrics <https://github.com/NitorCreations/wicket-metrics>`_ provides easy integration for your `Wicket <http://wicket.apache.org/>`_ application.
+Instrumented Libraries
+~~~~~~~~~~~~~~~~~~~~~~
+
+* `camel-metrics <https://github.com/InitiumIo/camel-metrics>`_ provides component for your `Apache Camel <https://camel.apache.org/>`_ route.
+* `hdrhistogram-metrics-reservoir <https://bitbucket.org/marshallpierce/hdrhistogram-metrics-reservoir>`_ provides a Metrics reservoir implementation backed by `HdrHistogram <http://hdrhistogram.org/>`_.
+* `hdrhistogram-metrics-reservoir <https://bitbucket.org/marshallpierce/hdrhistogram-metrics-reservoir>`_ provides a Histogram reservoir backed by `HdrHistogram <http://hdrhistogram.org/>`_.
```
marshallpierce: this one has a slightly more useful description than the one on line 14, I think, so let's keep this one and ditch that one

raskasa: Fixed.

ryantenney:
Influxdb isn't a first party module until v4, so we'll keep it in here for now.
camel-metrics is in the right place, MetricsCatcher isn't, but it would be in a category all its own. It's fine there for now
This organization is already far better than what we had, and we can make further improvements in the future.  Thanks for doing this!

PR #905, opened 2016-01-18
Export boolean values to graphite as 1/0

swehner:
When exporting metrics from hystrix to graphite (https://github.com/Netflix/Hystrix using  https://github.com/Netflix/Hystrix/tree/master/hystrix-contrib/hystrix-codahale-metrics-publisher) we've seen that there are some interesting properties  like isCircuitBreakerOpen that are not exported to Graphite.
These properties are boolean values. We think it'd make sense to export them as 0/1 so they can be graphed alongside some other values.


jplock:
Can you add test for this?

swehner:
Hey,

Sorry for the delay - I've added a testcase for this and rebased the branch on top of the master.
Cheers

ryantenney:
Thanks!

PR #907, opened 2016-01-19
Thread name should be quoted (as jstack does)

Fuud:
JStack puts thread name in quotes and IntelliJ Idea can parse (via Analyze Stacktrace) such format.


Fuud:
any updates on this?

arteam:
Looks like a reasonable improvement for me.

PR #908, opened 2016-01-19
Allow set custom prefix Jetty QueuedThreadPool's metrics

dagi:
Prior to this fix all Jetty QueuedThreadPool's metrics are prefixed
by the class. The fix allows set the metric's name prefix. The origin value is
used as the default value because of backward compatibility. We don't want to
arbitrary chnage the name of metrics once existing users swith to new version of
the library.


arteam:
This was contributed to the 3.2.x branch in #947. But we can use the unit tests from your pull requests.
Applied to the 3.2-development branch as 240fed4. Thank you.

PR #910, opened 2016-01-20
Add basic sanitization for default CsvFileProvider

fzakaria:
The current sanitization offered by the default CsvFileProvider does nothing.
Although this may be acceptable in most cases there are some known bad filename chars that we can restrict that cover
99% of the dropwizard user base (windows, unix, linux).

The forward slash '/' is illegal in all three OS and should be at a minimum sanitizied. The forward slash is a common
character in metric names if people are constructing metric names from URIs.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/FixedNameCsvFileProvider.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/FixedNameCsvFileProvider.java
@@ -13,6 +13,9 @@ public class FixedNameCsvFileProvider implements CsvFileProvider {
     }
 
     private String sanitize(MetricName metricName) {
-        return metricName.getKey();
+        //Forward slash character is definitely illegal in both Windows and Linux
+        //https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
+        final String sanitizedName = metricName.getKey().replaceFirst("^/","").replaceAll("/",".");
+        return sanitizedName;
     }
 }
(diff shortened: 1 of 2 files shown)


fzakaria:
is there a way to re-do the travis ? I ran `mvn verify` locally with JDK8 and it passes.
Perhaps it was transient.

ryantenney:
Just restarted it, more than likely transient, this happens fairly frequently.

fzakaria:
@ryantenney  thanks.

Is it transient locally as well? I can see if i can reproduce & file issue

jplock:
This looks good to me

ryantenney:
Let's backport this to 3.2 as well.

fzakaria:
sweet first contribution to metrics.
Long time user/fan.

ps: The recent artifact groupid change without changing the actual package names caused us a lot of headache at work due to our large codebase since there were some API compatibility - but love the lib.

ryantenney:
Thanks! I understand and regret the trouble that caused. Unfortunately that was the best available option.

fzakaria:
Yea no worries - testament to how good the library is the willingness to continue using it even with these *minor* hiccups.

Curious: Why weren't the package names changed with the groupId / artifactId (same time) ?
There might be some other factors that are clearly unknown to me, but definitely interesting to hear.

ryantenney:
Version 4 will be a breaking change, and seeing as metrics has already had 1 package name change I wanted to avoid breaking backwards compatibility again just for some minor bug fixes to v3.

PR #911, opened 2016-01-28
Add statistics metric

tkp1n:
Add a statistics metric to keep track of values which are strictly monotonically nondecreasing.
Neither Histogram nor Counter support setting new values.
Statistics fills this gap.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/ConsoleReporter.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/ConsoleReporter.java
@@ -167,6 +167,7 @@ public class ConsoleReporter extends ScheduledReporter {
     @Override
     public void report(SortedMap<MetricName, Gauge> gauges,
                        SortedMap<MetricName, Counter> counters,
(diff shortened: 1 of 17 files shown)


thattolleyguy:
How is this different than a counter besides that it's non decreasing?

ryantenney:
@tkp1n `Gauge` already supports this functionality, though I appreciate your contribution!

tkp1n:
With a Gauge you provide a callback which is evaluated when the gauge is measured.
I'd like to be able to explicitly set the value (like a Counter#inc()). To have a metric for values which I receive from an external source and which are unfortunately nondecreasing.

ryantenney:
From my reading of the code, `Statistic` doesn't enforce that the values are nondecreasing.
As `Gauge` is an interface you can create a `Statistic` or `SettableGauge` class which implements it:

```java
public class Statistic implements Gauge<Long> {
    private Long value;
    public Long getValue() {
        return value;
    }
    public void setValue(Long value) {
        this.value = value;
    }
}
```

Issue #890, opened 2015-11-19
Document thread-safety


cowwoc:
I have read unofficial forum posts which hint that parts (or all) of the Metrics library is thread-safe but found no official documentation to back this up.

Please update the class-level Javadoc to indicate which classes are thread-safe. I assume that any classes that does not promise to be thread-safe, is not.

randomstatistic:
+1
I spent some time digging through the code in order to satisfy myself that my highly multi-threaded usage based on SharedMetricsRegistries would work out ok.

ryantenney:
The library is designed for use in large multithreaded applications, thread safety is of paramount importance. As such all classes are thread safe. I don't have the time to mark all classes as such (and I'm not sure I want to clutter the javadocs in that way). I would consider a 1-line addition to the documentation in order to clarify this issue.

ryantenney:
Resolved by #913

PR #913, opened 2016-02-05
Update Metric Registries section of the docs

randomstatistic:
Here's a thought on some minor tweaks to the documentation.

I couldn't find any references to SharedMetricsRegistries in the docs, and it seems like a natural thing to use in many cases.
I was also thinking of this in the context of Issue #890, and that it might be a good place to insert that mention about the thread-safe design goal.

I removed most of the bit about multiple registries in a JVM. It never made much sense to me since registry instances don't have names, except in the context of SharedMetricsRegistries.


ryantenney:
Thank you!

PR #915, opened 2016-02-09
Adds a link to a third-party reporter for Hadoop Metrics2.

joshelser:
Just a link to a reporter which enables metrics collected by Dropwizard to be pushed into the Hadoop Metrics2 system.


jplock:
Thanks!

PR #921, opened 2016-02-24
Format Number and Map instances for graphite Gauge values.

neilcronin:
Fall back to java.lang.Number interface for publishing values to graphite.  This allows alternate number types to be used as gauges, such as guava's AtomicDouble .

Issue: #922

### DIFF
--- a/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporter.java
@@ -294,35 +294,24 @@ public class GraphiteReporter extends ScheduledReporter {
     }
 
     private String format(Object o) {
-        if (o instanceof Float) {
-            return format(((Float) o).doubleValue());
-        } else if (o instanceof Double) {
-            return format(((Double) o).doubleValue());
-        } else if (o instanceof Byte) {
-            return format(((Byte) o).longValue());
-        } else if (o instanceof Short) {
-            return format(((Short) o).longValue());
-        } else if (o instanceof Integer) {
-            return format(((Integer) o).longValue());
-        } else if (o instanceof Long) {
-            return format(((Long) o).longValue());
-        } else if (o instanceof Number) {
-            return format(((Number) o).longValue());
+        String value = null;
+        if (o instanceof Number) {
+            Number n = (Number) o;
+            if (o instanceof Float || o instanceof Double) {
+                // the Carbon plaintext format is pretty underspecified, but it seems like it just wants
+                // US-formatted digits
+                value = String.format(Locale.US, "%2.2f", n.doubleValue());
+            } else {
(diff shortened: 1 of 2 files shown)


ryanrupp:
I don't use Graphite but would you want to fall back to Number.doubleValue() as a catch all, keeps the precision then on the example you had with AtomicDouble with the downsides though that it would represent things like AtomicLong (or other custom integral types) as a double and also precision issues above 2^53. Looks like the format truncates down to 2 digits though anyway I guess so not sure how much it matters.

Anyway, if Number falls back to longValue() you can drop the else ifs for all the integral types (byte, short etc.) since they all subclass Number.

neilcronin:
Yup, I agree that formatting code could be simplified, I did that in the latest commit.  Also you are right that truncation happens anyway, so this is not introducing any new loss of precision.

neilcronin:
Hi @jplock, any chance of getting this merged?  We've been running with our fork in production for almost 6 months without any issue, and I'd love to move to a standard 4.0 jar.  

Thanks!
-neil

ryantenney:
I have concerns about this. Will elaborate a little later.

neilcronin:
Hi @ryantenney, what are your concerns?  

thanks! 
-neil

neilcronin:
Hi @ryantenney Can you explain why you closed this without any comment?  The PR fixes a couple obvious issues.

ryantenney:
@neilcronin apologies, I tried renaming the master branch to 4.0-development and ended up closing all PRs in the process.

neilcronin:
@ryantenney LOL.  no worries.  still waiting on your concerns, though.

ryantenney:
I'm sorry about that! The past month has been absolutely crazy for me.

My primary concern is that this only changes the behavior of the graphite reporter. If we change this for the Graphite reporter, we should really change it for all first-party reporters, and even still we'd be unable add this behavior to third-party reporters.

Then taking into consideration that v4 will allow metrics to define arbitrary attributes (just as a meter currently defines a 'p999' attribute, a gauge would be able to define a 'foo' attribute), I'm sorry to say it doesn't make sense to add this feature.

neilcronin:
The other reporters that I've used deal with these types fine, it's only the graphite reporter that silently does nothing with certain number types.  

If adding Map as a reportable type is redundant with a different feature v4, that's fine, I understand that.  

But the v4 branch of Graphite reporter still silently does nothing if you pass in a high level Number type.  That's the bug I was fixing.  Graphite reporter should have a fallback for Numbers, and it shouldn't silently not do anything with types it doesn't understand.

neilcronin:
This is the portion that I believe is still a bug and is relevant to v4:

https://github.com/dropwizard/metrics/pull/921/files#diff-72486b088eeb4b470f31907cdcd0a916R310

ryantenney:
Sure I'd accept that PR.

neilcronin:
--> https://github.com/dropwizard/metrics/pull/986

PR #923, opened 2016-02-26
generics fix in ScheduledReporter and depending classes

gitsnaf:
Hi!
I tried to create a class derived from ScheduledReporter. Since I want to have clean code and no warnings, I was not able to do it, because the gauge interface is defined with a type parameter "Gauge<T>" whereas the abstract method "report" in ScheduledReporter.java only uses the raw type in it's method signature.
I digged into the code and the use of generics there is quite tricky sometimes. My little fix is no perfect solution, but I hope it makes it a little better than before.

Regards,
Sebastian

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
@@ -76,7 +76,7 @@ public class MetricRegistry implements MetricSet {
      */
     protected MetricRegistry(ConcurrentMap<MetricName, Metric> metricsMap) {
         this.metrics = metricsMap;
-        this.listeners = new CopyOnWriteArrayList<MetricRegistryListener>();
+        this.listeners = new CopyOnWriteArrayList<>();
     }
 
     /**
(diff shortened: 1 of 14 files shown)


metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
```diff
@@ -271,8 +271,13 @@ public void removeListener(MetricRegistryListener listener) {
      * @param filter    the metric filter to match
      * @return all the gauges in the registry
      */
-    public SortedMap<MetricName, Gauge> getGauges(MetricFilter filter) {
-        return getMetrics(Gauge.class, filter);
+    public SortedMap<MetricName, Gauge<?>> getGauges(MetricFilter filter) {
+        final SortedMap<MetricName, Gauge> metrics = getMetrics(Gauge.class, filter);
+        final SortedMap<MetricName, Gauge<?>> result = new TreeMap<>();
```
arteam: Could we avoid creating a copy of `metrics`? I guess `javac` doesn't allow to cast `SortedMap<MetricName, Gauge>` to `SortedMap<MetricName, Gauge<?>>`, but I would like to avoid creating an additional copy just because of compiler peculiarities.

PR #931, opened 2016-04-05
Negative and large tick value fixes

mozotom:
Collision buffer will cause overflow at 417 days of server uptime (not
JVM). Also it would take 1,000,000 requests in the same millisecond to
cause loss of accuracy of 1 ms, so this buffer is not needed and given
that it causes problems, should be removed.
Also added test case for large negative values of tick (rarely happens,
but according to nanoTime api, it can)

### DIFF
--- a/metrics-core/src/test/java/com/codahale/metrics/SlidingTimeWindowReservoirTest.java
+++ b/metrics-core/src/test/java/com/codahale/metrics/SlidingTimeWindowReservoirTest.java
@@ -24,6 +24,39 @@ public class SlidingTimeWindowReservoirTest {
     }
 
     @Test
+    public void storesMeasurementsWithDuplicateTicksTresholdNeg() throws Exception {
+        when(clock.getTick()).thenReturn(Long.MIN_VALUE / 256);
+        SlidingTimeWindowReservoir reservoir = new SlidingTimeWindowReservoir(10, TimeUnit.NANOSECONDS, clock);
+
+        reservoir.update(1);
+        reservoir.update(2);
+
+        assertThat(reservoir.getSnapshot().getValues())
+                .containsOnly(1, 2);
+    }
+
+    @Test
+    public void storesMeasurementsWithDuplicateTicksMinNeg() throws Exception {
+        when(clock.getTick()).thenReturn(Long.MIN_VALUE);
+        SlidingTimeWindowReservoir reservoir = new SlidingTimeWindowReservoir(10, TimeUnit.NANOSECONDS, clock);
+
+        reservoir.update(1);
+        reservoir.update(2);
+
+        assertThat(reservoir.getSnapshot().getValues())
+                .containsOnly(1, 2);
+    }
+
+    @Test
+    public void storesMeasurementsWithTresholdTicks() throws Exception {
+        when(clock.getTick()).thenReturn(TimeUnit.DAYS.toNanos(416));
+        reservoir.update(1);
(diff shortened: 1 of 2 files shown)


ryantenney:
While this does, in a way, remedy the issue you're experiencing, it results in inaccurate data.  The collision buffer exists to allow the SlidingTimeWindowReservoir to record up to 256 values per nanosecond.  Without the collision buffer, subsequent updates in a given nanosecond will result in the tick being incremented:

update 1, ns = 1, lasttick = 0, tick = 1
update 2, ns = 1, lasttick = 1, tick = 2
update 3, ns = 1, lasttick = 2, tick = 3
update 4, ns = 1, lasttick = 3, tick = 4
update 5, ns = **2**, lasttick = 4, tick = **5**

It doesn't take very long before the tick drifts significantly and is no longer representative of the instant of the update. Granted the same could conceivably happen with the collision buffer in place, but that would require 257 updates per nanosecond, not 2.

mozotom:
Collision will not happen, because we constantly check if "tick - oldTick > 0" in line 71. While drift might seem large on nanosecond level, it would take 1,000,000 requests per millisecond to affect it on millisecond level.

ryantenney:
If that's an acceptable loss of precision for you then that's fine, however I'm not prepared to fix this issue by breaking something else.  I have a fix ready in a branch (see https://github.com/dropwizard/metrics/tree/sliding-time-window-overflow), and I'm aiming to release it soon.

mozotom:
We've had your fix in live and we just started having OOM when JVM was alive during 417 days of up time on the server. I've added tests to detect that issue. Please run those tests against your code and you'll see where it fails

metrics-core/src/test/java/com/codahale/metrics/SlidingTimeWindowReservoirTest.java
```diff
@@ -24,6 +24,39 @@ public void storesMeasurementsWithDuplicateTicks() throws Exception {
...
+
+        reservoir.update(1);
+        reservoir.update(2);
+
+        assertThat(reservoir.getSnapshot().getValues())
+                .containsOnly(1, 2);
+    }
+
+    @Test
+    public void storesMeasurementsWithTresholdTicks() throws Exception {
```
mozotom: This test detects overflow on 417th day of server uptime (we've seen this in live and have heap dumps showing it)

metrics-core/src/test/java/com/codahale/metrics/SlidingTimeWindowReservoirTest.java
```diff
@@ -24,6 +24,39 @@ public void storesMeasurementsWithDuplicateTicks() throws Exception {
     }
 
     @Test
+    public void storesMeasurementsWithDuplicateTicksTresholdNeg() throws Exception {
+        when(clock.getTick()).thenReturn(Long.MIN_VALUE / 256);
```
mozotom: since nanoTime can be arbitrary number (even negative per definition), let's test all possible values - this one breaks current solution on sliding-time-window-overflow branch

DardenneLanding:
Ryan, how about making the buffer configurable, so users can choose the size of it? That way they'll be able to trade precision for longer periods without restarting physical node?

mozotom:
Ryan, how about making buffer size configurable, so users can choose between precision and longer period of time without having to reboot physical node to reset up time?

PR #932, opened 2016-04-05
Making collision bit width configurable,

mozotom:
Collision buffer will cause overflow at 417 days of server uptime (not
JVM). Making it configurable, so users can choose between precision and
longer period of time without having to reboot physical node to reset
up time


mozotom:
@ryantenney would you consider making the buffer bit width configurable?

ryantenney:
That's a reasonable. I'll look it over tomorrow.

ryantenney:
Can you look into the build failure?

mozotom:
@ryantenney I'm not sure why this test fails, but it fails the same way without my changes. Those changes should be fully backwards compatible with the same default values, so there should be no impact to existing tests. Once you get this branch build without my changes, it should build with those changes as well. If not, I'll review and fix any issues.

ryantenney:
Okay I'll look into what's failing in that branch.

PR #933, opened 2016-04-08
Support for publishing metrics of BigInteger and BigDecimal type

nurkiewicz:


jplock:
Seems like a simple additional to me, thanks!

ryantenney:
Actually I have a problem with this. I don't mind supporting BigInteger and BigDecimal, but this is the wrong way to implement it. BigDecimal -> long discards the fractional part, and both conversions return only the low order 64-bits. Graphite stores everything as doubles, so I think we ought to be converting both BigInteger and BigDecimal to doubles.

jplock:
Ahh ok, I reverted the commit, but we'll need another PR to address @ryantenney's comments @nurkiewicz.  Sorry about that.

Issue #663, opened 2014-09-16
Making HealthCheck.Result class extensible


hmpatel:
I want to extend the Result class to enhance the health checks module for providing additional details. For e.g. I would add another health state in there - MAINTENANCE to indicate my servers are in the maintenance mode and also provider an optional placeholder Map to provide any additional details. Currently, Result class have only healthy and unhealthy states and furthermore, there is no way to provide additional info other than message. 

So, I tried to extend the HealthCheck class and add these details in there. However, due to absence of default no arg constructor in Result class, it's not possible to subclass Result.

So would like to propose this change or get the feedback.

Thanks!

JonMR:
I'd also like to see this happen.  I'd be open to creating a PR as well.

In our case, the two states of healthy and unhealthy aren't enough.  For us, unhealthy means drop the service from the pool.  I'd like a third state that indicates there's an issue but the service is still able to handle traffic.

Specific example: We have a pool of memcache servers that the server talks to.  If it loses connectivity with one of those server, it's not a big deal, but something to be alerted and investigated.

I suppose we could write a parser for the output, but I'd rather not have to do that for each case.

mjvandel:
+1 yes, please! It would make it possible to add JSON nodes to the response with extra details

arnaud-deprez:
+1 too ! :-)

zaphinath:
+1

arnaud-deprez:
I think it would be possible to allow custom details by adding a Map<String,Object> as spring-boot-actuator does without breaking the API.
Like that, we wouldn't be forced to wait for version 4.0.0.
What do you think ?

ryantenney:
Branch off 3.1-maintenance and I'd be happy to look at a pull request that adds that support for a 3.2 release.

arnaud-deprez:
Okay, I'll take a look asap. Keep in touch.

arnaud-deprez:
So if I get it right, master is for 4.0.0.
I suppose I need to base my code on 3.1-maintenance and make a pull request on a new branch. Correct ?

dhay:
Our use case is this:  We use the health check framework to inform our on-call alerting.  As is, we basically have two states: UP or DOWN.  We wanted a third state, WARN, to indicate a problem with our application, but not worthy of waking someone up in the middle of the night.

So being able to have additional HealthCheck result types would be very helpful.

cdeszaq:
[...]

PR #934, opened 2016-04-08
METRICS-#663: add support for custom details in HealthCheck.Result

arnaud-deprez:
This can be a solution for #663 without breaking the api. Can be merged in a 3.2 version.
Regards

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
@@ -1,15 +1,20 @@
 package com.codahale.metrics.health;
 
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
 /**
  * A health check for a component of your application.
  */
 public abstract class HealthCheck {
     /**
-     * The result of a {@link HealthCheck} being run. It can be healthy (with an optional message)
-     * or unhealthy (with either an error message or a thrown exception).
+     * The result of a {@link HealthCheck} being run. It can be healthy (with an optional message and optional details)
+     * or unhealthy (with either an error message or a thrown exception and optional details).
      */
     public static class Result {
-        private static final Result HEALTHY = new Result(true, null, null);
+        private static final Result HEALTHY = new ResultBuilder(true).build();
         private static final int PRIME = 31;
 
         /**
@@ -28,7 +33,7 @@ public abstract class HealthCheck {
          * @return a healthy {@link Result} with an additional message
          */
         public static Result healthy(String message) {
-            return new Result(true, message, null);
+            return new ResultBuilder(true).withMessage(message).build();
         }
 
         /**
@@ -52,7 +57,7 @@ public abstract class HealthCheck {
          * @return an unhealthy {@link Result} with the given message
          */
         public static Result unhealthy(String message) {
-            return new Result(false, message, null);
+            return new ResultBuilder(false).withMessage(message).build();
         }
 
         /**
@@ -73,20 +78,50 @@ public abstract class HealthCheck {
          * Returns an unhealthy {@link Result} with the given error.
          *
          * @param error an exception thrown during the health check
-         * @return an unhealthy {@link Result} with the given error
+         * @return an unhealthy {@link Result} with the given {@code error}
          */
         public static Result unhealthy(Throwable error) {
-            return new Result(false, error.getMessage(), error);
+            return new ResultBuilder(error).build();
+        }
+
+		/**
+         * Returns an healthy {@link ResultBuilder}
+         *
+         * @return an healthy {@link ResultBuilder}
+         */
+        public static ResultBuilder healthyBuilder() {
+            return new ResultBuilder(true);
+        }
+
+        /**
+         * Returns an unhealthy {@link ResultBuilder}
+         *
+         * @return an unhealthy {@link ResultBuilder}
+         */
+        public static ResultBuilder unHealthyBuilder() {
+            return new ResultBuilder(false);
+        }
+
+        /**
+         * Returns an unhealthy {@link ResultBuilder} with the given {@code error}
+         *
+         * @param error an exception thrown during the health check
+         * @return an unhealthy {@link ResultBuilder}
+         */
+        public static ResultBuilder unHealthyBuilder(Throwable error) {
+            return new ResultBuilder(error);
         }
 
         private final boolean healthy;
         private final String message;
         private final Throwable error;
+        private final Map<String, Object> details;
 
-        private Result(boolean isHealthy, String message, Throwable error) {
-            this.healthy = isHealthy;
-            this.message = message;
-            this.error = error;
+        private Result(ResultBuilder builder) {
+            this.healthy = builder.healthy;
+            this.message = builder.message;
+            this.error = builder.error;
+            this.details = Collections.unmodifiableMap(builder.details);
         }
 
         /**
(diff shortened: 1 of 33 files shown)


ryantenney:
Thanks, I'll take a look!

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
```diff
@@ -146,11 +185,83 @@ public String toString() {
...
+     * or unhealthy (with optional message, error and detail)
+     */
+    public static class ResultBuilder {
+        private boolean healthy;
+        private String message;
+        private Throwable error;
+        private Map<String, Object> details;
+
+        protected ResultBuilder(boolean healthy) {
+            this.details = new LinkedHashMap<String, Object>();
```
ryantenney: Details should default to null

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
```diff
@@ -73,20 +78,50 @@ public static Result unhealthy(String message, Object... args) {
...
+        public static ResultBuilder healthyBuilder() {
+            return new ResultBuilder(true);
+        }
+
+        /**
+         * Returns an unhealthy {@link ResultBuilder}
+         *
+         * @return an unhealthy {@link ResultBuilder}
+         */
+        public static ResultBuilder unHealthyBuilder() {
```
ryantenney: Should be `unhealthy` for consistency

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
```diff
@@ -1,15 +1,20 @@
...
 public abstract class HealthCheck {
     /**
-     * The result of a {@link HealthCheck} being run. It can be healthy (with an optional message)
-     * or unhealthy (with either an error message or a thrown exception).
+     * The result of a {@link HealthCheck} being run. It can be healthy (with an optional message and optional details)
+     * or unhealthy (with either an error message or a thrown exception and optional details).
      */
     public static class Result {
-        private static final Result HEALTHY = new Result(true, null, null);
+        private static final Result HEALTHY = new ResultBuilder(true).build();
```
ryantenney: Do we need to use the Builder internally? Can we keep the old constructor for use internally?

ryantenney:
I've made a couple comments, but it looks good otherwise.  I'm sorry to bother you with this, but could you reopen this pull request against the new 3.2-development branch?

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
```diff
@@ -146,11 +185,83 @@ public String toString() {
...
+        }
+
+		/**
+         * Add an optional detail
+         *
+         * @param key a key for this detail
+         * @param data an object representing the detail data
+         * @return this builder with the given detail added
+         */
+        public ResultBuilder withDetail(String key, Object data) {
```
ryantenney: Maybe details should be 'attributes' or 'properties' instead?  And I think this method should be `add(Detail/Attribute/Property)` </bikeshedding>

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: Not sure what is the goal to have null in default ? 
If you take a look in the HealthCheckModule, I only serialize details when it's not empty.
But if you insist, I can put it to null by default

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: I can do that for unhealthy but not for healthy as it has already a method call healthy(). But I agree, I'm not a big fan of this method name so I'm open to suggestion

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: The idea was to use the same Builder everywhere but and have only one single constructor which takes the builder as parameter. I think it give us more scalability and a better encapsulation. For example if we would like to add a new specific field, we can do it and add a default value in the builder. So if you don't use it, it will have the default value without the need to change this code.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: I use the same name as spring does in its module spring-actuator, I'm not closed to it. What do you prefer: details ? attributes ? properties ? Then I can adapt the method to be add*** instead of with***

arnaud-deprez:
I've made a couple of comments too where I need your suggestion/decision for changes you request. Please keep me in touch, I'm open for each changes.
I'll rebase my changes to 3.2 and make a new pull-request.
Regards

arnaud-deprez:
I propose to continue the discussion here: #943

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
ryantenney: The problem I have with it is that it allocates memory that we really don't need to allocate. Yes, it's a small amount of memory (my math says ~104 bytes), but it's also very easy to avoid doing it.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
ryantenney: Sorry, what I meant was should be `unhealthBuilder`. But yea that name isn't awesome either. Maybe it's just `builder()`, the builder defaults to healthy, and calling `unhealthy()`, `healthy(false)`, or `error(Throwable)` on the builder sets it to unhealthy.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
ryantenney: I think we can be pretty sure that we won't add any new fields to Result after this. My concern again is that we're allocating additional memory that will require later garbage collection.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
ryantenney: I'm actually fine with the method name now.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: Yeah I see your point but my goal here was to avoid NPE exception for the end user.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: That's definitely a better idea ! I propose to use builder and then an healthy(), unhealthy() and unhealthy(Throwable error) in the builder. Like that the api avoid you to have an healthy builder with an exception.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: Ok fine, I'll change it.

metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
arnaud-deprez: ok :)

PR #936, opened 2016-04-11
Graphite: Replace names with non-alphanumeric chars with '-'

nickbabcock:
Fix for #637, that supersedes #731 

The fix won't allow any characters that will interfere with Graphite's display or functions.

Instead of using fancy regex, this PR uses code ported from Guava's CharMatcher and Apache's StringUtils.

This PR was made against master because this may break other's graphite metric names, even if some of them aren't 100% valid (eg. `foo:bar` will be logged fine, but some graphite functions will break). Any potential breakage should be able to be fixed through [whisper-merge](https://github.com/graphite-project/whisper/blob/master/bin/whisper-merge.py). If this is wanted for 3.2.0, I'd be more than happy to send another PR against that.

### DIFF
--- a/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteRabbitMQ.java
+++ b/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteRabbitMQ.java
@@ -9,14 +9,11 @@ import java.io.IOException;
 import java.net.Socket;
 import java.nio.charset.Charset;
 import java.util.concurrent.TimeoutException;
-import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


ryantenney:
Thank you! I've just created the branch 3.2-development, and it'd be great to see this make it into 3.2 as well.

ArturGajowy:
FYI this change has been reverted by #1099, sadly without referencing this PR

nickbabcock:
Yup, it was reverted because it broke everything 😄 

Things may be different now, and it may now be possible to sanitize individual portions of a metric name -- but I haven't checked.

PR #938, opened 2016-04-11
Graphite: Replace names with non-alphanumeric chars with '-'

nickbabcock:
Same as #936 but on 3.2-development branch


ryantenney:
Thanks a lot!

PR #940, opened 2016-04-12
New feature: ability to query individual health checks

osigida:
We have few health checks in our application. and one of them is basic dead-lock threads.
When this happens we have no choice than to restart the app.
We would like to automate it but we don't like to query all health checks and then parse the response, because it is useless to call all of them and bring some logic to the monitoring tool. Instead we would like to query individual check and act immediately if its failing.

thanks to @ingwarsw  for sorting out unit-tests

### DIFF
--- a/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
+++ b/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
@@ -1,10 +1,11 @@
 package com.codahale.metrics.servlets;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectWriter;
 import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.codahale.metrics.json.HealthCheckModule;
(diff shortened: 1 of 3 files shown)


metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -86,7 +88,13 @@ public void init(ServletConfig config) throws ServletException {
     @Override
     protected void doGet(HttpServletRequest req,
                          HttpServletResponse resp) throws ServletException, IOException {
-        final SortedMap<String, HealthCheck.Result> results = runHealthChecks();
+        final Map<String, HealthCheck.Result> results;
+        String name = getHealthCheckName(req);
+        if (name != null) {
+            results = Collections.singletonMap(name, registry.runHealthCheck(name));
```
ingwarsw: Actually that method throws NoSuchElementException when check is not found.
You should create method that would check if that check exists and only then call method.

If check is not existing you should return empty map so responses (all/one) would be consistent.

osigida: yep, please have a look again

PR #941, opened 2016-04-14
Create null versions of the metric types and a NullMetricRegistry that builds them

Intropy:
The purpose of this change is to allow users to code against the base metric types and registry but have the flexibility to swap in null implementations of the counters. It enables a usage pattern like:

    class Foo {
        private final MetricRegistry registry;

        Foo(MetricRegistry registry) {
            this.registry = registry;
        }

        Foo() {
            this(new NullMetricRegistry());
        }

        public void someFunction() {
            registry.meter("someFunctionCalled").mark();
        }
    }

A client of Foo can pass in a MetricRegistry if it wants metrics collected, but it also doesn't have to use or be aware of Foo's use of metrics at all.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -112,7 +112,7 @@ public class MetricRegistry implements MetricSet {
      * @return a new or pre-existing {@link Counter}
      */
     public Counter counter(String name) {
-        return getOrAdd(name, MetricBuilder.COUNTERS);
+        return getOrAdd(name, getCounterMetricBuilder());
     }
 
     /**
@@ -123,7 +123,7 @@ public class MetricRegistry implements MetricSet {
      * @return a new or pre-existing {@link Histogram}
      */
     public Histogram histogram(String name) {
-        return getOrAdd(name, MetricBuilder.HISTOGRAMS);
+        return getOrAdd(name, getHistogramMetricBuilder());
     }
 
     /**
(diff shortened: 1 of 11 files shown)


Intropy:
Is this test flaky? After I saw this failure I set up Travis on my fork. It failed, too but on jdk8 instead of 7. A few more runs with no code changes and it succeeded.

https://travis-ci.org/Intropy/metrics

ryantenney:
I like the idea, however this PR introduces new protected methods and changes a private method to protected, which I can't do for a minor release.  I'd consider it for 3.2, however v3.2.0 will more than likely be the final v3 release, and I don't want to expand the scope.

Would you consider applying this change to the `v4.0` branch?  Two notes, 1) it will take a little bit of work to adapt it to that branch, as file and package names have changed, and MetricName was introduced, 2) I want to include a capability in v4 to override how metrics are created, and while this looks like a good first step, the implementation might be wholly unrecognizable by the time v4 is released.

Thanks!

Intropy:
Sure, I'm willing to apply it on the v4.0 branch. I want to make sure of what you're saying with "the implementation might be wholly unrecognizable by the time v4 is released", though. Are you suggesting reworking it and applying it to v4.0 as v4.0 currently stands but that the implementation may be drastically changed as v4.0 evolves? Or are you saying that I should wait because if I build it for the current state of v4.0 then before the pull goes through the organization is likely to change significantly enough to require a second rewrite?

Also a note on why I closed and reopened since in retrospect it might have looked like I was pinging you. I think there may be some flakiness on the ScheduledReporterTest. I had consistent success locally before I put in the pull request, but on Travis it sometimes succeeds and sometimes fails. The first pull request resulted in a failure (https://travis-ci.org/dropwizard/metrics/jobs/123204697). Closing and reopening triggered Travis to run it again (https://travis-ci.org/dropwizard/metrics/builds/124043581).

ryantenney:
The former, rework the patch to v4 as it is currently and then it will evolve along with that release.

That's not typically the test I have problems with, but its not surprising as it is time sensitive. I think that if you authenticate with Travis it will allow you to rerun a build without closing/reopening the PR. Also feel free to @ mention me to request that I rerun it.

PR #947, opened 2016-05-06
Allow setting custom prefix in InstrumentedQueuedThreadPool

krasinski:
Hello,
I want to add ability to set custom prefix in InstrumentedQueuedThreadPool in metrics-jetty9. In similar way to InstrumentedHandler.
I've added setter because I think there should be a simple way to set the prefix when someone wants to use existing defaults. Another way would be adding some additional constructors (obviously I don't want to break the existing api) - but that would bring some other design issues.
I'm not sure what's better - what do you think about it?


arteam:
Looks good to me. I like the setter approach.

arteam:
Thanks for the contribution!

PR #948, opened 2016-05-18
Add links to more third party libs

marshallpierce:
Turns out I hadn't added all the libraries I maintain.

(Also it looks like the doc site isn't up to date with the 3.1 branch, as the live page doesn't have the entry for metrics-munin-reporter.)


ryantenney:
Thanks! I'll get that updated today once things aren't literally on fire at work 🔥

PR #950, opened 2016-05-20
Update third-party to include metrics-circonus

postwait:


ryantenney:
Thanks! I'll get this deployed later today.

PR #955, opened 2016-06-03
Improve hostname extraction for wrapped HTTP requests

phearnot:
When executing a request, HttpClient will wrap the original request in `HttpRequestWrapper` so that e.g. redirects can be handled with ease. In such cases, the wrapper request has its URI rewritten (see `org.apache.http.impl.execchain.ProtocolExec#rewriteRequestUri`): the host part is stripped, so the wrapper's `RequestLine` can no longer be used to reliably obtain the hostname.

This pull request improves `HOST_AND_METHOD` and `QUERYLESS_URL_AND_METHOD` metric naming strategies by checking if the request is actually a wrapper and trying to obtain URI from the original request.

### DIFF
--- a/docs/source/manual/core.rst
+++ b/docs/source/manual/core.rst
@@ -77,7 +77,7 @@ return the number of evictions from the cache.
 JMX Gauges
 ----------
 
-Given that many third-party library often expose metrics only via JMX, Metrics provides the
+Given that many third-party libraries often expose metrics only via JMX, Metrics provides the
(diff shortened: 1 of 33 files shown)


ryantenney:
This is really great, but could you rebase this branch PR against the 3.2-development branch? The next release of 3.1 will only be bugfixes, no functional changes.

phearnot:
@ryantenney, sure. See #961.

PR #957, opened 2016-06-09
Gauge addition should not throw exception if already in registry

souravmitra:
Just like the other metrics, gauge addition should not throw an exception when adding to the registry with the same metric name. It should return the existing Gauge.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -149,6 +149,32 @@ public class MetricRegistry implements MetricSet {
     }
 
     /**
+     * Return the {@link Gauge} registered under this name; or create and register
+     * a new {@link Gauge} if none is registered.
+     *
+     * @param name the name of the metric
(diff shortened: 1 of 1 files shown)


ryantenney:
I'm afraid this behavior doesn't make sense for gauge. For all other metric types the MetricRegistry is responsible for creating the instance, but gauge is created by the user and supplied to the MetricRegistry. We can't simply assume they will be identical.

souravmitra:
@ryantenney : Thanks for your comments.

So we in our app have a single MetricRegistry.
Now we have func A() that registers our gauge with this registry.

Since func A() is called by an external framework and on different threads, eg: Apache Storm, we start to get exceptions when their is a race between threads, and both end up trying to add the same gauge with the same name to the registry.
In meter() and timer(), the same happens but we have a getOrAdd() there, so that even though two threads attempt addition with the same name, an exception is thrown, but is caught, and the existing metric returned.

Do you think its ours is a valid use case?

souravmitra:
@ryantenney Hello, any thoughts?

PR #961, opened 2016-06-14
Improve hostname extraction for wrapped HTTP requests

phearnot:
When executing a request, HttpClient will wrap the original request in HttpRequestWrapper so that e.g. redirects can be handled with ease. In such cases, the wrapper request has its URI rewritten (see org.apache.http.impl.execchain.ProtocolExec#rewriteRequestUri): the host part is stripped, so the wrapper's RequestLine can no longer be used to reliably obtain the hostname.

This pull request improves HOST_AND_METHOD and QUERYLESS_URL_AND_METHOD metric naming strategies by checking if the request is actually a wrapper and trying to obtain URI from the original request.


ryantenney:
Thanks! I've got to take a closer look at this before merging. I'll get to it soon.

ryantenney:
@phearnot Thanks!

PR #965, opened 2016-06-20
Switched to AtomicLongFieldUpdater so it works on AppEngine

riccardoref:
I am not sure why in 3.2 and also 3.1.2 the AtomicLongFieldUpdater has been replaced with sun.misc.Unsafe again. Applied the patch made by ryantenney (commits f9b4572 and d83b989) in the old project to make it work on AppEngine.  Let me know if there are any reasons why you needed to switch it back.

Many thanks

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/Striped64.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Striped64.java
@@ -9,6 +9,7 @@
 package com.codahale.metrics;
 
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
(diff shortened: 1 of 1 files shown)


ryantenney:
There was a small problem with this PR, the `busy` field was changed from an int to a long. In my first commit (f9b4572) I inadvertently used AtomicLongFieldUpdaters for both fields, but my second commit (d83b989) fixed that.  I've applied those patches in PR #966 which I'll merge just as soon as the builds pass. Thanks!

riccardoref:
Thank you for the reply. I will use the fix #966 then. And use the 3.2 version as soon as you merge the branch with the fix. Thanks again!

Issue #970, opened 2016-06-28
Optimize UniformSnapshot construction


bentatham:
When creating a `UniformSnapshot` from a `UniformReservoir` three copies are made of the original values in the reservoir, with boxing back and forth as well. From a `SlidingTimeWindowReservoir`, there are two copies made.  

While it somewhat "dangerous", I think it would be better in both cases to pass in a raw `long[]` directly to the constructor that can be used directly by the `UniformSnapshot` so that there is only one copy of the values, and only one exchange from `Long` to `long`.

bentatham:
(looking at 3.2-development) I see there is a test case ``UniformSnapshotTest.worksWithOverestimatedCollections``.  This seems like a pretty white-box test to ensure that the constructor does not rely on the ``size()`` method of the collection at all.  I'm trying to understand where that case happens?  Is it a concurrency issue from the ``ConcurrentSkipListMap`` values being passed in from ``SlidingTimeWindowReservoir``?    Looking at the impl of the values collection it gives for ``toArray``, that is not thread-safe either, so I'm not sure what extra values this provides.

Looking at commit history, this was all done as part of a big metrics-core refactor back in 2013, so no guidance given there.

storozhukBM:
@bentatham if you're using `SlidingTimeWindowReservoir` try new drop-in replacement called `SlidingTimeWindowArrayReservoir` it should make snapshots a lot faster due to vectorized `System.arraycopy` usage.

PR #972, opened 2016-06-29
#970: optimize creation of UniformSnapshot

bentatham:
Optimize creation of ``UniformSnapshot`` with less copies and boxing to ``Long`` and ``long``.

I kept the copy of the array itself when passed into the constructor to satisfy exposing a mutable array into the object, as per the findbugs warning.  Personally, I would like to remove that also since this is all internal to the library, but I understand why it is there.

Also, see comment in #970 about the removed tests.


ryantenney:
Thanks!

Issue #609, opened 2014-07-01
Asynchronous health checking


tomfitzhenry:
Problem: I'd like my calls to `/healthcheck` to execute quickly (say, <50ms), but some of my health checks are slow (say, 500ms).

Potential solution: Execute the health check every minute, using a `ScheduledThreadPoolExecutor` or some such, store the result, and have `check()` return the latest health checks results (checking that they're not stale too).

We can achieve this with either:
1. Smart HealthCheck instances: Clients extend `HealthCheck`, creating a new method, `checkSync()`, which they execute every minute, and define `check()` to return the latest result of checkSync(). (Making sure the result isn't stale)
2. Smart HealthCheckRegistry: clients submit `HealthCheck` instances to a `AsyncHealthCheckRegistry` or some such.

(2)'s API would potentially look like: `AsyncHealthCheckRegistry#submit(mySlowCheck, 2, TimeUnit.Minutes)`, to have `mySlowCheck` executed once every two minutes.

Thoughts?

tomfitzhenry:
Two years to the day... freaky.

stockmaj:
The change for this introduced a dependency on JRE 1.7 (XXX in the date format string).  Nothing in the documentation says JRE 1.7 is required, is there any reason this can't be changed back to Z for backwards compatibility?

arteam:
I guess you meant​ to comment on #1077, right? I think it's possible to change the format to be compatible with 1.6, if it doesn't change behaviour for 1.7 and 1.8.

ryantenney:
Great catch. The issue seems to be that `XXX` formats the offset as `"-05:00"`, while `Z` formats as `"-0500"`, which while still ISO8601 compliant, is not RFC3339 compliant.

stockmaj:
I can't tell from your answer if it's OK to change it or not. If 3.2 intentionally requires 1.7, I think it should say that in the Release Notes.  If not, this needs to be changed, one way or another.

stockmaj:
@arteam When I looked at the history it looked like it came in on #609, but I do see it on #1077 also.  I don't use github a lot.  The reason I'm just asking and not submitting a pull is that I would have to do the pull from home, not from work.  If I'm going to do that, I'd like to work out the details ahead.

arteam:
I think it would be correct to change the pattern, because we still want to maintain runtime compatibility with JDK 1.6.

arkigabor:
[...]

PR #973, opened 2016-07-01
Async health check

tomfitzhenry:
For issue #609

Proposed API for users is:

```
HealthCheckRegistry registry = new HealthCheckRegistry();
ScheduledExecutorService e = new ScheduledThreadPoolExecutor(1);
HealthCheck foo = ...
AsyncHealthCheck asyncFoo = new AsyncHealthCheck(foo, 2, TimeUnit.MINUTES);
e.scheduleWithFixedDelay(asyncFoo, 0, 30, TimeUnit.SECONDS);
registry.register("foo", asyncFoo)
```

In the above example, the ScheduledExecutorService will execute the health checks with 30s delay. The health check will fail if its cached result is older than 2 minutes.

Design choices:
- Give user control over the async checking, so they can plug it into whatever executor service they have
- Reads must be quick
- Detect if the cached result is too old, and allow user to define what too old is


tomfitzhenry:
For users that are happy to cede control over the execution of the health check run() methods, perhaps `HealthCheckRegister#registerAsync(HealthCheck, int refreshFrequency, TimeUnit refreshFrequencyUnit, int int staleAge, TimeUnit staleAgeUnit)` could be added, which managed an internal ScheduledExecutorService

vbauer:
Any about this PR?... We'd like to have async-feature too

arteam:
Contributed in #1077

PR #978, opened 2016-07-06
Optimize creation of UniformSnapshot

ryantenney:
Applies #972 to master, and adds an unsafe protected constructor for use by trusted code.


ryantenney:
@bentatham Thoughts on the change in b77971a?

bentatham:
Ok with me.  I would have said never copy in the constructor, but at least
this makes it explicit.

ryantenney:
Something weird happened here and this PR was not marked as merged. It was merged though.

Issue #837, opened 2015-07-23
MetricAttribute Key


bentatham:
I'm not sure I'm using the right terminology, but here I'm using _attribute_ to refer to a specific value of a _metric_, eg _75thPercentile_ of a `Histogram`.  In practice, an _attribute_ is what gets sent out of the system to external tools like graphite.

This is a fairly large scope idea, but it seems to me that through Metrics, there is often a whole whack of methods per-metric.  This makes it virtually impossible to add any custom metrics or do any intelligent filtering of attributes.

See discussion in #835 

So the idea here is to use a _key_-based system for the metric attributes, instead of each metric having attribute-specific methods.
- this could be an `enum`, but likely better to just use some  interface like `MetricAttribute` or something to make it easier to define them in the appropriate places and allow for custom extensibility.
- each metric would have: 
  - `Object getAttribute(MetricAttribute)` method, and would throw `IllegalArgumentException` if the given attribute did not apply to that `Metric`.
  - could get fancy with generic, if desired, so that each `MetricAttribute<T>` could be more specific `<T> T getAttribute(MetricAttribute<T>)` to avoid casting and checking issues.  Obviously, all the primitive types that are typically returned would be boxed.
  - `Set<MetricAttribute> getAttributes()`
- This would then allow us to arbitrarily add custom `Metric`s to the system that could still be handled by `Reporter` by injecting a handler for each specific `MetricAttribute` to the `Reporter`. (or use some generic attribute handling, based on `<T>` in each `Reporter`).
- Each `MetricAttribute` could have a `getKey()` or some such method that could be shared among `Reporter`s (eg `p99`, `median`) so they aren't repeated everywhere;

Lots of details to be fleshed out, but hopefully the goal is clear -- to allow more extensibility.

ryantenney:
:+1: this is almost exactly (not 100% on MetricAttrbute for the name, but that's less important) what I've been planning for v4.

bentatham:
Great!  I'll get started on something in my fork.  If not ``MetricAttribute``, what name did you have in mind?  ``MetricValue``?  The terminology is certainly overlapping -- these _attributes_ are the _metrics_ in the graphite world, whereas the dropwizard _metric_ is a group of graphite _metrics_.  Obviously, I'm not suggesting we change the dropwizard definition of _metric_, but we just need to be clear in the naming and docs to avoid confusion wherever possible.

bentatham:
[...]

PR #986, opened 2016-07-14
Fall back to java.lang.Number for publishing values to graphite.

neilcronin:
And log a warning if the value to be reported cannot be parsed.


neilcronin:
Thank you.  Can you point me to the docs for the v4 feature of adding arbitrary dimensions to gauges?  I don't see that feature in the v4 code, but I may have missed it.  TIA!

-neil

ryantenney:
Not finished yet, tracking in #837. @bentatham is currently working on it.

PR #989, opened 2016-07-21
Add an overload of Timer#time that takes Runnable

stevenschlansker:
This allows you to invoke method handles or lambdas that
don't return a value easily.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/Timer.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Timer.java
@@ -105,6 +105,21 @@ public class Timer implements Metered, Sampling {
     }
 
     /**
+     * Times and records the duration of event.
+     *
+     * @param event a {@link Runnable} whose {@link Runnable#run()} method implements a process
+     *              whose duration should be timed
+     */
+    public void time(Runnable event) {
+        final long startTime = clock.getTick();
+        try {
+            event.run();
+        } finally {
+            update(clock.getTick() - startTime);
+        }
+    }
+
+    /**
      * Returns a new {@link Context}.
      *
      * @return a new {@link Context}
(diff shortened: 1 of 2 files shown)


arteam:
Thanks for your contribution, Steven.

mwhipple:
I may be missing something but this seems like a potentially bad design change.  Now #time accepts 2 possible SAM argument types which will complicate implicit coercion of lambdas or similar first class functions.

samperman:
FYI, this breaks groovy users of this api.  Code like this appears to always call the `Runnable` version and doesn't return anything:

```
def result = timer.time {
  doSomething()
}
```

A workaround is to modify the call to look like this:
```
def result = timer.time ({
  doSomething()
} as Callable)
```

ryantenney:
@mwhipple I don't think that had been considered. We could revert this change as it's likely no one is using it yet. Either way we'll be sure to avoid this problem in v4.

mwhipple:
@ryantenney cool, thanks for the quick response!

ryantenney:
Could I just ask that you open a new issue for this?

arteam:
I had an impression that the compiler is smart enough to figure out which SAM should applied depending on the argument. For instance, this snippet compiles for me:
```java
Timer timer = new MetricRegistry().timer("test");
timer.time(() -> System.out.println());
timer.time(() -> System.currentTimeMillis());
```

mwhipple:
Ok, thanks for the additional info...it looks like this may just be a Groovy issue.
http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2.1
```
If the target type's function type has a void return, then the lambda body is either a statement expression (§14.8) or a void-compatible block (§15.27.2).

If the target type's function type has a (non-void) return type, then the lambda body is either an expression or a value-compatible block (§15.27.2).
```

stevenschlansker:
If it is really that bad, you could just give the method a different name so it is not an overload.

PR #991, opened 2016-07-22
Added support for customizing AdminServlet.

tommy-bo:
Support for removing default elements and adding new elements.

I needed to customize the admin-page. Just thought I'd share my solution. See CustomizedAdminServletTest for an example of usage.

### DIFF
--- a/metrics-servlets/src/main/java/com/codahale/metrics/servlets/AdminServlet.java
+++ b/metrics-servlets/src/main/java/com/codahale/metrics/servlets/AdminServlet.java
@@ -8,6 +8,9 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 public class AdminServlet extends HttpServlet {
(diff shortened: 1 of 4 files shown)


jgangemi:
+1 for this or something like it. i would find this useful as well.

arteam:
Hello @tommy-bo, 

Thanks for the pull request! That's a very interesting approach for customizing the admin servlet, but I'm a little bit reluctant to merge it. I would prefer to keep the servlet simple rather than provide an ability to customize its output via Java code. There is a big amount of HTML template engines out there which are much better in this job. Advanced users can use those should they need something more complex. Let's put this pull requests on hold.

davidmankin:
Hello @arteam,
I understand why you put this PR on hold, but I'm at a loss for how to do something more complex.  The AdminServlet offers no customizability or pluggability for providing any form of templates or content beyond the fixed template.  I can't easily replace it wholesale because it is the code that handles the delegation to the other metrics servlets, and it is instantiated by dropwizard in code that cannot be easily customized.   

Are you aware of any examples or recommendations on how to integrate the various admin servlets with a home page provided by the application, hopefully including adding additional admin servlets too?

If it's as hard as it seems to be, then some refactoring in this servlets package seems wise even if it's not this PR.

PR #992, opened 2016-07-24
Adding more unit tests for the Counter class to cover corner cases

husseincoder:
Sometimes developers pass variables without checking them .. so it could happen that they pass negative values to the inc method and positive values to the dec methods .. and thats why I added more unit tests to cover these 2 corner cases to the CounterTest class.


arteam:
Thanks for your contribution!

PR #993, opened 2016-07-24
Add link to metrics-opentsdb

sps:


sps:
I wasn't really paying attention to the fact these pages are built via Jekyll. I'll close and resubmit a proper PR with the source files updated not the jekyll output...

PR #999, opened 2016-08-16
ScheduledReporter.start() starts with undesired delay

sofax:
`ScheduledReporter.start()` incorrectly uses the interval size for the start delay, too:

```
    public void start(long period, TimeUnit unit) {
        executor.scheduleAtFixedRate(new Runnable() {
            @Override
            public void run() {
                try {
                    report();
                } catch (RuntimeException ex) {
                    LOG.error("RuntimeException thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
                }
            }
        }, period, period, unit);
    }
```

This is a serious problem when the interval is sufficiently big. I suggest to add an alternative version of `ScheduledReporter.start()` to fix the bug without affecting code that relies on this behavior:

```
    public void start(long period, long initialDelay, TimeUnit unit) {
        executor.scheduleAtFixedRate(new Runnable() {
            @Override
            public void run() {
                try {
                    report();
                } catch (RuntimeException ex) {
                    LOG.error("RuntimeException thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
                }
            }
        }, initialDelay, period, unit);
    }
```

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -141,30 +141,6 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
     /**
      * Starts the reporter polling at the given period.
      *
-     * @param initialDelay the time to delay the first execution
-     * @param period       the amount of time between polls
-     * @param unit         the unit for {@code period}
-     */
-    synchronized public void start(long initialDelay, long period, TimeUnit unit) {
-      if (this.scheduledFuture != null) {
-          throw new IllegalArgumentException("Reporter already started");
-      }
-
-      this.scheduledFuture = executor.scheduleAtFixedRate(new Runnable() {
-         @Override
-         public void run() {
-             try {
-                 report();
-             } catch (RuntimeException ex) {
-                 LOG.error("RuntimeException thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
-             }
-         }
-      }, initialDelay, period, unit);
-    }
-
-    /**
-     * Starts the reporter polling at the given period.
-     *
      * @param period       the amount of time between polls
      * @param initialDelay the time to delay the first execution
      * @param unit         the unit for {@code period}
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -3,7 +3,13 @@ package com.codahale.metrics;
 import java.io.Closeable;
 import java.util.Locale;
 import java.util.SortedMap;
-import java.util.concurrent.*;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
@@ -66,7 +72,7 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
      *                 reporter will report
      * @param name     the reporter's name
      * @param filter   the filter for which metrics to report
-     * @param rateUnit a unit of time 
+     * @param rateUnit a unit of time
      * @param durationUnit a unit of time
      */
     protected ScheduledReporter(MetricRegistry registry,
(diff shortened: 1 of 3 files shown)


metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -145,6 +145,26 @@ public void run() {
     }
 
     /**
+     * Starts the reporter polling at the given period.
+     *
+     * @param period       the amount of time between polls
+     * @param initialDelay the time to delay the first execution
+     * @param unit         the unit for {@code period}
+     */
+    public void start(long period, long initialDelay, TimeUnit unit) {
```
arteam: I think it would be better to change the order of parameters, so `initialDelay` comes first, to maintain the same order as in the `scheduleAtFixedRate` method.

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -145,6 +145,26 @@ public void run() {
...
+       executor.scheduleAtFixedRate(new Runnable() {
```
arteam: I think we could avoid duplication and implement the method as:

```java
synchronized public void start(long initialDelay, long period, TimeUnit unit) {
        if (this.scheduledFuture != null) {
            throw new IllegalArgumentException("Reporter already started");
        }
        this.scheduledFuture = executor.scheduleAtFixedRate(new Runnable() {
           @Override
           public void run() {
               try {
                   report();
               } catch (RuntimeException ex) {
                   LOG.error("RuntimeException thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
               }
           }
       }, initialDelay, period, unit);
   }
```

as then make the `start(long period, TimeUnit unit)` method to delegate to it as
```java
start(period, period, unit)
```

metrics-servlets/src/main/java/com/codahale/metrics/servlets/MetricsServlet.java
```diff
@@ -64,6 +64,9 @@ protected String getAllowedOrigin() {
 
         /**
          * Returns the name of the parameter used to specify the jsonp callback, if any.
+         *
```
arteam: It seems to me as an unrelated change. Could you please revert it?

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -145,6 +145,26 @@ public void run() {
     }
 
     /**
+     * Starts the reporter polling at the given period.
```
arteam: Could you add a simple unit test to `ScheduledReporterTest`? A simple test which verifies that the scheduler wakes up with a correct initial delay. Very similar to the `pollsPeriodically` test.

metrics-servlets/src/main/java/com/codahale/metrics/servlets/MetricsServlet.java
sofax: Sorry, already working on it - it got pushed too soon.

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
sofax: Agreed. :)

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
sofax: Again, agreed. 

The lines

```
        if (this.scheduledFuture != null) {
            throw new IllegalArgumentException("Reporter already started");
        }
```

Are new, but they do look reasonable to me.

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
sofax: Will do. All those tests are a little bit unreliable though, as the executor services rely on the system time instead of using a dedicated (mockable) clock.

sofax: Actually, instead of waiting for the reporter to actually do its work, it would be much better to just verify that the method scheduleAtFixedRate() was called with the correct arguments. There is no need to test the executor itself, as that's a collaborator that has already been tested elsewhere.

arteam: > Actually, instead of waiting for the reporter to actually do its work, it would be much better to just verify that the method scheduleAtFixedRate() was called with the correct arguments. There is no need to test the executor itself, as that's a collaborator that has already been tested elsewhere.

Agrred. This would work, too.

sofax: Also, the tests are not isolated. The collaborators (mocks) should be initialized per test method, not per class.

sofax: Gnnhhh ... my IDE is configured to replace ".*" imports with the explicit ones - but I did NOT want to push those changes, as they were not necessary for the fix.

metrics-servlets/src/main/java/com/codahale/metrics/servlets/MetricsServlet.java
sofax: This change has been reverted.

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
sofax: Test cases have been added.

arteam: > Gnnhhh ... my IDE is configured to replace ".*" imports with the explicit ones - but I did NOT want to push those changes, as they were not necessary for the fix.

I think expanding star imports is a good change, we can leave it.

arteam:
Thank for your contribution!

PR #1004, opened 2016-09-08
Ability to provide custom implementations of counters, histograms, etc

Fuud:
Ability to provide custom implementations of counters, histograms, meters and counters.


daniilguit:
This is exactly what I wanted.

arteam:
@Fuud, thank for the pull request! Sorry to get back to you late, but I think this feature now should be covered by #1023 in a little different manner. Could you please have a look?

Fuud:
@arteam It is not same thing. 3rd party libraries still will use default counters and histograms. But this PR will change all implementations to specific one.

PR #1008, opened 2016-09-23
Disable Sent Metric Types Globally

phauer:
Hi everyone,

I like to contribute a new feature to metrics-graphite.

Currently, all "metric types" of a metric are sent to graphite. If you have a Timer you will **always** get max, mean, min, stddev, p50, p75, p95, p98, p99, p999. However, in reality only some of them are really used for monitoring. It would be great, if I could select the metric types that will be send. This way, we can reduce the load on both the network and graphite and save storage.

Cheers, Philipp


arteam:
Hello Philipp,

Sorry to get to back to you late. Your pull request looks reasonable, but could you say why you don't want to use `MetricFilter` to specify metrics which should be reported? I guess your use case could be implement as: 
```java
final Set<String> disabledMetricTypes = new HashSet<String>();
disabledMetricTypes.add(MetricTypes.M15_RATE);
disabledMetricTypes.add(MetricTypes.M5_RATE);

GraphiteReporter.forRegistry(registry) .filter(new MetricFilter() {
                    @Override
                    public boolean matches(String name, Metric metric) {
                        return !disabledMetricTypes.contains(name);
                    }
                })
```

phauer:
HI Artem,
thanks for your feedback. Does this work for you? It doesn't work for me, because `name` doesn't contain the metric type. The values look like this:
```
heap.committed
heap.used
classes
httpsessions.max
systemload.average
mem.free
processors
gc.ps_marksweep.count
threads.peak
mem
nonheap.used
heap
```
I'm using Metrics in conjunction with Spring Boot Actuator. Maybe this is the reason.

arteam:
Got it. My biggest concern for this feature is that we implement it only for `GraphiteReporter`, but not for other reporters. On the other hand, judging that every reporter has an own mechanism for sending metrics, it's rather hard to generalize it. I think we can start with Graphite for the time being.

metrics-graphite/src/main/java/com/codahale/metrics/graphite/MetricTypes.java
```diff
@@ -0,0 +1,23 @@
+package com.codahale.metrics.graphite;
+
+public class MetricTypes {
```
arteam: I think we could make it an enum and move it to `metrics-core`. Upsides:

* Additional type-safety;
* We can use an `EnumSet` for storing a blacklist of metric types;
* Other reporters could access it if needed.

metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
```diff
@@ -40,6 +44,7 @@ public static Builder forRegistry(MetricRegistry registry) {
         private TimeUnit rateUnit;
         private TimeUnit durationUnit;
         private MetricFilter filter;
+        private Set<String> disabledMetricTypes;
```
arteam: We could use `EnumSet` here, if we make `MetricTypes` as an enum `MetricType`

metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
```diff
@@ -48,6 +53,7 @@ private Builder(MetricRegistry registry) {
             this.rateUnit = TimeUnit.SECONDS;
             this.durationUnit = TimeUnit.MILLISECONDS;
             this.filter = MetricFilter.ALL;
+            this.disabledMetricTypes = Collections.emptySet();
```
arteam: If `disabledMetricTypes` is an `EnumSet`, we could use `EnumSet.noneOf`.

phauer:
Good points. I changed the code according to your feedback. Are you okay with my changes?

arteam:
Looks better! Unfortunately, the pull requests still has merging conflicts. I think we could make at least mergeable, so folks who are interested in this feature could build it locally.

arteam:
@ryantenney mentioned that we will likely pass on this feature for the 3.2.* branch, because it creates a disparity between reporters. But we definitely can consider for it Metrics 4.0.0, perhaps in a more generic manner.

arteam:
Superseded by #1048

PR #1012, opened 2016-10-06
Meters for httpclient exceptions

reftel:
Adds counters for exceptions that happen after `InstrumentedHttpRequestExecutor.execute` has been invoked (some errors, like connection refused, invalid URLs, etc. don't make it that far). By default, use the simple name of the exception as the name of the meter, yielding names like     `org.apache.http.client.HttpClient.foobar.SocketTimeoutException`.

### DIFF
--- a/metrics-httpclient/src/test/java/io/dropwizard/metrics/httpclient/InstrumentedHttpClientsTest.java
+++ b/metrics-httpclient/src/test/java/io/dropwizard/metrics/httpclient/InstrumentedHttpClientsTest.java
@@ -1,6 +1,12 @@
 package io.dropwizard.metrics.httpclient;
 
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
+
 import org.apache.http.HttpRequest;
+import org.apache.http.NoHttpResponseException;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpGet;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


ryantenney:
Thanks, I'll take a look!

metrics-httpclient/src/test/java/io/dropwizard/metrics/httpclient/InstrumentedHttpClientsTest.java
```diff
@@ -10,6 +16,8 @@
 import io.dropwizard.metrics.MetricRegistry;
 import io.dropwizard.metrics.MetricRegistryListener;
 import io.dropwizard.metrics.Timer;
+import static junit.framework.TestCase.fail;
```
arteam: I think it's better to use `import static org.junit.Assert.fail` from the consistency point of view, because we already import `org.junit.Before` and `org.junit.Test`

metrics-httpclient/src/test/java/io/dropwizard/metrics/httpclient/InstrumentedHttpClientsTest.java
```diff
@@ -41,4 +49,39 @@ public void registersExpectedMetricsGivenNameStrategy() throws Exception {
...
+            .thenReturn(exceptionMetricName);
+
+        try {
+            client.execute(get);
+            fail();
+        } catch (NoHttpResponseException expected) {
+
+        }
+        serverThread.interrupt();
+        serverThread.join();
```
arteam: I think we could move the cleanup code to the `finally block` in case we *actually* have a failed test case.

metrics-httpclient/src/test/java/io/dropwizard/metrics/httpclient/InstrumentedHttpClientsTest.java
```diff
@@ -41,4 +49,39 @@ public void registersExpectedMetricsGivenNameStrategy() throws Exception {
...
+
+        assertThat(metricRegistry.getMeters()).containsKey(new MetricName("exception"));
```
arteam: I think it's better to move the assertion to the catch block. Doing so we indicate that we expect the client code throws an exception.

reftel:
Thanks for the attention and feedback! I've updated according to your comments.

arteam:
Thanks for the contribution!

PR #1014, opened 2016-10-14
Minor fix in docs, using PickledGraphite instead of Graphite

simad:
`Graphite` and `PickledGraphite` are both subclasses of `GraphiteSender`, so the edited line in docs does not compile.


jplock:
Travis failed for something unrelated to this PR.  I'm going to merge this in anyway since its just a documentation fix.

PR #1017, opened 2016-10-26
Add Hawkular Metrics reporter

jotak:
Note that I only commited on branch 3.2-develoment, please tell me if I should report changes to master/4.0


jplock:
This is fine, thanks!

Issue #968, opened 2016-06-23
GraphiteReporter builder does not allow to set executor


jarek-przygodzki:
When using Metrics GraphiteReporter in Java EE environment with [metrics-cdi](https://github.com/astefanutti/metrics-cdi) it's quite important to use executor service provided by the container

```
    @Inject
    private MetricRegistry metricRegistry;

    @Resource
    ManagedScheduledExecutorService executorService;

    GraphiteReporter reporter = GraphiteReporter.forRegistry(metricRegistry)
                                                      .withExecutor(executorService)
                                                      .build(graphite());
```

But as of now, GraphiteReporter builder does not allow to set executor  and neither does GraphiteReporter constructor.

ryantenney:
Is it not sufficient to tie the reporter instance to the lifecycle by creating a @PreDestroy method to call `close()` on the reporter?

svendiedrichsen:
I don't think it is sufficient to tie the reporter to the lifecycle of the @PreDestroy method because thus the ExecutorService as a resource cannot be managed by the container. You could make use of the administration and monitoring features of the container if you used the injected ExecutorService and allow for runtime parameter changes.
I suggest you take a look at changes #1018 from @vladimir-bukhtoyarov. They are pretty simple and straight forward.

PR #1018, opened 2016-10-29
Make reporters more user friendly for managed environments like GAE or JEE

vladimir-bukhtoyarov:
Added two configuration options for all reporters:
1.  Ability to configure the custom scheduler. This Is very useful option when instantiation of new Thread is problematic.
2. Option which allow to avoid termination the executor which is managed by external resource container(like JEE server).

This request also closes [issue #968](https://github.com/dropwizard/metrics/issues/968)

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
@@ -3,6 +3,7 @@ package com.codahale.metrics;
 import java.io.PrintStream;
 import java.text.DateFormat;
 import java.util.*;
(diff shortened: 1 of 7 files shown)


svendiedrichsen:
I would like to vote for this to go into the next release. Helps saving resources a lot in the specified environments.

vladimir-bukhtoyarov:
@svendiedrichsen What I need to do with current request? Should I rebase it to another branch?

svendiedrichsen:
I have no saying on wether this change gets pulled. I just happen to like it and support it.

arteam:
@vladimir-bukhtoyarov, Thank you very much for your contribution!

PR #1022, opened 2016-11-28
Store total elapsed duration for a Timer

### LINKED ISSUE #712: Store total elapsed duration for a Timer
I would find it useful to have the total elapsed duration for a Timer instance available. Then it would be possible to eg. calculate the average for all timed events. I'm well aware of how inaccurate an average can be, but that's also the case for Histogram snapshots when sampling reservoirs are used.

It could easily be solved by adding a `Counter totalDuration` to Timer. I'd be happy to implement it and submit a pull request if you find it useful as well. Otherwise I guess I have to keep building metrics-core myself.

hashbrowncipher:
I have taken @slovdahl's linked commits from #712 and merged them with master. I think this is a very worthy feature, and #712 has been stalled for over a year. In the intervening period, monitoring systems that encourage applications to expose raw ascending counters (e.g. Prometheus) have gained popularity, and this change would ease integration with such systems. 

Resolves #712.


okutane:
Nice! Too bad no one is gonna merge this. :(

arteam:
I like the pull request and would accept it. I think raw ascending metrics can be quite important for certain monitoring systems and it would be great to expose them for timers. @ryantenney, you had some concerns about this improvement in the origin issue. Would you mind to take a look at this implementation?

ryantenney:
@arteam the problems I see with the feature and implementation are:

 - not all reporters have been updated to include this field, so we'd be introducing inconsistency in the project
 - I haven't seen another request for this feature
 - since the demand is low, I look to whether this can be accomplished with the existing library, and it should be possible to accomplish this with a counter (though admittedly not ideal)
 - an extremely common complaint has been that we already expose too many metrics
 - for v4.0 I'm planning to introduce a feature to allow individual metrics to define what properties they export, so meter or timer could be subclassed and this field (or any new field) could be added and reporters would pick it up

mevdschee:
We also require the sum of the duration to be exposed. Let me explain our use case: We are using DropWizard Metrics exposed by FastJMX, loaded via CollectD, stored into InfluxDB, graphed by Grafana. This setup is working wonders for us as it allows to graph the usage of API calls. 

Time series databases execute the derivative function over monotonous increasing counters. "Count" is such a value and it is useful to us as it allows us to see what calls are often executed. "Sum" is another value that would allow to see how much time was spent executing the specific function, which is (even more) important for performance analysis. We can already look at the gauge values of mean and 95th that are very important to us when deep diving into the performance of a specific API call. 

For the overview of most impacting calls we do need the "Sum" though as InfluxDB cannot multiply the Mean with the Count as they are reported as separate values in a time series database (which has no full SQL capability).

It would be great if you could merge this PR (I can add the sum to other reporters if that is keeping you from merging).

arteam:
I think it's a worthy feature, especially for time-series databases which use monotonous counters (like Prometheus). I will consider this pull requests for inclusion to Metrics 5.0 (which will be after I release Metrics 4.0(Java 8 cleaned up version of Metrics 3.2).

qualidafial:
Just to provide additional context:

At my work we are using Datadog to aggregate metrics from Dropwizard across a fleet of hosts in several environments. Datadog supports metrics with key/value tagging.

The aggregate metrics like p99, mean, etc are only valid within the originating process. Once you start to aggregate metrics across processes, hosts, datacenters, regions, etc, you _must_ have the total time to get an accurate global picture. Otherwise you're e.g. taking averages of averages, and nothing is accurate.

This gets even worse when you add tags into the picture. To publish metrics to Datadog from Dropwizard metrics, we encode our metric names with tags like so: `rest[resource:accounts,method:post,status_code:201]`. The more tags you use, and the more distinct values your tags can have, the more your metrics get broken up into different buckets, and the value of aggregates like p99 in smaller individual buckets is limited.

As an example of how this can bite you: we time database requests, and use tags to distinguish the database engine, schema, operation, and status (success or failure) of the request. Yesterday, the response time graph inexplicably jumped from 2ms average to 300ms, with no explanation. Hours of digging revealed that the server was fine, but that a _single_ timeout to the database created a distinct bucket for that one error on a single server. Thus our average of averages got bumped by a single outlier.

Had the total time metric been available to us, we could have simply divided the total time by the count to get the global mean.

mevdschee:
@qualidafial maybe [mevdschee/tqdev-metrics](https://github.com/mevdschee/tqdev-metrics) works for you? I wrote it with this different strategy/approach towards metrics in mind. Maybe you like it. I would love to hear your feedback!

arteam:
I'm going to release Metrics 4.0 during the weekend and then take a look at merging #1181 to the 5.0 branch.

qualidafial:
It's a Festivus miracle!

qualidafial:
@arteam Any news? I see you closed #1181, is this now supported in the 5.0 branch?

arteam:
It was closed because I renamed the `4.0-development` branch to  `4.0-maitenance`, I will port #1181 to the 5.0.0 branch after #1250 is merged.

arteam:
Submitted #1254

Issue #642, opened 2014-09-03
Suppress runtime exceptions thrown from Gauge


ryantenney:

tkowalcz:
Hi,

this is something we ran into recently. I see that you do not want to handle exceptions from metric classes, but currently this exception will not be reported anywhere and GraphiteReporter will silently stop sending metrics.

Are you going to merge this change or at least log the exception?

dbaggott:
@tkowalcz (and future readers), it's worth noting that the unpleasant "all reporters silently die in the face of any runtime exception" bug present in version 3.0.2 has been addressed in version 3.1.0 and onward.

`ScheduledReporter.start()` [now correctly handles any exception](https://github.com/dropwizard/metrics/blob/v3.1.0/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java#L118) that bubbles up to it: the exception will be logged and subsequent reporting events are not suppressed.

So, as the code currently stands, a `RuntimeException` from a custom `Gauge` (or any other `Metric`) will break reporting for that cycle but reporting will continue to be attempted on the regular schedule.  More importantly, a clear error will be logged which should help people understand what's going on quickly.  Thank you to whoever made that change!

If the Metric runtime exception is intermittent, you'll still get data for reporting cycles where the exception doesn't happen.

All of that being said, @ryantenney, I love the Metrics library but I have to say there's really not a lot of great choices for developers in this scenario and I think there's a pretty reasonable case to be made that the library should support handling `RuntimeException`s from `Metric`s.  Consider this interface:

```java
public interface Gauge<T> extends Metric {
    T getValue();
}
```

For example, let's say it's a `Gauge<Integer>` that reports on the number of "some undesirable event" and your code encounters an exception, what's the correct return value?

You definitely don't want to return 0 and risk falsely representing that there are no such events.  Nor do you want to return some "magic" number like Integer.MAX_VALUE as its misleading and, in that example, blows out the y-axis on any downstream graphs.  A "magic" number like -1 is a little better but it's still problematic since it can appear awfully close to 0 on a graph.  In short, in order to not obscure downstream graphs by blowing out the axis, your "magic" number would have to be tuned to the range of the actual data!  It just not a workable solution.

[...]

PR #1029, opened 2016-12-10
Update Jackson to 2.8.5

jensraaby:
I've been having problems hooking up Dropwizard Metrics with [Finagle Metrics](https://github.com/rlazoti/finagle-metrics) and [Finatra](https://github.com/twitter/finatra) because of conflicting Jackson versions.

I've bumped the version of Jackson to the latest from Maven Central (2.8.5) and the tests all appear to pass successfully.

Are there any other areas that need looking at to ensure the upgrade doesn't break anything?

It may be worth considering this issue as well: https://github.com/dropwizard/metrics/issues/642


arteam:
Looks like everything works with the latest Jackson version. Thanks for your contribution!

Issue #1005, opened 2016-09-13
NaN mean timer value and standard deviation


gkopff:
I am using `3.1.0` and have started seeing `NaN` values for the mean and standard deviation of timers.  (Until recently, I was using `3.0.1` and I never had a NaN value.)

```
"theTimer" : {
      "count" : 4332,
      "max" : 0.288696783,
      "mean" : "NaN",
      "min" : 6.00388E-4,
      "p50" : 0.003298735,
      "p75" : 0.003298735,
      "p95" : 0.003298735,
      "p98" : 0.003298735,
      "p99" : 0.003298735,
      "p999" : 0.003298735,
      "stddev" : "NaN",
      "m15_rate" : 0.08561987306180467,
      "m1_rate" : 0.03960663833117155,
      "m5_rate" : 0.10383331627579845,
      "mean_rate" : 0.02467728882193572,
      "duration_units" : "seconds",
      "rate_units" : "calls/second"
    },
```

I don't think this is expected, is it?

Looking back through the change-log, it appears that `v1.0.2` (Mar 08 2010) eliminated NaN values from timers:

> Timer now uses Welford’s algorithm for calculating running variance, which means no more hilariously wrong standard deviations (e.g., NaN).

Is this a bug?

arteam:
Should be fixed by #1033

Issue #793, opened 2015-05-06
ExponentiallyDecayingReservoir stops registering updates


sotodel:
In `ExponentiallyDecayingReservoir`, if enough time has passed since you created it (~13+ hours), the weight calculation https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java#L97 returns `Double.Infinity`, causing `priority` to be also `Double.Infinity`, causing collisions in the `Map` and therefore no new updates (assuming `newCount > size`).

`update` calls are therefore almost no-op. They only increment the `count`.

We originally noticed this in the .NET port https://github.com/danielcrenna/metrics-net/issues/45.

nickbabcock:
Does [longPeriodsOfInactivityShouldNotCorruptSamplingState test](https://github.com/dropwizard/metrics/blob/1668978c5c2ac1a4a2b853347efba88f48044fc0/metrics-core/src/test/java/com/codahale/metrics/ExponentiallyDecayingReservoirTest.java#L65-99) not cover this concern?

ryantenney:
You're right, it does.

arteam:
Should be fixed by #1033

PR #1033, opened 2016-12-30
Fix race condition in ExponentiallyDecayingReservoir's rescale method

ho3rexqj:
Fixes issues #1005 and #793 - note that in #793 the `longPeriodsOfInactivityShouldNotCorruptSamplingState` test did not cover multiple concurrent requests after a long idle period.

The race condition currently occurs when two threads update the reservoir concurrently - after the first thread has successfully updated `nextScaleTime` (line 162) but before the write lock is obtained (line 163) a second thread updates the reservoir (line 95+).  With several threads pushing updates concurrently this presents a narrow window during which the second thread may obtain the lock before the first resulting in a corrupt reservoir state.  Note, however, that if a snapshot is requested just prior to the two concurrent updates the read lock will already be active when the two threads enter `rescaleIfNeeded`, resulting in a much larger window for this race condition.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java
@@ -159,9 +159,9 @@ public class ExponentiallyDecayingReservoir implements Reservoir {
      * a linear pass over whatever data structure is being used."
      */
     private void rescale(long now, long next) {
-        if (nextScaleTime.compareAndSet(next, now + RESCALE_THRESHOLD)) {
-            lockForRescale();
-            try {
+        lockForRescale();
+        try {
+            if (nextScaleTime.compareAndSet(next, now + RESCALE_THRESHOLD)) {
                 final long oldStartTime = startTime;
                 this.startTime = currentTimeInSeconds();
                 final double scalingFactor = exp(-alpha * (startTime - oldStartTime));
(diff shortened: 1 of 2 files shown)


arteam:
Thank you very much for the fix and very detailed unit test!

ryantenney:
I have an issue with this. The compareAndSet operation is atomic so it shouldn't allow more than one thread into that if block. Granted 'lockForRescale()' does effectively the same thing, however it is a more expensive operation than the compareAndSet.

arteam:
The issue as I see it:

* Thread 1 calls `update`
* Thread 2 calls `update`
* Thread 1 calls `rescaleIfNeeded`
* Thread 2 calls `rescaleIfNeeded`
* Thread 1 reads the CAS and updates it
* Thread 1 gets swapped out
* Thread 2 reads the updated CAS and exits
* Thread 2 obtains the read lock and works with a new weight, but old (not scaled) values
* Thread 1 gets swapped in and acquires a write lock, but it's too late

Basically, the CAS only guarantees that `nextScaleTime` gets updated atomically, but not other reservoir's state.

arteam:
In case the fix the correct, I would like to cherry-pick this change to the 3.1.3-maintenance branch to provide users who affected by this bug an easy path to upgrade.

arteam:
I didn't get more reasons opposing this fix, so I'm going to move this bugfix to the 3.1.3 branch. In the worst case, we have a performance regression, in the best we fix a concurrency bug.

ahadadi:
@arteam, I think your comment:
"Thread 2 obtains a read lock and works with a new weight, but old (not scaled) values"
is incorrect.
The weight becomes *new* when startTime gets updated, but it gets updated only under the write lock by thread 1.
I still think this is a good change, as it simplifies reasoning about the code.

PR #1034, opened 2017-01-04
Add per-measurement level filtering of metrics

michaelsembwever:
This configuration has existed previously, and still exists in metrics-reporter-config.
This commit provides support for it again within the GraphiteReporter.

ref: https://github.com/addthis/metrics-reporter-config/pull/2

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricFilter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricFilter.java
@@ -12,6 +12,10 @@ public interface MetricFilter {
         public boolean matches(String name, Metric metric) {
             return true;
         }
+        @Override
+        public boolean matches(String name, Metric metric, String measurement) {
+            return true;
+        }
     };
 
     /**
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -205,8 +205,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
```
michaelsembwever: The default method in the `MetricFilter` interface requires Java8.

For a fork that provides this functionality this is the simplest solution.

If the pull request would otherwise be accepted I can backport the code to Java6.

arteam: I think if you plan to implement this functionality with Java 8 features, it's better to submit this pull request to the master branch (for the version 4.0.0 which supports Java 8). For the 3.2 branch we still need to maintain Java 6.

michaelsembwever: I'll be able to code it against Java 6 if there's otherwise no objection to the PR conceptually.

michaelsembwever:
> And submit this Java 8 version as a separate PR to master

done in https://github.com/dropwizard/metrics/pull/1041

pom.xml
michaelsembwever: > I'll be able to code it against Java 6 if there's otherwise no objection to the PR conceptually.

Done.

arteam:
After a little bit of thought, I believe I can't merge the pull requests to the 3.2.* branch, because it breaks backward compatibility. Custom `MetricFilter` implementations will become incompatible after the update to 3.2, because they will be forced to implement a new method. @phauer started to work on the same feature in #1008, which, I believe, is a better way to address the problem. The implementation doesn't use `MetricFilter`, but rather a set of disabled metric types. I would like to close this pull requests in favour of #1008. Thank you for your time and effort on this pull request.

PR #1035, opened 2017-01-05
Add support for Pickled Graphite over UDP

michaelsembwever:
Redo of https://github.com/dropwizard/metrics/pull/714

> Pulled the logic to create the Pickled data format into its own class (GraphitePickler).
> Adjusted PickledGraphite to use the GraphitePickler class.
> Created PickledGraphiteUDP class which is to PickledGraphite what GraphiteUDP is to Graphite.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
@@ -162,8 +162,11 @@ public final class PickledGraphiteUDP implements GraphiteSender {
 
     @Override
     public void close() throws IOException {
-        datagramChannel.socket().close();
-        datagramChannel = null;
+        try {
+            datagramChannel.socket().close();
+        } finally {
+            datagramChannel = null;
+        }
     }
 
     @Override
@@ -186,7 +189,6 @@ public final class PickledGraphiteUDP implements GraphiteSender {
                 ByteBuffer byteBuffer = ByteBuffer.allocate(4 + payload.length).putInt(payload.length).put(payload);
 
                 datagramChannel.send(byteBuffer, address);
-                this.failures = 0;
                 if (LOGGER.isDebugEnabled()) {
                     LOGGER.debug("Wrote {} metrics", metrics.size());
                 }
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
@@ -1,0 +1,203 @@
+package com.codahale.metrics.graphite;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.net.SocketFactory;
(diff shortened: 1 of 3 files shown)


metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
```diff
@@ -0,0 +1,203 @@
...
+import java.io.*;
+import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.nio.ByteBuffer;
+import java.nio.channels.DatagramChannel;
+import java.nio.charset.Charset;
+import java.util.LinkedList;
+import java.util.List;
+
+public final class PickledGraphiteUDP implements GraphiteSender {
```
arteam: Could you please add a unit test for `PickledGraphiteUDP`?

metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
```diff
@@ -0,0 +1,203 @@
...
+        if (!isConnected()) {
+            connect();
+        }
+        if (metrics.size() > 0) {
+            try {
+                byte[] payload = GraphitePickler.pickleMetrics(metrics,charset);
+                ByteBuffer byteBuffer = ByteBuffer.allocate(4 + payload.length).putInt(payload.length).put(payload);
+
+                datagramChannel.send(byteBuffer, address);
+                this.failures = 0;
```
arteam: Could you please elaborate why we need to reset the counter? In `PickledGraphite` we don't do that.

metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphiteUDP.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Override
+    public void flush() throws IOException {
+        // No such concept in UDP
+    }
+
+    @Override
+    public void close() throws IOException {
+        datagramChannel.socket().close();
+        datagramChannel = null;
```
arteam: I think it's better to dereference the channel in a finally block in case the socket failed to close.

michaelsembwever:
The code commits are from @philwill-nap, someone i do not know, so I'd like to hold them intact to represent the authorship correctly.

I will add an additional commit addressing the comments and referencing this PR.

arteam:
> so I'd like to hold them intact to represent the authorship correctly.

Sure, you can push your changes as separate commits.

arteam:
So, we have only pme review point to address. If you contribute a unit test for `PickledGraphiteUDP`, I will happily merge the pull request.

michaelsembwever:
>  If you contribute a unit test for PickledGraphiteUDP …

will do, but will need to come back to it soon, being holiday season n all :-)

arteam:
No rush, thanks for taking your time to contribute!

iksaif:
FYI: pickle is usually less performant that lines (uses more CPU) and will produce bigger datagrams that are more likely to be silently dropped (because UDP)

PR #1038, opened 2017-01-05
Fix "Add `getProtocols` method"

arteam:
This method is declared in ConnectionFactory in Jetty 9.3.*.
To be compatible with it, we should have a method with the same
signature.

We can't use a compile-time dependency on Jetty 9.3.*, because
it's compiled against Java 8, but codahale-metrics against Java 6.

Therefore, we use reflection to access this method.

Backport of #823


marsmining:
Thanks for this work! Any idea when this will land in a release?

PR #1041, opened 2017-01-06
Add per-measurement level filtering of metrics

michaelsembwever:
Forward port of https://github.com/dropwizard/metrics/pull/1034

This configuration has existed previously, and still exists in metrics-reporter-config.
This commit provides support for it again within the GraphiteReporter.

ref: https://github.com/addthis/metrics-reporter-config/pull/2

### DIFF
--- a/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporter.java
@@ -139,6 +139,7 @@ public class GraphiteReporter extends ScheduledReporter {
     private final GraphiteSender graphite;
     private final Clock clock;
     private final MetricName prefix;
+    private final MetricFilter filter;
 
     private GraphiteReporter(MetricRegistry registry,
                              GraphiteSender graphite,
@@ -151,6 +152,7 @@ public class GraphiteReporter extends ScheduledReporter {
         this.graphite = graphite;
         this.clock = clock;
         this.prefix = MetricName.build(prefix);
+        this.filter = filter;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


arteam:
This looks better for the 4.0 branch, and implements the same feature as #1008. Unfortunately, the implementation has the same problem (it updates only one reporter, but I would like to avoid disparity between reporters). Let me collect my thoughts on how to implement this feature in a more generic manner.

michaelsembwever:
I can update all the reporters, or at least take a stab at doing do, but let me know first the precedence between this and https://github.com/dropwizard/metrics/pull/1008

metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporter.java
```diff
@@ -223,63 +225,116 @@ public void stop() {
...
-        graphite.send(prefix(name, "mean"), format(snapshot.getMean()), timestamp);
-        graphite.send(prefix(name, "min"), format(snapshot.getMin()), timestamp);
-        graphite.send(prefix(name, "stddev"), format(snapshot.getStdDev()), timestamp);
-        graphite.send(prefix(name, "p50"), format(snapshot.getMedian()), timestamp);
-        graphite.send(prefix(name, "p75"), format(snapshot.get75thPercentile()), timestamp);
-        graphite.send(prefix(name, "p95"), format(snapshot.get95thPercentile()), timestamp);
-        graphite.send(prefix(name, "p98"), format(snapshot.get98thPercentile()), timestamp);
-        graphite.send(prefix(name, "p99"), format(snapshot.get99thPercentile()), timestamp);
-        graphite.send(prefix(name, "p999"), format(snapshot.get999thPercentile()), timestamp);
+        if (filter.matches(name, histogram, "count")) {
```
ryantenney: The issue I have with merging this to v4 is that in v4 all of these lines will have been removed. Instead of each reporter being responsible for selecting which values will be reported for a particular metric, the reporter will either be provided with a collection of measurements or will iterate over the measurements provided by a metric. The effort to update all reports with this logic will be wasted.

arteam:
Superseded by #1048

PR #1048, opened 2017-01-28
Support for disabled metrics

arteam:
This is for based on the work of @phauer in #1008. The set of disabled metrics is implemented in `ScheduledReporter`, so all reporters have an opportunity to implement this feature, but it's not necessarily (for instance `ConsoleReporter` doesn't really need it). Along with the Graphite reporter, the Ganglia reporter is also updated to support of disabling sending metrics.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricAttribute.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricAttribute.java
@@ -1,9 +1,9 @@
 package com.codahale.metrics;
 
 /**
- * Represents types of metrics which can be reported.
+ * Represents attributes of metrics which can be reported.
  */
-public enum MetricType {
+public enum MetricAttribute {
 
     MAX("max"),
     MEAN("mean"),
@@ -23,7 +23,7 @@ public enum MetricType {
 
     private final String code;
 
-    MetricType(String code) {
+    MetricAttribute(String code) {
         this.code = code;
     }
 
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricType.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricType.java
@@ -1,0 +1,33 @@
+package com.codahale.metrics;
+
+/**
+ * Represents types of metrics which can be reported.
+ */
+public enum MetricType {
+
+    MAX("max"),
+    MEAN("mean"),
+    MIN("min"),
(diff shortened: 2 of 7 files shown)


arteam:
I think this should be good to merge. Could anybody give a review?

ryantenney:
Looks good, except that metric types are meter, timer, counter, gauge…

ryantenney:
And is there a reason to go with `disabledMetricTypes` instead of `enabledMetricTypes`?

arteam:
> And is there a reason to go with disabledMetricTypes instead of enabledMetricTypes?

I don't like the inverted logic too. But I think it makes sense from the the user's perspective. The user reports some metrics to its time-series database and sees that some of them are useless, because he/she doesn't use them. The user wants to disable them. For instance, `Timer` has 15 metrics, and most of them make sense; it would be strange to ask the user to pass  a 14-sized set of metrics just to disable one.

ryanrupp:
Is this only for the 3.2 branch or would it go into master/4.0 too? See the conversation in https://github.com/dropwizard/metrics/issues/837 and https://github.com/dropwizard/metrics/issues/178 - the idea that each "Metric" has a set of "Attributes" (name/value) off it. I'm not sure on the status of those but maybe instead of a MetricType enum, allow filtering by String based "attribute names" and just provide a helper class with the list of standard/known attribute names that currently exist (what is present in the MetricType enum currently). For 3.2 the enum makes sense as there's a fixed number of attributes, for 4.0 if https://github.com/dropwizard/metrics/issues/837 becomes a thing then this would just want to be changed to a String (suppose there'd be so many other breaking changes that it could just be worried about at that point). Anyway, thought of it when I saw the comment about "metric types".

ryantenney:
@ryanrupp Good point. This would only go to 3.2, for exactly the reasons you mention. Also there will be some inversion of control with regard to reporters in v4 to make this sort of filtering easier to implement.

@arteam Ryan reminded me that we're calling these 'Attributes', can we rename MetricType and disabledMetricTypes to MetricAttribute and disabledMetricAttributes?

arteam:
That's a good suggestion. Renamed `MetricType` to `MetricAttribute`.

arteam:
I fixed the build (it was affected by relevant changes from the 3.1 branch moved to 3.2), now it should be good to merge.

PR #984, opened 2016-07-12
Wrap try-catch each report method.

voyageth:
Wrap try-catch each report method for prevent lose metrics when some metric goes wrong.

Following error is occured on Cassandra 3.3 + metrics-graphite.
After that error, i lose all metered, timer metric + some histogram metric.

>  ERROR [metrics-graphite-reporter-1-thread-1] 2016-07-12 16:22:41,625 ScheduledReporter.java:119 - RuntimeException thrown from GraphiteReporter#report. Exception was suppressed.
> java.lang.IllegalStateException: Unable to compute ceiling for max when histogram overflowed
>         at org.apache.cassandra.utils.EstimatedHistogram.rawMean(EstimatedHistogram.java:231) ~[apache-cassandra-3.3.jar:3.3]
>         at org.apache.cassandra.metrics.EstimatedHistogramReservoir$HistogramSnapshot.getMean(EstimatedHistogramReservoir.java:103) ~[apache-cassandra-3.3.jar:3.3]
>         at com.codahale.metrics.graphite.GraphiteReporter.reportHistogram(GraphiteReporter.java:252) ~[metrics-graphite-3.1.0.jar:3.1.0]
>         at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:166) ~[metrics-graphite-3.1.0.jar:3.1.0]
>         at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:162) ~[metrics-core-3.1.0.jar:3.1.0]
>         at com.codahale.metrics.ScheduledReporter$1.run(ScheduledReporter.java:117) ~[metrics-core-3.1.0.jar:3.1.0]
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_45]
>         at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [na:1.8.0_45]
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [na:1.8.0_45]
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [na:1.8.0_45]
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_45]
[...]

PR #1049, opened 2017-01-30
Suppress all kinds of Exceptions raised by report()

iksaif:
With the current code exceptions that do not inherit from RuntimeException will
make this task die silently.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -155,8 +155,8 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
          public void run() {
              try {
                  report();
-             } catch (RuntimeException ex) {
-                 LOG.error("RuntimeException thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
+             } catch (Exception ex) {
+                 LOG.error("Exception thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
              }
          }
       }, initialDelay, period, unit);


arteam:
Looks good. We should catch all exceptions, despite it's a rare case when a reporter implementation throws a unchecked exception.

arteam:
Thank you for your contribution!

iksaif:
Thanks for merging. Any idea when 3.2.0 will be released ?

arteam:
I'm currently merging/reviewing outstanding pull requests and bug fixes from the backlog to the 3.2.x branch. A big part of backlog was processed, but I would like to land #1047 and #1048 to 3.2.* and after that make a release. I will discuss it with @ryantenney (he has the deployment rights). Hopefully, we can make a release this week, but I can't promise it.

iksaif:
Great thanks, I'll ask for a version bump in Cassandra once it's released

arteam:
If the Cassandra project is depended on it, I think it makes sense to backport the fix to the 3.1.* branch as well, so it's easier to upgrade for the Cassandra folks.

iksaif:
Looks like #1047 and #1048 got merged, I guess the release is getting closer ?

arteam:
Looks like so. @ryantenney told me yesterday he is ready to cut the releases.

iksaif:
Looks like this wasn't enough :/ We should replace Exception by Throwable

arteam:
Does the Cassandra reporter throw `Error`s? I thought it's a bad practice to catch them, because you could end up swallowing unrecoverable errors.

vladimir-bukhtoyarov:
Yes, Cassandra throws exception on snapshot extraction time, and it does it by design(not because of bug). Currently I am trying to promote [this request](https://github.com/vladimir-bukhtoyarov/cassandra/commit/4dca54c1000576a892a77bc716f87adc4bc05ecc) which solves this issue [CASSANDRA-13223](https://issues.apache.org/jira/browse/CASSANDRA-13223). I hope I will achieve success after fixes for mentioned notes.

I want to raise following thought: it is better to surround by **try/catch** the code places [like this](https://github.com/criteo-forks/metrics/blob/16587fb330f84c2094ddc8fe9611e0bd1c48cbb2/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java#L216), because if any histogram throws exception then we have no reports at all. Missing of reports can lead to epic fail when something going wrong and you are unable to know how much wrong. The think that MetricsServlet should act in similar way and surrounds by try/catch each particular metric. And that why I do not use any built-in Dropwizard reporter, because I do not want to come in situation when reports are missed because of tricky bugs(like this #1104) or absurd design solution(like Cassandra).

arteam:
There is a PR for GraphiteReporter which provides something like that (#984), I could take a look at it. 

I think it's not a bad idea and could improve reliability and resilience. Doing so, we avoid situations when a failure for a set of metrics prevents the whole MetricsRegistry from reporting (or worse sends bogus data). If we adopt it, it means we will not imply that all metrics implementation don't throw exceptions, but embrace failures and find a means to mitigate from them. That might be not so easy, but it looks like a step in the right direction.

iksaif:
@arteam the thing is that currently the silentely ignore Throwables and Errors subclasses. IMOH it's better to log something.

arteam:
I'm still a lit bit reluctant to change `Exception` to `Throwable`. This is an easy fix for the Cassandra issue, but potentially can have a negative impact on other reporters. I still believe that handling `Error`s is not a right thing to do. Could point me on the place where Cassandra throws an `Error`?

iksaif:
That's the thing: I don't know since the reporter just silently gets un-scheduled forever. I'll try a patched version catching everything to see if I can find out.

arteam:
Ah, I see. Thanks for hunting the issue!

iksaif:
Ok I can confirm that fixes the issue because there are a lot of cases where an AssertionError (Error, not Exception) is being throwned, for example:

{code}
/var/log/cassandra/system.log.18.zip:ERROR [metrics-graphite-reporter-1-thread-1] 2017-05-02 15:17:47,178 ScheduledReporter.java:176 - Exception thrown from GraphiteReporter#report. Exception was suppressed.
/var/log/cassandra/system.log.18.zip:java.lang.AssertionError: null
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.io.util.Memory.size(Memory.java:376) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.utils.obs.OffHeapBitSet.serializedSize(OffHeapBitSet.java:131) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.utils.BloomFilterSerializer.serializedSize(BloomFilterSerializer.java:65) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.utils.BloomFilter.serializedSize(BloomFilter.java:64) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.io.sstable.format.SSTableReader.getBloomFilterSerializedSize(SSTableReader.java:1295) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.metrics.TableMetrics$29.getValue(TableMetrics.java:601) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.metrics.TableMetrics$29.getValue(TableMetrics.java:596) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.metrics.TableMetrics$35.getValue(TableMetrics.java:742) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
/var/log/cassandra/system.log.18.zip:	at org.apache.cassandra.metrics.TableMetrics$35.getValue(TableMetrics.java:736) ~[apache-cassandra-3.10-criteo1.jar:3.10-criteo1]
{code}

I've sent https://github.com/dropwizard/metrics/pull/1128

PR #1051, opened 2017-01-31
Sending double / float metrics to graphite without truncating

asv:
Previous implementation truncates floating point number to 2 decimal places. This limits the applicability of metrics
with a floating point in some situations. Bacause graphite-carbon uses float() method (see [1]) to parse value in
plaintext protocol, we can safely do Double.toString to convert Java Double to a propper string for plaintext protocol
value.

[1] https://github.com/graphite-project/carbon/blob/a52cfe026fc283546e40e4aecded8ac3c194def9/lib/carbon/protocols.py#L119

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -323,7 +323,7 @@ public class GraphiteReporter extends ScheduledReporter {
 
     private String format(Object o) {
         if (o instanceof Float) {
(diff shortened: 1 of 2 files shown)


arteam:
Thank you for the pull request! Unfortunately, it can be accepted because it changes the format of data sent to Graphite. As a user, I would be surprised if after an update to a new version of the metrics library, my charts would start showing numbers like 0.234223345898 instead 0.23 without my intention to change the format.

Having said that, I can the value in sending floating point numbers in a more precise format. I think, a better approach would be to mark the `format` method as protected, so users who want to have a custom format, can override it.

PR #1054, opened 2017-02-01
metrics-httpclient test fix

rconn01:
Fixed test to catch `IOException` instead of just `NoHttpResponseException` since the test can throw a `SocketException: Connection reset` which would cause the test to fail


arteam:
Thank you for the contribution.

PR #1052, opened 2017-02-01
Support the JDK's ThreadLocalRandom

arteam:
Add a factory which provides a ThreadLocalRandom implementation depending on its availability in runtime. If the JDK provides ThreadLocalRandom, use it. Otherwise, fallback to the internal implementation.

Issue #375, opened 2013-04-07
Remove Striped64 and LongAdder when switching to Java 8


codahale:
No real timeline on this, either.

(c.f. ffeed35b063d9be1f3d69ff7d1272612f95b60db)

ikedahi:
I think you can replace UNSAFE with AtomicLongFieldUpdater for Cell's value, AtomcLong for Striped64's base, and AtomicBoolean (or AtomicInteger) for Striped64's busy.

matlach:
Striped64 also hold a reference to ThreadLocal in its actual form.
If you have a look to the latest JDK8 implementation source code this has been removed.

I'm pretty sure this can lead to the same issue discussed in #742 .

ikedahi:
I was surprised but I agree.

In order to avoid holding a specific class via Striped64, it might be fixed by making ThreadHashCode extend ThreadLocal&lt;int[]&gt; (with int[1]) instead of ThreadLocal&lt;HashCode&gt;, which is just used to hold a mutable hash for each thread.

PR #1055, opened 2017-02-02
Support JDK's LongAdder

arteam:
Add an ability to use the JDK's implementation if it's available in runtime. Add a factory for creating LongAdders which tries to use the JDK's implementation and fallbacks to the internal one if the JDK doesn't provide any. Unfortunately, LongAdders don't have a common interface, therefore we introduce `LongAdderAdapter`. It will be used in other components as an interface, a concrete implementation of it will be created by `LongAdderFactory`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ install: echo "I trust Maven."
 script: mvn verify
 
 jdk:
-  - oraclejdk7
   - oraclejdk8
 
 notifications:
--- a/metrics-core/src/main/java/com/codahale/metrics/Counter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Counter.java
@@ -4,10 +4,10 @@ package com.codahale.metrics;
  * An incrementing and decrementing counter metric.
  */
 public class Counter implements Metric, Counting {
-    private final LongAdder count;
+    private final LongAdderAdapter count;
 
     public Counter() {
-        this.count = new LongAdder();
+        this.count = LongAdderProxy.create();
     }
 
     /**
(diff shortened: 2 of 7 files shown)


arteam:
@ryantenney did something similar for the 4.0 branch in #876

arteam:
This also should fix `ThreadLocal` leaks in the internal LongAdder implementation (by using the JDK implementation on a modern JRE) as reported in #375.

arteam:
Bummer. The build can't compiled on Travis on JDK 7, because LongAdder doesn't exist in the JDK.

@dropwizard/dropwizard-metrics I'm wondering if we can remove it (JDK 6/7 will be still supported in runtime).

ryantenney:
You've got the right idea, but what I'd recommend is adding a metrics-java8 module that is compiled using Java 8 (and is ignored if the compiler is <8). Then if someone wants to use the native longadder, they can import that module.

I went looking for an old branch where I started to do just that, but I can't find it.

arteam:
Thanks for the feedback! I got your point, but I think it would be better to implement this feature in a maintenance release without additional modules. The internal `LongAdder` is not only has worse performance, but it's prone to memory leaks (because of its use of thread locals). I think it would be great to provide a release which will just fix the issue without requiring the developer/administrator to enable an additional module (so the metrics jar could be just replaced in an application server).

arteam:
I've updated the pull request to align the implementation with the `ThreadLocal` one from #1052

jplock:
Looks good to me, want me to merge this in @ryantenney ?

ryantenney:
@jplock go for it!

PR #1066, opened 2017-02-07
Build with error-prone

dpursehouse:
Update maven-compiler-plugin to 3.6.1.

Add dependency on error-prone [1] and configure the compiler to use it.

Remove test statement that causes "[SelfEquals] An object is tested
for equality to itself" error [2].

[1] http://errorprone.info
[2] http://errorprone.info/bugpattern/SelfEquals

Change-Id: Id96a211e3013ca8913d3d7fd2eb1d7233890cd72


arteam:
Looks good to me. Thanks for the contribution!

PR #693, opened 2014-10-20
Adding a HealthCheckRegistryListener

v-garki:
A HealthCheckRegistryListener similarly to MetricRegistryListener which could be used to expose health checks elsewhere besides the admin page (like JMX).

PR #1068, opened 2017-02-13
Adding a HealthCheckRegistryListener

v-garki:
A HealthCheckRegistryListener similarly to MetricRegistryListener which could be used to expose health checks elsewhere besides the admin page (like JMX).
Rebased version of #693


ryantenney:
Thanks!

PR #1071, opened 2017-02-15
Merge changes from the 3.1 branch to 3.2

arteam:
Merge changes from the 3.-maintenance branch to the 3.2 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ install: echo "I trust Maven."
 script: mvn verify
 
 jdk:
-  - oraclejdk7
   - oraclejdk8
 
 notifications:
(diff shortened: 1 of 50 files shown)


arteam:
We actually interested only in the last commits plus some docs update,

jgangemi:
may i ask why the `cpu profiling` support change got reverted? i was looking forward to using that.

arteam:
It is available in the 3.2 branch. It was originally submitted by a mistake to the 3.1 branch and then reverted. The revert commit shows up on the merge, but if you take a look at the diff, no CPU profiling changes are reverted.

jgangemi:
thanks for the clarification!!

PR #1073, opened 2017-02-15
Add support for overriding format of floating numbers in Graphite

arteam:
Currently, we trim all decimal values to 2 digits after the point. Despite that's a good default, some applications require reporting floating numbers with better precision.

For those users we provide the ability to override the `format(double v)` method of `GraphiteReporter`. User can provide a custom format (or just use a plain double-to-string conversion). We also need to make the constructor of `GraphiteReporter` protected, because users can't override a class with a private constructor.


ryantenney:
I like this approach! Makes it possible for the few people that want this feature without adding any cruft.

PR #1074, opened 2017-02-17
Fix for the monitoring jetty9 async servlet requests.

arteam:
Async requests are monitored via the addition of an async listener.
The bug was that `onStartAsync(AsyncEvent)` is called when the handler
invokes `startAsync()` but the `InstrumentedHandler` only added the
listener after the call to the delegate handler, thus would miss
the all onStartAsync event.

The net result of the bug was that the start time of all async requests
was reported as 0 so the calculated elapsed times where then rather
large (i.e. currentTimeMillis()).

### DIFF
--- a/metrics-jetty9/src/main/java/com/codahale/metrics/jetty9/InstrumentedHandler.java
+++ b/metrics-jetty9/src/main/java/com/codahale/metrics/jetty9/InstrumentedHandler.java
@@ -227,6 +227,7 @@ public class InstrumentedHandler extends HandlerWrapper {
             // new request
             activeRequests.inc();
             start = request.getTimeStamp();
+            state.addListener(listener);
         } else {
             // resumed request
             start = System.currentTimeMillis();
@@ -246,9 +247,6 @@ public class InstrumentedHandler extends HandlerWrapper {
             dispatches.update(dispatched, TimeUnit.MILLISECONDS);
 
             if (state.isSuspended()) {
-                if (state.isInitial()) {
-                    state.addListener(listener);
-                }
                 activeSuspended.inc();
             } else if (state.isInitial()) {
                 updateResponses(httpRequest, httpResponse, start);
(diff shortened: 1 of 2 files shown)


arteam:
This is the work of @tempredirect in #715 rebased to the 3.2 branch.

ryantenney:
LGTM

tempredirect:
Great work, thank you for doing this.

tnn:
Hi @arteam, we just observed this on `3.2.3`, `3.2.6` and `4.0.2`:
```
jetty_active_requests -4.0
jetty_active_suspended -5.0
```
We have verified this on multiple systems. The service were this is observed, uses the following setup:
jetty: 9.4.8.v20171121
jersey: 2.25.1
Java: 9.0.0 and 9.0.4.

We are using Jersey/JAX-RS AsyncResponse. We have inspected the heap and thread - and found no evidence that the AsyncResponse has not properly been resumed. We have verified both running as `-jar` and `jmod`.

It appears that a request is immediately completed (buffered by Jersey), but still a AsyncResponse, the `if (state.isSuspended())` branch in `InstrumentedHandler` is never hit, because while the request has state (`HttpChanelState`):
```
_state = {org.eclipse.jetty.server.HttpChannelState$State@7071} "DISPATCHED"
_async = {org.eclipse.jetty.server.HttpChannelState$Async@7072} "COMPLETE"
_asyncRead = {org.eclipse.jetty.server.HttpChannelState$AsyncRead@7073} "IDLE"
```

The state while incrementing (line 224, `if (state.isInitial()) {`):
```
_state = {org.eclipse.jetty.server.HttpChannelState$State@7071} "DISPATCHED"
_async = {org.eclipse.jetty.server.HttpChannelState$Async@7229} "NOT_ASYNC"
_asyncRead = {org.eclipse.jetty.server.HttpChannelState$AsyncRead@7073} "IDLE"
```

So it appears that the `activeSuspended.dec();` on line 232 needs to take into account if the request is async, is completed and that the `async` state has gone from `NOT_ASYNC` to `COMPLETE`.

Please let us know, if we can help in anyway.

Cross posted as an issue: https://github.com/dropwizard/metrics/issues/1324.

PR #1076, opened 2017-02-18
ObjectNameFactory throws MalformedObjectNameException

### LINKED ISSUE #1065: change io.dropwizard.metrics.ObjectNameFactory to throw exception
While breaking changes can still be made on the 4.x branch, would it make sense to change the signature of `io.dropwizard.metrics.ObjectNameFactory#createName` to throw a `MalformedObjectNameException`. This way implementations can throw that exception if they can't generate an object name. This would ultimately bubble up to the listener methods, such as `onCounterAdded` where a `JMException` is caught and logged. 

As the code stands now there is no way for an `ObjectNameFactory` to give up. The best option would be to throw a RuntimeException, which could break application flow.

I'm willing to propose a change but wanted to get some feedback first

rconn01:
change signature on ObjectNameFactory to throw a MalformedObjectNameException

closes #1065

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/DefaultObjectNameFactory.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/DefaultObjectNameFactory.java
@@ -3,15 +3,10 @@ package io.dropwizard.metrics;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 public class DefaultObjectNameFactory implements ObjectNameFactory {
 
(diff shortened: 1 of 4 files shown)


ryantenney:
I don't entirely understand the point of this. In either case we're catching an logging the exception, right?

rconn01:
Yes it is true that it is being logged in both cases. In the existing case however a runtime exception is thrown, which will not be handled anywhere and bubble up all the way to the calling code preventing the metric from being added to the metric registry. Adding `throws MalformedObjectNameException` also prevents other implementations from needing to do the catch and logging logic since it's already being done in the JMXReporter

ryantenney:
Thanks for explaining, I missed the RuntimeException!

rconn01:
@ryantenney @jplock what else needs to happen to get this merged?

arteam:
Thank you for your contribution!

PR #1077, opened 2017-02-18
(#609) Introducing async health checks

arkigabor:
A backward compatible solution for asynchronous health checks (#609). HealthCheck implementations can be marked with the Async annotation while providing:
- execution period (mandatory)
- scheduling type (fixed rate or fixed delay)
- initial delay before first execution
- unit of period and delay
- initial health state until the first result is available

Async checks can be registered the same way regular health checks. On registration the health check registry is wrapping the async health check with a decorator that caches the last result and schedules a periodic execution. This decorator object is put into the underlying map and when its check is invoked it simply returns a cached result. On unregistration the registry cancels the task. To preserve the asynchronous behavior elsewhere the listeners are notified with the async decorator instead of the original health check.

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/AsyncHealthCheckDecorator.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/AsyncHealthCheckDecorator.java
@@ -1,0 +1,60 @@
+package com.codahale.metrics.health;
+
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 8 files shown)


arteam:
Generally looks good to me. Thanks for your contribution! This could be a very useful feature and using annotations makes it very easy for users to convert a health check to an async health check. My biggest concern is to how correctly shut down a `HealthCheckRegistry` with async executors. Let me play a little bit with it, maybe I will add a follow-up a PR.

PR #1078, opened 2017-02-19
Introduce a MetricNameFormatter 

rconn01:
This PR is an initial proposal on adding a `MetricNameFormatter`. The idea is users can provide an implementation that will format a `MetricName` into a string for reporters that only support single strings. This PR also makes the reporters aware of tags in the `MetricName`. Most of them currently just print out `MetricName.getKey()` which is not guaranteed to provide uniqueness for things such as the slf4j and console reporters. Let me know your thoughts


rconn01:
@ryantenney I have added a `MetricNameFormatter` that calls `MetricName.toString()`

rconn01:
@ryantenney any additional comments are thoughts here?

arteam:
Looks good to me. Thanks for the contribution!

PR #1080, opened 2017-02-19
Move closing` DatagramChannel` to the close method

arteam:
We can align `GraphiteUDP` with the TCP based `Graphite` and close the
socket in the `close` method instead of the connect method. The close
method will be invoked every time after a portion of metrics has been
sent, so we can close the socket there. We also do not need to call
the method `isConnected` during sending metrics because the socket is
opened right before sending data, so it can't be closed by timeout.


ryantenney:
I think I understand, since we now always close the socket after reporting a batch of metrics, it doesn't make sense to wait for the next connect call to close the DatagramSocket?

arteam:
Right. The `close` method is always invoked, so we can close the socket there instead leaving it open and closing on the next connection.

PR #1084, opened 2017-02-24
Shutdown health check registry

arteam:
A follow-up for #1077 

* Add a method for graceful shutdown of HealthCheckRegistry
Add an ability to gracefully shutdown the executor for async tasks in `HealthCheckRegistry`.
Now it creates a ScheduledExecutorService which should be correctly closed when `HealthCheckRegistry` is not used anymore.

*  Make the pool size more pessimistic by default
Some CI environments return the amount of available processors as 128.
It's rather expensive to create such big thread pools by default. Let's start
with 2 threads and the user wants to increase the amount,  she can do that.

* Make the async health check threads as daemons
So they don't prevent the JVM from shutdown if the user didn't shut down
the executor gracefully.


jplock:
LGTM

arteam:
Bummer, I screwed up and this change was not added to 3.2.0. I will mark it to 3.2.1.

PR #1086, opened 2017-02-24
3.2.0 release

arteam:


jplock:
@arteam should we try updating dropwizard 1.1.x to use metrics 3.1.3 and then update dropwizard `master` to use metrics 3.2.0?

arteam:
Sure, maybe it even makes sense to update 1.1.x to 3.2.0 straight ahead.

jplock:
@arteam agreed, it's been a long time coming and i think getting 3.2.0 into DW 1.1.0 would be great.

PR #1088, opened 2017-02-24
Link Coursera's fork for metrics-datadog

xiaochuanyu:
Coursera's fork is much more up to date than the original.


arteam:
Thanks! I already updated the link to the Datadog reporter to the Coursera's fork during the overhaul of reporters in #1089. Now http://metrics.dropwizard.io/3.2.0/manual/third-party.html points out to it.

PR #1090, opened 2017-02-25
When the developer wants to ignore the counter attribute in the defin…

husseincoder:
When the developer wants to ignore the counter attribute in the defined meters metrics .. that doesn't mean he/she is still not interested in the defined counter metrics to be logged .. especially that counter metrics contains only the count attribute - this PR will force sending the counter.count even if its configured to be ignored for meters


arteam:
I think this makes sense. I fail to see a use case when the user wants to disable reporting of a `Counter`. It disables the whole metric from being reported and this can be done by `MetricFilter`. On the other hand, I can see a situation when the users wants to disable the counter attribute from a `Meter`, but allow `Counter`s to be reported.

@husseincoder Could you please update the `GangliaReporter` for consistency as well?

husseincoder:
@arteam I've changed it too with a small unit test edit .. unless you want me to extract it in a separate unit test?

arteam:
Thanks for your contribution!

Issue #595, opened 2014-06-04
Support for tapping into submitted measurements


JensRantil:
(Metrics=this project, metrics=measurement sensors)

**Background:** I have a dynamic set of machines that are running Metrics. Each machine has a dynamic (< 30) set of similar services that I store metrics for (counters and timers). I am using Graphite reporter with hostname prefix to view stats in Graphite. I think above is a fairly common setup.

**Problem:** Global statistics. The problem is that it's hard get an overall perspective of the all my hosts. Under the assumption that I'd like to use Metrics for my data collection there are two solutions, but both are flawed:
- Don't prefix using hostname, instead add another Graphite reporter with prefix "global" on all your Metrics instances. Great idea, until you realize that Graphite only stores the _last write_ on its smallest time resolution (can be a second as smallest). You will loose data from an instance as soon as it submits the same second as someone else. Not only that, Metrics will aggregate values such as percentiles (Graphite agnostic). This is hard to summarize on a global level.
- Use [statsd](https://github.com/etsy/statsd/) (or similar). Good idea, but...statsd required a more pure data than what Metrics delivers. statsd calculates percentiles itself, and does not expect precalculated percentiles.

**Proposed solution:** Allow custom taps on all submitted metrics to Metrics. This would allow for a statsd emitter that would let statsd (or other collection/backend solution) do all of it's work. Also, it would be a generic solution to easily collect data globally somewhere. Possibly using sampling for bigger clusters etc.

What do you think? Has this been discussed before?

Issue #822, opened 2015-06-25
Metric event streams


ryantenney:
Provide an interface through which a listener can be registered to receive MetricEvent objects that match a MetricName query. MetricEvent would include the time and type of event (Timer, Meter, Counter, Histogram) and the associated value (timer=time, meter and counter=count, histogram=value).

felixbarny:
Sounds cool. What is it good for?

pvalsecc:
Example of use case:
Integration with statsd.
The current implementation doesn't send timer events right away but, collects them for a given period then sends 10 messages to statsd (max, mean, min, stddev, ...). This is not very well integrated with my other statsd metric coming from other softwares.

Having such feature would allow a better statsd integration.

marshallpierce:
This would be pretty expensive for frequently updated metrics, though (it allocates, for instance). While a useful feature, I'm not sure that it should be part of the typical Meter, etc, types since those are all about local aggregation and occasional publishing of batches of data. Perhaps this is a separate use case that deserves a separate set of types.

pvalsecc:
Proposal here:
https://github.com/pvalsecc/metrics/tree/streaming

jmason:
if a metric-recording event suddenly starts happening more frequently than usual, this risks exceeding the bandwidth available for the listener (or the upstream service ie. statsd).  It may be a good idea to have some way to drop limit-exceeding updates from the stream, and/or support approximation or sampling.

pvalsecc:
@marshallpierce: saying something is expensive without actually trying can be surprisingly wrong in computer science. For example, your statement "it allocates" is invalid in the context of performance. Since java 1.7 (available but disabled in java 1.6), the JIT is doing escape analysis and uses the stack or even the registers for storing temporary objects. Therefore, nowadays, allocation doesn't necessarily mean bad perfs.

@jmason: this is more a problem of the "plugin" using the event system to send them to another service. For example statsd uses UDP. If you put the socket in asynchronous, the OS or the network is going to drop some stats for you if it becomes too busy. But since UDP doesn't have any congestion control, unlike TCP, you may put your network to its knees. But still, IMHO, it's not the business of the event system within Metrics.

marshallpierce:
@pvalsecc EA is quite limited in its applicability in the JVM. I would be cautious (especially in a library like this, not a full service that can more tightly control its runtime environment) of relying on it, even if it turns out to be applicable in some setups.

pvalsecc:
[...]

PR #1091, opened 2017-02-25
Provides extension point to access to individual measurements

mrmanc:
As suggested in #595.

I’ve done this work on the 3.2-development branch as it seemed most up to date. I’d be happy to move it to 4.0-development as it is a breaking change to the public API of `Metric`.

* `MetricRegistry.addMeasurementListener()` now provides an extension point which lets users add listeners which will receive individual measurements for all registered metrics.
* The registry holds a single `MeasurementPublisher` so that each metric object does not need to hold a separate collection.
* Adds the metric name and a publisher to the signature of all metric constructors to allow them to emit measurements with the associated name. I couldn’t see a nicer simple way of doing this while retaining the metrics as the source of measurements.
* Nested metrics do not report their measurements since these are meta-metrics—aggregations themselves.
* There is now a slightly wrinkly problem around `MetricRegistry.register()` since the name of the metric is supplied *and* present in the metric. I could add a `name()` accessor to `Metric` but this would be a further break to the interface.
* `Gauge` measurements do not make it to the listeners. I can contribute an instance of `ScheduledReporter` which optionally passes (configured through the `MetricRegistry`) `Gauge` samples on a regular interval to listeners, and will contribute this approach if the maintainers are happy.

### DIFF
--- a/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/CounterBenchmark.java
+++ b/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/CounterBenchmark.java
@@ -9,10 +9,12 @@ import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 
+import static com.codahale.metrics.MeasurementPublisher.DO_NOT_PUBLISH;
+
 @State(Scope.Benchmark)
 public class CounterBenchmark {
 
-    private final Counter counter = new Counter();
+    private final Counter counter = new Counter("benchmark-counter", DO_NOT_PUBLISH);
 
     // It's intentionally not declared as final to avoid constant folding
     private long nextValue = 0xFBFBABBA;
--- a/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/MeterBenchmark.java
+++ b/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/MeterBenchmark.java
@@ -9,10 +9,12 @@ import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 
+import static com.codahale.metrics.MeasurementPublisher.DO_NOT_PUBLISH;
+
 @State(Scope.Benchmark)
 public class MeterBenchmark {
 
-    private final Meter meter = new Meter();
+    private final Meter meter = new Meter("meter", DO_NOT_PUBLISH);
 
     // It's intentionally not declared as final to avoid constant folding
     private long nextValue = 0xFBFBABBA;
--- a/metrics-core/src/main/java/com/codahale/metrics/Counter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Counter.java
@@ -5,8 +5,12 @@ package com.codahale.metrics;
  */
 public class Counter implements Metric, Counting {
     private final LongAdderAdapter count;
+    private final String name;
+    private final MeasurementPublisher measurementPublisher;
 
-    public Counter() {
+    public Counter(String name, MeasurementPublisher measurementPublisher) {
+        this.name = name;
+        this.measurementPublisher = measurementPublisher;
         this.count = LongAdderProxy.create();
     }
 
@@ -24,6 +28,7 @@ public class Counter implements Metric, Counting {
      */
     public void inc(long n) {
         count.add(n);
+        measurementPublisher.counterIncremented(name, n);
     }
 
     /**
@@ -40,6 +45,7 @@ public class Counter implements Metric, Counting {
      */
     public void dec(long n) {
         count.add(-n);
+        measurementPublisher.counterDecremented(name, n);
     }
 
     /**
--- a/metrics-core/src/main/java/com/codahale/metrics/Histogram.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Histogram.java
@@ -7,15 +7,21 @@ package com.codahale.metrics;
  *      variance</a>
  */
 public class Histogram implements Metric, Sampling, Counting {
+    private final String name;
+    private final MeasurementPublisher measurementPublisher;
     private final Reservoir reservoir;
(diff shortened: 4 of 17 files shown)


vladimir-bukhtoyarov:
I see following problems with this PR:
- All third-party implementations of histograms, timers and counters will be broken by this change.
- All client code which instantiates any metric directly through constructor will be broken.
- The metric name must not be a private field of metric itself, because one metric can be added to multiple registers, and same metric can has different names in different registers.

Also the problem of listeners can be easy solved by using decorator pattern, without adding unnecessary complicity to the core of library.

mrmanc:
Thanks for the comments!

I have gone back and forth on the design of this change, trading off complexity to the design and breaking existing code and felt that actually submitting something might lead to some good suggestions.

> All third-party implementations of histograms, timers and counters will be broken by this change.

I’m not sure that third party implementations histograms, timers and counters would be broken unless you’re referring to the change to `MetricSupplier`?

> All client code which instantiates any metric directly through constructor will be broken.

What do you think about retaining the previous constructors in a deprecated form? That way no client code would break, but measurements would be lost for code which wasn’t updated.

> The metric name must not be a private field of metric itself, because one metric can be added to multiple registers, and same metric can has different names in different registers.

That’s interesting—I wondered if there was an explicit design decision behind the name not existing on the metrics themselves and hadn’t considered that. Our more complicated design for this change gets around having to add the name to the metrics by holding it on an intermediate object which acts as a listener on the metric. It’s a layer of complication I wanted to avoid but could reintroduce to resolve that problem.

> Also the problem of listeners can be easy solved by using decorator pattern, without adding unnecessary complicity to the core of library.

Are you suggesting decorating the registry rather than changing any classes in the library itself? Following that approach, I could not see a way to obtain measurements emitted by third party client code. I believe a change to each implementation of `Metric` would still be required, although I believe that can be done without breaking any public API.

clohfink:
Why not just make a custom reporter that pushes the metric to the listeners? Then there wont be any broken apis and no changes needed in metrics core.

ryantenney:
Something resembling this is planned for v4. Let me take a look at it
tomorrow and I'll see if it's a good fit.

vladimir-bukhtoyarov:
@mrmanc,

> I’m not sure that third party implementations histograms, timers and counters would be broken unless you’re referring to the change to MetricSupplier?

There are many open-source projects like Cassandra or Rabbitmq Java Driver uses Dropwizzard/Metrics for monitoring and makes own implementation of timers or histograms. In general Metrics Core is well extensible, but absence of interfaces(only Gauge has interface) for metrics is worst drawback for extensibility. As result, there is only one way to provide own implementation would be subclassing already existed class from core. But subclass must call constructor of superclass and if already used constructor changes its signature then subclass will be broken. So each client which uses this kind of third-party libraries will decline migration to new version of metrics-core until thirdparty library will be compatible with new versions of metrics-core, and this creates the risks of community splitting.

> What do you think about retaining the previous constructors in a deprecated form? That way no client code would break, but measurements would be lost for code which wasn’t updated.

This solution help to preserve backward compatibility, but introduces the point of inconsistency. Your request is heavy concentrated on flow when MetricRegistry creates the new Metric instance, but there is another flow when Metric firstly instantiated in client code through constructor and then [added to registry](https://github.com/dropwizard/metrics/blob/3.2-development/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L83), and that is why specifying listeners at construction time is undesirable, because code which creates new metric potentially is very far from MetricsRegistry and knows nothing about listeners which can be applied on metric.

mrmanc:
@vladimir-bukhtoyarov 

> There are many open-source projects like Cassandra or Rabbitmq Java Driver uses Dropwizzard/Metrics for monitoring and makes own implementation of timers or histograms.
…
> As result, there is only one way to provide own implementation would be subclassing already existed class from core. But subclass must call constructor of superclass and if already used constructor changes its signature then subclass will be broken.

Yep, I see that now, thanks for clarifying. Extracting interfaces from each of these implementations would also break downstream implementations but would be easier to resolve. Could you point me at those implementations you mentioned as I was unable to find them? The implementations I did find appeared to exist to [alter the aggregation mechanism](https://github.com/HuangZhenQiu/Progression-Framework/blob/c7f0a5b794951a7825d6a56e93a39dda7cae93ef/server/src/main/java/edu/uci/eecs/wukong/framework/api/metrics/Timer.java) or to [hook into the measurements](https://github.com/circonus-labs/metrics-circonus/blob/master/metrics-circonus/src/main/java/com/circonus/metrics/TimerAlaCoda.java). The latter case should be satisfied by providing a mechanism to hook into measurements.

I think the first case highlights a problem with the current design. The identity of the metrics and mechanism to report a measurement is too closely coupled to the metrics aggregation. That reminds me of a [comment](https://github.com/dropwizard/metrics/issues/822#issuecomment-228406852) made by @bentatham in #822 about making the current aggregations observers themselves.

Metrics has become somewhat a de-facto standard around recording metrics, and I think that has been positive, encouraging more third party instrumentation of code with correct semantics around the type of events. I think I would love to see more separation between the interface (to register, retrieve and measure against metrics) and the rest of the functionality. This would allow people to make things ‘metrics-compatible’ without making any assumptions about what happens downstream.

> So each client which uses this kind of third-party libraries will decline migration to new version of metrics-core until thirdparty library will be compatible with new versions of metrics-core, and this creates the risks of community splitting.

Since 4.0 is going to be a breaking change as it is then third party library updates seem like they are going to be necessary anyway.

> This solution help to preserve backward compatibility, but introduces the point of inconsistency. Your request is heavy concentrated on flow when MetricRegistry creates the new Metric instance, but there is another flow when Metric firstly instantiated in client code through constructor and then added to registry, and that is why specifying listeners at construction time is undesirable, because code which creates new metric potentially is very far from MetricsRegistry and knows nothing about listeners which can be applied on metric.

Agreed. How would you feel about making the implementations mutable by adding an `addMeasurementListener()` method instead? Mutability is something I try to avoid.

vladimir-bukhtoyarov:
> How would you feel about making the implementations mutable by adding an addMeasurementListener() method instead?

@mrmanc I think it would be better to not change existing Metric implementations and MetricRegistry. You can easy achieve your goals by extending registry and metrics, else there is no clean and effective way to keep API and documentation in consistency.

PR #1093, opened 2017-02-26
Link to rolling-metrics library

vladimir-bukhtoyarov:
- Added new section for custom metrics implementations.
- Added link to rolling-metrics.

@marshallpierce, also I think that it would be better to move hdrhistogram-metrics-reservoir  to the new "Advansed metrics implementations" section because it is really provide new implementation of histogram and does not provide ability to monitor any library or server.


arteam:
Thank you!

PR #1094, opened 2017-02-27
Make SharedMetricRegistries thread-safe

arteam:
Change `defaultRegistryName` from volatile to `AtomicReference`. This allows to us to perform an atomic set of the default registry bane and not end  up with a lost update.

This also adds a package-private setter for `defaultRegistryName` for testing purposes, so we don't perform any reflection hacks in the unit tests.


arteam:
Based on the work of @mattnelson in #849

PR #1095, opened 2017-02-27
Add support for the default shared health check registry name

arteam:
Add the ability to use a shared health check registry across different modules. The managed health check registry is set via the `setDefault` method and then it could be accessible by consumers via the `getDefault` method.

This is aligned with `SharedMetricRegistries`, which provides the same functionality.


arteam:
The original idea from @mattnelson in #849

sata:
@jplock Does the tests work for you?

I get 

> Expecting empty but was:<["db"]>
>         at com.codahale.metrics.health.SharedMetricRegistriesTest.clearsRegistries(SharedMetricRegistriesTest.java:52)

Cheers.

jplock:
The tests passed on travis.

Issue #694, opened 2014-10-22
GraphiteReporter should reconnect when connection fails temporarily


jsyrjala:
With Metrics 3.1.0 GraphiteReporter we had a problem where data was no longer sent to graphite server. This happened after some network component was rebooted. Same server is sending data also with Metrics  2.1.3  GraphiteReporter. Older graphite reporters kept on sending data just fine.

Sounds like this could be related to #607.

Stacktrace from 3.1.0:

```
2014-10-22 12:00:21,036 [WARN] [] [] com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:179) Unable to report to Graphite: java.net.SocketException: Broken pipe
        at java.net.SocketOutputStream.socketWrite0(Native Method)
        at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:113)
        at java.net.SocketOutputStream.write(SocketOutputStream.java:159)
        at sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:221)
        at sun.nio.cs.StreamEncoder.implWrite(StreamEncoder.java:282)
        at sun.nio.cs.StreamEncoder.write(StreamEncoder.java:125)
        at java.io.OutputStreamWriter.write(OutputStreamWriter.java:207)
        at java.io.BufferedWriter.flushBuffer(BufferedWriter.java:129)
        at java.io.BufferedWriter.write(BufferedWriter.java:230)
        at java.io.Writer.write(Writer.java:157)
        at com.codahale.metrics.graphite.Graphite.send(Graphite.java:130)
        at com.codahale.metrics.graphite.GraphiteReporter.reportGauge(GraphiteReporter.java:270)
        at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:158)
        at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:162)
        at com.codahale.metrics.ScheduledReporter$1.run(ScheduledReporter.java:117)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:304)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:178)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
[...]

Issue #664, opened 2014-09-22
MetricFilter's matches(final String name, final Metric metric) returns true if Includes are specified and Excludes are not/


qinfchen:
If Includes are not empty and excludes are empty, this should be the includes-only case that 
No metrics are reported, except those with a name listed in includes.

But !getExcludes().contains(name) would always return true which ends up reporting all the metrics;

PR #1096, opened 2017-02-28
By sending a "garbage" byte we're keeping the socket open to hostedgr…

sata:
…aphite.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/PickledGraphite.java
@@ -40,7 +40,6 @@ public class PickledGraphite implements GraphiteSender {
     private final Charset charset;
 
     private Socket socket;
-    private Writer writer;
     private int failures;
 
     /**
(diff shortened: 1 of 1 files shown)


arteam:
It should not be a problem now, because we don't maintain a persistent TCP connection to Graphite anymore in 3.1.1 and 3.2.0. But the idea is good, I believe the Apache HTTP Client does something similar to detect dead connections in the pool.

sata:
@arteam PickledGraphite is maintains a persistent socket in 3.2 development branch?

arteam:
No. PickledGraphite is just an implementation of GraphiteSender, which is supplied to a GraphiteReporter. GraphiteReporter invokes the connect method on the sender, send metrics, flushes them, and the invokes the close method. So a new TCP connection opens every time.

sata:
@arteam Is that the intended design? Because that's not how Druid graphite extension is doing which is using this library[1] from what I can understand. Nor does it seem to be with given these two tickets #664 #694. Am I missing something here?
 
[1] https://github.com/druid-io/druid/blob/master/extensions-contrib/graphite-emitter/src/main/java/io/druid/emitter/graphite/GraphiteEmitter.java

arteam:
No, it's not the original design. The mechanism was changed in the 3.1.3 and 3.2.0 versions (they were released last week), because many users reported issues with persistent connections; a connection could become broken and there were no reliable approach, how to detect it. The new approach leads to bigger latencies during sending metrics (3-way handshake is rather expensive), but on the other hand guards against halted reporters.

arteam:
I think the Druid database could try to use this approach as well or try to maintain a pool of renewable PickledGraphite instances.

sata:
Thanks! I'll keep it in mind when discussing it with Druid guys. I wasn't planning on creating a PR to dropwizard but rather our own forked version of it but thanks for commenting!

arteam:
Glad to help.

PR #1097, opened 2017-03-01
GroupId change

### LINKED ISSUE #1069: Change groupId on 4.x
Are there any plans to change the maven `groupId` on the current master (4.x) branch. Since it shares the same groupId `io.dropwizard.metrics` with the 3.1.x branch there is no way for the two versions of the code to live in the same code base. All of the java code looks like it was repackaged from `com.codahale.*` to `io.dropwizard.*` which is good so the classes of the two versions can live side by side. But there is no way to pull in the two artifacts with the same group

rconn01:
change the groupId so there is no collision with 3.x artifacts.

If there is another groupId that is preferred let me know.  

fixes #1069

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -21,7 +21,7 @@ You need the ``metrics-core`` library as a dependency:
 
     <dependencies>
         <dependency>
-            <groupId>io.dropwizard.metrics</groupId>
+           <groupId>io.dropwizard.metrics4</groupId>
             <artifactId>metrics-core</artifactId>
             <version>${metrics.version}</version>
         </dependency>
@@ -121,7 +121,7 @@ So the complete Getting Started is
 
     <dependencies>
       <dependency>
-        <groupId>io.dropwizard.metrics</groupId>
+       <groupId>io.dropwizard.metrics4</groupId>
         <artifactId>metrics-core</artifactId>
         <version>${metrics.version}</version>
       </dependency>
@@ -376,7 +376,7 @@ To use this servlet, include the ``metrics-servlets`` module as a dependency:
 .. code-block:: xml
 
     <dependency>
-        <groupId>io.dropwizard.metrics</groupId>
+       <groupId>io.dropwizard.metrics4</groupId>
         <artifactId>metrics-servlets</artifactId>
         <version>${metrics.version}</version>
     </dependency>
--- a/metrics-annotation/pom.xml
+++ b/metrics-annotation/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>io.dropwizard.metrics</groupId>
+       <groupId>io.dropwizard.metrics4</groupId>
         <artifactId>metrics-parent</artifactId>
         <version>4.0.0-SNAPSHOT</version>
     </parent>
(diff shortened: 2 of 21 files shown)


ryantenney:
After thinking about it some more, I don't think this is necessary. Anyone that wants to use, say, `metrics-httpclient` @ 3.1.3 would specify that version, then also specify `metrics-core` and `metrics-adapter` @ 4.0.0, then `metrics-httpclient` would interact with the backwards compatible API exposed by `metrics-adapter`. Am I wrong?

ryantenney:
Didn't mean to close it, sorry!

rconn01:
I see what you are saying. I think that mostly makes sense. I think there is still a potential for collision on larger projects that depend on other libraries that might be out of their control. This is a made up situation so might not be unrealistic but lets say...

- Project `A` is an opensource project that uses `metrics-httpclient` of 3.x to instrument their http calls (say a lib to interact with a rest api)

- Project `B` is a newer opensource lib that is instrumenting its calls using `metrics-httpclient` of 4.x
- Project `C` makes use of both of these libraries, but only one of the versions will exist in the classpath. 
- Project `C` will get a classNotFound for the `metrics-httpclient` classes in one of the branches

- Project `C` has no easy way to make both libraries work without forking and changing the code of one of the libs

ryantenney:
Ah yea, that would be a problem. Probably wouldn't affect more than a couple people, if that, but the groupId change could be preferable to causing this problem for anyone. Another possibility could be to make one final release of Metrics 3.x using the groupId `io.dropwizard.metrics3`, for the couple of people that encounter the scenario you've described, they can add a dependency on `io.dropwizard.metrics3:metrics-httpclient` to project `C`.

rconn01:
Agree that would work as well, users would just need to know that is available and that they need to make those changes. I think the groupId would be a little more seamless to end users, but I also agree the number in the groupId seems strange. I'll lean on you guys with how you want to proceed

rconn01:
@ryantenney looking at the adapter module I see a few issues that could cause problems for users.
-  `com.codahale.metrics.MetricRegistry` is missing the default constructor that it previously had.
-  `addListener` method calls were not implemented on `com.codahale.metrics.MetricRegistry`
- `com.codahale.metrics.SharedMetricRegistries` is missing
- `InstrumentedExecutorService` is missing
- `InstrumentedScheduledExecutorService` is missing
- `InstrumentedThreadFactory` is missing 
- `JmxAttributeGauge` is missing
- `JvmAttributeGaugeSet` is missing
- All the `Metric` implementations (meters,counters,etc) are missing matching constructors to what they previously had. 

There may be a few others that I have not run across yet.

What would be the idea to support all these use case that could break for users. Would we make all the required changes in the adapter package to account for full backwards compatibility?

Personally I do kind of like the idea of having the adapter package, this way there is still only one registry in your application and you don't have to manage a 3.x and a 4.x registry. Which would then require having multiple reporters, one for each registry version. With the adapter the 3.x registry is simply adapted to the 4.x version and it all "becomes" one registry. I'm just curious if we think will truly be able to provide full backwards compatibility everywhere for users to keep a painless upgrade path

arteam:
Thank you. I think the idea to switch to the new namespace is right. Metrics4 will be significantly different from Metrics 3.

PR #1098, opened 2017-03-07
To be honest I disagree with the way we do sanitizing here and would …

husseincoder:
To be honest I disagree with the way we do sanitizing here and would even remove it as graphite supports many special characters and it depends on the developer to choose better naming conventions to his/her metrics .. when I upgraded the library to use the new version all my dashboards didn't work as it uses '.' in metric names which in this case has been converted to '-' developers should be able to track metrics in the following formats (e.g., search.total, search.error, search.done) and as soon as these characters are fine with graphite then why would we need to strip them or normalize them?! .. I just added some of the special characters that I'm sure are accepted by graphite and I think we should support whatever graphite supports to be backward compatible .. especially when users upgrade to the new version

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteSanitize.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteSanitize.java
@@ -1,26 +1,32 @@
 package com.codahale.metrics.graphite;
 
 class GraphiteSanitize {
+
+    private static final String ACCEPTED_CHARS = ".-_";
+
     /** Replaces all characters from a given string that are not ascii and not alphanumeric
      *  with a dash */
-    static String sanitize(String string, char replacement) {
-        String replaced = replaceFrom(string, replacement);
+    static String sanitize(String string, char replacementCh) {
+        String replaced = replaceFrom(string, replacementCh);
+
+        final String replacementStr = String.valueOf(replacementCh);
+        final String duplicateChars = (replacementStr+replacementStr);
 
         // Consolidate multiple dashes into a single one
-        String result = replaced.replace("--", "-");
+        String result = replaced.replace(duplicateChars, replacementStr);
         while (!result.equals(replaced)) {
             replaced = result;
-            result = replaced.replace("--", "-");
+            result = replaced.replace(duplicateChars, replacementStr);
         }
 
         // Remove any leading or trailing dashes
-        return strip(result, replacement);
+        return strip(result, replacementCh);
     }
 
     /** A char matches when it is a letter or digit and it is ASCII, in Guava terminology,
      *  this would be CharMatcher.ASCII.and(CharMatcher.JAVA_LETTER_OR_DIGIT).negate() */
     private static boolean matches(char c) {
-        return !(Character.isLetterOrDigit(c) && c <= '\u007f');
+        return !((Character.isLetterOrDigit(c) || ACCEPTED_CHARS.contains(String.valueOf(c))) && c <= '\u007f');
     }
 
     /** Replace all characters that we're interested in with a replacement character,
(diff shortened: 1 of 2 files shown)


husseincoder:
If we are going to leave the sanitization happening in metrics .. so I think it should be fare to allow developers pass the ACC characters in the GraphiteReporter .. Users don't have to change their WORKING dashboards when they upgrade to a newer version .. all the changes should cope with previous versions and allow ways to control the logic being done.

ryanrupp:
Replacing dots with a dash does seem like a bug here as Graphite uses dot delimited naming sequence for path segments (and changing these to slashes would break wildcard queries and such that are across the path segment, see render API note about wildcards/path segments [here](http://graphite.readthedocs.io/en/latest/render_api.html).

I'm not sure which Graphite sender you're using but it looks like the sanitize method is protected so you can override it in the meantime if you're configuring the reporter directly.

Side note but I also noticed the TCP based graphite sender wasn't actually updated to use this new sanitizer, see [here](https://github.com/dropwizard/metrics/blob/3.2-development/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java#L189). Going back to the change I think it came up because of other characters like apostrophes mentioned here - https://github.com/dropwizard/metrics/issues/637 - but yeah I don't think periods should be replaced.

ryanrupp:
Also the value gets sanitized the same way so I'm guessing that creates a problem for decimals [here](https://github.com/dropwizard/metrics/blob/3.2-development/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteUDP.java#L71) for instance :)

arteam:
I agree with @ryanrupp and will put this pull request on hold, unless we have an evidence the sanitizing is broken.

> Side note but I also noticed the TCP based graphite sender wasn't actually updated to use this new sanitizer

Thanks! I will submit a PR to address that.

husseincoder:
@arteam I'm not sure if we really need the sanitize functionality .. that means we need users of our library to override the sanitize method to remove special characters they have declared? may be I'm missing something .. would you please clarify the main purpose of having it .. also graphite works fine with at least the characters I put in this PR .. please let me know :)

ryanrupp:
@arteam to clarify my comment I think the existing behavior released in 3.2 has a bug so this PR may be valid (or at least some sort of changes so dots in metric names do *not* get replaced with slashes). Specifically, this test case [here](https://github.com/dropwizard/metrics/blob/3.2-development/metrics-graphite/src/test/java/com/codahale/metrics/graphite/GraphiteSanitizeTest.java#L22) is invalid. I haven't tested this locally but the values are getting sanitized too so:

12.34 --> 12-34

which would be bad. Basically dots/periods should be left alone as they're expected to be used in a metric name (in both metric conventions and graphite conventions). The only time it would make sense to sanitize a period would be if it was within one of the "components" of a metric name e.g.:

```
// foo.bar.baz
// Debatable if it's really foo.bar-baz in Graphite land
MetricRegistry.name("foo", "bar.baz");
```

However with the 3.x API in Metrics that cannot be detected (since a String is just being produced).

arteam:
Thanks for clearing this up. I will try to test 3.2.0 with on a real Graphite installation.

nickbabcock:
(Original feature implementer here): Oof, that does seem like a pretty bad bug!

My intent was for the sanitize feature to be used on individual parts of a named metric. Akin to:

```java
MetricRegistry.name("A", "B", sanitize("St. Foo's of Bar"))
```

This is especially important with IPs/hosts:

```java
MetricRegistry.name("A", "B", sanitize("127.0.0.1"))
```

One would expect: "A.B.127-0-0-1"

arteam:
I think the best remedy is too fallback to removing whitespaces how we do in the Graphite TCP sender.
This should remove the bug and make all senders consistent.

husseincoder:
Closing this one then ;)

arteam:
Metrics 3.2.1 has been released with the less aggressive sanitizer.

PR #1099, opened 2017-03-09
Rollback GraphiteSanitize to replacing whitespaces

arteam:
As reported in #1098 the current sanitizer is too aggressive. It strips dots which play an important role in the Graphite metrics name convention. Until we develop a better algorithm, let's do only basic sanitization.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
@@ -10,13 +10,11 @@ import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.UnknownHostException;
 import java.nio.charset.Charset;
-import java.util.regex.Pattern;
 
 /**
  * A client to a Carbon server via TCP.
  */
 public class Graphite implements GraphiteSender {
-    private static final Pattern WHITESPACE = Pattern.compile("[\\s]+");
     // this may be optimistic about Carbon/Graphite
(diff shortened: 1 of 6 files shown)


jplock:
LGTM

nickbabcock:
I feel like we should keep the ability to create metric names that don't crash graphite or else issues like #637 becomes relevant again. IMO, rollback the behavior but expose `GraphiteSanitize#sanitize` so that users can create metric names that won't crash graphite.

husseincoder:
Awesome .. thanks alot for acting so fast on this :)

arteam:
> I feel like we should keep the ability to create metric names that don't crash graphite or else issues like #637 becomes relevant again. IMO, rollback the behavior but expose GraphiteSanitize#sanitize so that users can create metric names that won't crash graphite.

Users can override the `sanitize` method on graphite senders, so it should be possible to provide a custom sanitization.

nickbabcock:
But the issue is that the `Graphite` senders do not see the individual named parts of a metric, only the whole name, so the problem remains. Ideally one would sanitize the individual parts like I showed in #1098. I thought it would be convenient to have the code in the base metric library because anyone who is using graphite and may have metrics remotely based on user input should be sanitizing them.

arteam:
That's true. I just wanted to rollback to the original behaviour so non-TCP senders are not broken.

nickbabcock:
Yeah and that was the right decision. I'll kick around a PR for users to opt into an individual name graphite sanitizer 😄

PR #1107, opened 2017-03-17
Fix groupId in pom.xml files

dpursehouse:
It looks like the groupId was changed to io.dropwizard.metrics4
but was missed in a few places.


arteam:
Thank you!

jplock:
we could probably simplify a lot of these and set them to `${project.groupId}` to avoid hardcoding them everywhere (same with DW too).

arteam:
Good idea. I've been thinking to start actively develop the 4.0 branch (to move all bugfixes and features from 3.2, remove outdated code, actively target Java8, remove outdated modules).

dpursehouse:
@jplock @arteam done in PR #1109

PR #1109, opened 2017-03-18
Use ${project.groupId} in child pom.xml dependency declarations

dpursehouse:
It was already done like that for some; do it for all, to make it
consistent.


jplock:
Thanks!

PR #1110, opened 2017-03-18
Merge 3.2 to master

arteam:
[WIP] This pull requests ports changes from the 3.2 branch to master. The idea is too continue development in master with full support of Java 8.


arteam:
I'm inclined to merge it and then move everything which will be added to 3.2 by cherry-picking (like #1111)

arteam:
I also don't like that we lose git history, but there is so many merge conflicts because the was code ported back & forth, so it looks like it's really feasible.

jplock:
@arteam after merging this into `master`, will we be able to merge `3.2-development` into `master` or will we still have merge conflicts?  It'd be nice to follow git-flow or another flow that doesn't involve cherry picking to avoid forgetting to merge commits.  Is this ready to merge?

arteam:
Interesting question. I think there will be conflicts, unless we don't create a special merge commit. Let me try.

arteam:
Basically, the biggest problem is the namespace change and the `MetricName` abstraction. On a related note, I've been thinking recently to start the master branch from 3.2, with the goal keep the namespace and the public API compatible. I fear the Python 2/3 case, if the versions will be drastically incompatible. People just won't upgrade to the new library. I think it would be better to evolve the library in a backwards compatible manner. Just a thought after spending a lot of time merging codebases.

jplock:
@arteam is this good to merge?

arteam:
I would like to sit on this issue a little bit. I start experimenting in the https://github.com/dropwizard/metrics/tree/metrics4 branch and would like to see how it goes.

arteam:
I have decided to proceed with the `metrics4` branch as a new master. The current master, unfortunately, is too much diverged from `3.2-development`, which is the main metrics version. It makes more sense to start metrics 4 from 3.2-development without big changes like changing package names and major breaking API changes.

rconn01:
>It makes more sense to start metrics 4 from 3.2-development without big changes like changing package names and major breaking API changes.

Out of curiousity what are the goals for metrics 4.x then? It seems the big driver would be to allow for breaking changes to be introduced. If there are no breaking changes being added what would be the motivation to move to a new 4.x version?

Is there still a goal to have the `MetricName` introduced in this new version. I feel like that was the biggest benefit of the 4.x branch. With many data stores supporting tagging it seems that this metric library should support that ability as well. I would really hate to see that feature disappear.

>I fear the Python 2/3 case

I agree that is definitely a concern but I'm not sure it's exactly the same here. If the two versions live under different coordinates and packages (like they currently do) both versions of the library can co-exist in an application, so users don't have to pick one or the other like they do with python2/3. Ideally we would add enough features to the newer 4.x version that would make users want to switch. The nice part being that they don't have to switch everything only newer development that they are working on. 

This (changing package) seems to be a path that many libraries take for upgrading. For example the apache [http-client](https://hc.apache.org/)

PR #1111, opened 2017-03-19
Fix creating a uniform snapshot from a collection

### LINKED ISSUE #1104: ArrayIndexOutOfBoundsException in UniformSnapshot constructor
When constructing a UniformSnapshot, the passed in collection of values seems to get modified asynchronously. After the construction of the new value array and before the iteration of the collection, elements seem to get added, resulting in the array being to small.

In my opinion there needs to be some synchronization in SlidingTimeWindowReservoir when getting the collection of values, because there is no other way to get a clean copy of the collection. 

An alternative would be to use a list instead of an array in UniformSnapshot for the values, because due to dynamic sizing this issue won't occur.

Stacktrace:
```
java.lang.ArrayIndexOutOfBoundsException: 26591
        at com.codahale.metrics.UniformSnapshot.<init>(UniformSnapshot.java:29) ~[metrics-core-3.2.0.jar:3.2.0]
        at com.codahale.metrics.SlidingTimeWindowReservoir.getSnapshot(SlidingTimeWindowReservoir.java:67) ~[metrics-core-3.2.0.jar:3.2.0]
        at com.codahale.metrics.Histogram.getSnapshot(Histogram.java:54) ~[metrics-core-3.2.0.jar:3.2.0]
        at com.codahale.metrics.Timer.getSnapshot(Timer.java:159) ~[metrics-core-3.2.0.jar:3.2.0]
        at com.codahale.metrics.graphite.GraphiteReporter.reportTimer(GraphiteReporter.java:287) ~[metrics-graphite-3.2.0.jar:3.2.0]
        at com.codahale.metrics.graphite.GraphiteReporter.report(GraphiteReporter.java:259) ~[metrics-graphite-3.2.0.jar:3.2.0]
        at com.codahale.metrics.ScheduledReporter.report(ScheduledReporter.java:251) ~[metrics-core-3.2.0.jar:3.2.0]
        at com.codahale.metrics.ScheduledReporter$1.run(ScheduledReporter.java:174) [metrics-core-3.2.0.jar:3.2.0]
```

arteam:
The current code for creating a snapshot from the collection is unfortunately broken. The code fails if the provided collection is modified during creating of the snapshot and it has a weak iterator (precisely what `ConcurrentSkipListMap` provides). The iterator doesn't guarantee that it has exactly the same amount iteration as the size of the collection. As a result, the code throws an `ArrayIndexOutOfBoundException`.
The correct behaviour is too dump the collection to a resizable array on the fly and then convert it the array of long primitives.

The behaviour was correct in the 3.1 branch, but was broken in 3.2 by the optimization in #972.

Fixes #1104

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/UniformSnapshot.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/UniformSnapshot.java
@@ -23,14 +23,14 @@ public class UniformSnapshot extends Snapshot {
      * @param values    an unordered set of values in the reservoir
      */
     public UniformSnapshot(Collection<Long> values) {
-        this.values = new long[values.size()];
-        int i = 0;
-        for (Long value : values) {
-            this.values[i++] = value;
+        final Object[] copy = values.toArray();
+        this.values = new long[copy.length];
+        for (int i = 0; i < copy.length; i++) {
+            this.values[i] = (Long) copy[i];
         }
         Arrays.sort(this.values);
     }
-    
+
     /**
      * Create a new {@link Snapshot} with the given values.
      *
--- a/metrics-core/src/test/java/com/codahale/metrics/UniformSnapshotTest.java
+++ b/metrics-core/src/test/java/com/codahale/metrics/UniformSnapshotTest.java
@@ -3,8 +3,10 @@ package com.codahale.metrics;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Random;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


metrics-core/src/main/java/com/codahale/metrics/UniformSnapshot.java
```diff
@@ -23,14 +23,14 @@
      * @param values    an unordered set of values in the reservoir
      */
     public UniformSnapshot(Collection<Long> values) {
-        this.values = new long[values.size()];
-        int i = 0;
-        for (Long value : values) {
-            this.values[i++] = value;
+        final Object[] copy = values.toArray();
```
jplock: can this be `Long[]` to avoid the cast below?  Just curious

arteam:
It's possible to use the `toArray` version which is type-safe. I don't think there is much point, though. The array is then transformed to a primitive array and basically it's just a data dump. It would be great to avoid unnecessary copying, but I don't know a simple way to avoid it except changing the format of snapshot's storage from the array to the list.

ryantenney:
We might as well just revert #972, since toArray still results in copying the backing array.

PR #1115, opened 2017-03-26
Improve ScheduledReporter convertDurations precision

kyeah:
Hello! This was something I noticed while testing metric conversions recently.

For a precision test like this:

```java
@Test
public void shouldConvertDurationToMillisecondsPrecisely() {
  assertEquals(2.0E-5, reporter.convertDuration(20), 0.0);
}
```

The current implementation will fail:

```
java.lang.AssertionError: 
Expected :2.0E-5
Actual   :1.9999999999999998E-5
```

However, if we perform the division directly on given durations, the test will pass.

I don't believe this change degrades the precision for other inputs, but happy to hear otherwise and backpedal :)


arteam:
Thank you. I think it should be the same from the math's point of view, but more precise, because we do only one division, instead of division and multiplying.

PR #1116, opened 2017-03-27
Remove redundant specification of type arguments

dpursehouse:
Since the introduction of generic type inference [1] in Java 7, it
is redundant to specify the type both in the declaration and the
definition.

[1] http://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html


ryantenney:
Thanks!

dpursehouse:
That was quick.  Thanks!

PR #1117, opened 2017-03-27
Add missing @Override annotations

dpursehouse:
Methods that override the implementation from a superclass or
interface should be annotated with @Override as described in:

  http://stackoverflow.com/a/94411/381622


dpursehouse:
Ping @arteam -- any chance to merge this?

arteam:

dpursehouse:
@arteam thanks for merging.  Can you confirm which method you used to merge the PR?  Was it the rebase method?

The reason I ask is because I noticed that the commit author got changed to the personal email address that I use on github, rather than my corporate one that I actually used to do the commit.  I'm wondering if Github did that automatically.

dpursehouse:
@arteam sorry, never mind, I've reproduced this behaviour on a test project.  It seems Github changes the author when merging with the "squash and merge" option.

PR #1119, opened 2017-03-29
Feature/filter metrics

reftel:
This adds support for measuring time for the entire request processing, and the request and response filtering steps, in addition to the current measurement of the time used by the request method itself. In addition, support is added for letting users specify the clock to be used. This PR also includes a cherry-pick of randomstatic's thread-safe getters for metrics with custom instantiations from the 3.2 branch.

No extra metrics are added for methods with @Timed annotations that specify an explicit name, as it's not really clear what those extra metrics should be named, and it seems wrong to add attributes to the @Timed annotation for jersey-specific concepts. Ideas for how to add the extra metrics for explicitly named methods welcome!

Corresponding PR for 3.2: #1118

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics/MetricRegistry.java
@@ -141,6 +141,39 @@ public class MetricRegistry implements MetricSet {
     }
 
     /**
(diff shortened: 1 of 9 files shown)


arteam:
#1118 has been merged to 3.2. and will be included in the 4.0 release.

Issue #1120, opened 2017-04-03
How to set Allowed-Origin?


karlr42:
I have tried creating a class like this

````
@WebListener
public class MetricsConfig extends MetricsServlet.ContextListener {

	@Override
	protected String getAllowedOrigin() {
        return "*";
    }

	@Override
	protected MetricRegistry getMetricRegistry() {
		return new MetricRegistry();
	}
}
````

But it seems to have no effect, what do I need to do differently?

From what I've read, I might have to modify a web.xml file, but I don't have such a thing in my project - I'm using Tomcat 7.

karlr42:
Apologies, this does work just fine- I didn't realise we had already declared another listener in our project, and that in fact we do have an web.xml, where that listener class was declared. So I added my getAllowedOrigin implementation to that instead and it worked fine. Thanks.

For posterity if anyone finds this through search, this is how we got the Access-Control-Allow-Origin header set : 
````
public class MetricsServletContextListener extends MetricsServlet.ContextListener {

    @Override
    protected MetricRegistry getMetricRegistry() {
        return MetricsService.getInstance().getMetricRegistry();
    }
    
    @Override
	protected String getAllowedOrigin() {
        return "*";
    }

}
````
````
<listener>
    <listener-class>com.mycompany.metrics.MetricsServletContextListener</listener-class>
  </listener>
````

karlr42:
Actually, maybe this doesn't work- it adds the header to the GET, which is fine, but not to the OPTIONS request that Chrome makes in preflight, so the request still fails. Any thoughts?

PR #1121, opened 2017-04-04
Implement doOptions for CORS in MetricServlet

karlr42:
Set Access-Control-Allow-Origin on Options response headers if allowedOrigin is specified by a ContextListener.

Related to #1120

### DIFF
--- a/docs/source/about/release-notes.rst
+++ b/docs/source/about/release-notes.rst
@@ -4,12 +4,20 @@
 Release Notes
 #############
 
+.. _rel-3.2.2:
+
+v3.2.2: Mar 20 2017
+===================
+
+* Fix creating a uniform snapshot from a collection `#1111 <https://github.com/dropwizard/metrics/pull/1111>`_
+* Register metrics defined at Resource level `#1105 <https://github.com/dropwizard/metrics/pull/1105>`_
+
 .. _rel-3.2.1:
 
 v3.2.1: Mar 10 2017
 ===================
 
-* Suppot for shutting down the health check registry. `#1084 <https://github.com/dropwizard/metrics/pull/1084>`_
+* Support for shutting down the health check registry. `#1084 <https://github.com/dropwizard/metrics/pull/1084>`_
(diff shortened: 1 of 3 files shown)


karlr42:
I know you will want me to write a test, I'm just working on that now.

osigida:
I guess CORS headers should be managed differently, something like described here
http://jitterted.co.technology/tidbits/2014/04/04/handling-cors-in-dropwizard-and-jetty/

karlr42:
I will try that, but then what was the point of adding the ability to set Access-Control-Allow-Origin at the MetricsServlet level in #440 in the first place?

karlr42:
OK, so for my specific use case, using Tomcat, I actually was able to solve this by editing my web.xml
````
<filter>
	<filter-name>MetricsCorsFilter</filter-name>
	<filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
	<init-param>
    	<param-name>cors.allowed.origins</param-name>
    	<param-value>*</param-value>
    </init-param>
  </filter>
  <filter-mapping>
	<filter-name>MetricsCorsFilter</filter-name>
  	<url-pattern>/metrics/*</url-pattern>
  </filter-mapping>
````
So, using my application's server to handle CORS, like described in your link. This seems to work fine, it sets the appropriate headers on my OPTIONS requests. So I suppose this PR might not be needed.

But I still have the question, what is the point of setting Access-Control-Allow-Origin on the MetricsServlet directly? It doesn't seem like it would actually work because it doesn't support the OPTIONS method. It seems the correct way to do this is to set it up on Jetty/Tomcat/whatever, instead of the servlet, so maybe it makes sense to just remove this option entirely.

osigida:
@karlr42 good to know that it worked for you! 
I'm not sure why it was introduced... as main idea is to remove such a functionality form general endpoints (like metrics, swagger, etc) and handle it on the application level, where you can manage it based on your specific requirements.

arteam:
Please reopen if the issue still exists.

PR #1129, opened 2017-05-11
Check presence of JDK Class before instanciating JdkProvider

fabienbancharel:
Security needed, as Under Weblogic 10.3 Oracle Jdk 6, ClassNotFoundException is thrown only when current() is called, not at instanciation.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ThreadLocalRandomProxy.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ThreadLocalRandomProxy.java
@@ -35,8 +35,9 @@ class ThreadLocalRandomProxy {
     private static final Provider INSTANCE = getThreadLocalProvider();
     private static Provider getThreadLocalProvider() {
         try {
+            Class.forName("java.util.concurrent.ThreadLocalRandom");
             return new JdkProvider();
-        } catch (NoClassDefFoundError e) {
+        } catch (ClassNotFoundException e) {
             return new InternalProvider();
         }
     }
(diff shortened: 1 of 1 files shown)


ryantenney:
It appears there are one or more non-breaking spaces in the file:
`[ERROR] /home/travis/build/dropwizard/metrics/metrics-core/src/main/java/com/codahale/metrics/ThreadLocalRandomProxy.java:[38,1] illegal character: '\u00a0'`

fabienbancharel:
I used the Github online editor, and copy / pasting in Notepad++, I don't see any special character.
I will try a full clone on my laptop.

ryantenney:
Need to pull into 3.2 branch as well.

fabienbancharel:
I made a PR for 3.2-development branch : https://github.com/dropwizard/metrics/pull/1131

arteam:
Sorry it took so long to review. Thank you for testing the library under Weblogic! It's great to hear from users from different environments. Regarding the fix, I think a better approach would be to do the same as for `LongAdderProxy` and call `jdkProvider.current()`, this should trigger the class loading even in secure environments. We also could catch `Throwable` instead of `NoClassDefFoundError` to be more robust and not leave the INSTANCE variable as null.

fabienbancharel:
I had not looked at LongAdderProxy (since no problem with it), but yes, it would be consistent to use the same code. So, will you fix ThreadLocalRandom (at least in 3.1-maintenance), or do you expect me to commit it ? Thanks.

arteam:
Great! I will then prepare a pull request for the 3.1 branch and make a release!

arteam:
Closed via #1136

arteam:
I have release Metrics 3.1.5. It should be available in the Maven Central repository in a couple of hours.

fabienbancharel:
OK, 3.1.5 is validated on Weblogic 10.3. Thanks !

PR #1135, opened 2017-05-31
Exponentially decaying reservoir update

IstvanM:
Exponentially Decaying Reservoir was giving incorrect values in the snapshot if the inactive period was too long. This could result in wrong metrics aggregation.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/WeightedSnapshot.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/WeightedSnapshot.java
@@ -38,14 +38,13 @@ public class WeightedSnapshot extends Snapshot {
      * @param values    an unordered set of values in the reservoir
      */
     public WeightedSnapshot(Collection<WeightedSample> values) {
-        final WeightedSample[] copy = values.toArray( new WeightedSample[]{} );
-
         if (values.size()==1 && Double.compare(values.iterator().next().weight,0D)==0) {
             this.values = new long[0];
             this.normWeights = new double[0];
             this.quantiles = new double[0];
             return;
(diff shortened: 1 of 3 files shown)


IstvanM:
Thank you! Let me know if you need anything to be changed .

metrics-core/src/main/java/com/codahale/metrics/WeightedSnapshot.java
```diff
@@ -38,6 +38,12 @@ public WeightedSample(long value, double weight) {
      * @param values    an unordered set of values in the reservoir
      */
     public WeightedSnapshot(Collection<WeightedSample> values) {
+        if (values.size()==1 && Double.compare(values.iterator().next().weight,0D)==0) {
```
arteam: I think it's not very obvious why we perform this check in the constructor of the snapshot. I understand that the idea to avoid the situation when we have only one element in the sample with a 0 weight (because it provides bogus data), but I believe it should be performed in `ExponentiallyDecayingReservoir`, where the logic of rescaling resides. What do you think, if we add a guard condition:
```java
if (scalingFactor == 0) {
    values.clear();
}
``` 
in the `ExponentiallyDecayingReservoir#rescale` method? I think it could be a more clear solution.

IstvanM: Hi thanks, yes I will look into that today. I agree here it isn't the best to solve this problem.

IstvanM:
Should be done, can you check please?

arteam:
Thank you very much for the contribution!

PR #1140, opened 2017-06-14
getDefaultRegistry that does not throw

ellisjoe:


ellisjoe:
In library code I'd like to avoid using exceptions for control flow. Currently I have to write something like:
```
    private static MetricRegistry getMetricRegistry() {
        try {
            return SharedMetricRegistries.getDefault();
        } catch (IllegalStateException e) {
            return new MetricRegistry();
        }
    }
```

Ideally this method would return an Optional and I could then simplify the above to `SharedMetricRegistries.getDefault().orElseGet(MetricRegistry::new)`, but it looks like metrics still supports Java 7.

arteam:
LGTM! Thank you for the contribution! 

I think we could provide a similar method for `SharedHealthCheckRegistries` too.

PR #1141, opened 2017-06-15
Avoid npe in toString of HealthCheck.Result

georgespalding:
Use String.valueOf in toString of HealthCheck.Result to avoid NullPointerException when it has a detail with null value.
Prepend ", " to any details, to make the resulting String more readable.


georgespalding:
This issue burned me when debugging HealthChecks.
F.x doing `LOG.debug("What is happening {}", healthcheck)`
Causes the NPE to happen when enabling debug logging.

arteam:
LGTM. Thanks for submitting the fix!

PR #1144, opened 2017-06-16
ChunkedAssociativeLongArray traversable from both sides

storozhukBM:
This change makes `ChunkedAssociativeLongArray` traversable from both sides (from tail and
 from head). This actually simplifies the code and give us ability to optimize `ChunkedAssociativeLongArray.trim()` logic, what I actually did :smile:.

So now `SlidingTimeWindowArrayReservoir` speed is back and now it is even faster:
In ReservoirBenchmark: updates are **5.5x** faster and **~100x** less time spent in GC than `SlidingTimeWindowReservoir`
In SlidingTimeWindowReservoirsBenchmark: updates are **3.4x** faster, reads - **2.6x** faster, **35x** less time spent in GC.

With this changes `SlidingTimeWindowArrayReservoir` is actually comparable with `ExponentiallyDecayingReservoir` in terms of performance and now the only concern is memory overhead which is **~128** bits per stored measurement.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
@@ -1,16 +1,18 @@
 package com.codahale.metrics;
 
+import static java.lang.System.arraycopy;
 import static java.util.Arrays.binarySearch;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.ListIterator;
 
 class ChunkedAssociativeLongArray {
     private static final long[] EMPTY = new long[0];
     private static final int DEFAULT_CHUNK_SIZE = 512;
 
     private final int defaultChunkSize;
-    private Chunk activeChunk;
+    private final LinkedList<Chunk> chunks = new LinkedList<Chunk>();
 
     ChunkedAssociativeLongArray() {
         this(DEFAULT_CHUNK_SIZE);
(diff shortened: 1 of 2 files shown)


metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
```diff
@@ -1,95 +1,87 @@
...
+import java.util.LinkedList;
+import java.util.ListIterator;
 
 class ChunkedAssociativeLongArray {
     private static final long[] EMPTY = new long[0];
     private static final int DEFAULT_CHUNK_SIZE = 512;
 
     private final int defaultChunkSize;
-    private Chunk activeChunk;
+    private final LinkedList<Chunk> chunks = new LinkedList<Chunk>();
```
arteam: Would not be better to use an `ArrayList` instead of `LinkedList` if we care about reducing the amount of GC?

storozhukBM: LinkedList here has relatively small size 
`countOfStoredMeasurements / DEFAULT_CHUNK_SIZE`.
And we are heavily really on LinkedList implementation because:
1. Now we deleting chunks from both sides of the list  in` trim(long startKey, long endKey)` 
2. Deleting from and inserting chunks into the middle in `clear(long startKey, long endKey)`
And `LinkedList` give us O(1) complexity for all this operations.

arteam: Agreed. I run the benchmak and it seems `ÀrrayList` doesn't provide any advantages. Probably because of removing from the middle of the list during trimming of chunks.

arteam:
Looks good. Thanks for taking the time to make `ChunkedAssociativeLongArray` better!

storozhukBM:
Maybe we can add my answer as code comment for this line just to collection choice clear for other contributors.

PR #1145, opened 2017-06-18
ChunkedAssociativeLongArray re-use expired nodes

storozhukBM:
This change gives `ChunkedAssociativeLongArray` ability to store chunks that are expired and removed from main data structure.
Then instead of allocating new Chunk immediately we are trying to poll one from this cache.
So if you have constant or slowly changing load ChunkedAssociativeLongArray will never throw away old chunks or allocate new ones which makes this data structure almost garbage free. Also chunks stored in cache as `SoftReference`s so this memory can actually be freed if nesesary.

---
Numbers:
In `ReservoirBenchmark`: updates are **5.9x** faster and **~370x** less time spent in GC than `SlidingTimeWindowReservoir`.
In `SlidingTimeWindowReservoirsBenchmark`: updates are **3.2x** faster, reads - **2.7x** faster, **37x** less time spent in GC. (here we see small difference because most time we spent in cleaning snapshots)

---
Graphs:

_Comparison of implementations under constant load [duration 1h]_
constant load comparison

---
_Comparison of implementations under changing load [duration 1h]_
changing load comparison

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
@@ -16,22 +16,22 @@ class ChunkedAssociativeLongArray {
 
     private final int defaultChunkSize;
     /*
-    We use this ArrayDeque as cache to store chunks that are expired and removed from main data structure.
-    Then instead of allocating new Chunk immediately we are trying to poll one from this deque.
-    So if you have constant or slowly changing load ChunkedAssociativeLongArray will never
-    throw away old chunks or allocate new ones which makes this data structure almost garbage free.
+     * We use this ArrayDeque as cache to store chunks that are expired and removed from main data structure.
+     * Then instead of allocating new Chunk immediately we are trying to poll one from this deque.
+     * So if you have constant or slowly changing load ChunkedAssociativeLongArray will never
+     * throw away old chunks or allocate new ones which makes this data structure almost garbage free.
      */
     private final ArrayDeque<SoftReference<Chunk>> chunksCache = new ArrayDeque<SoftReference<Chunk>>();
 
     /*
-    Why LinkedList if we are creating fast data structure with low GC overhead?
-
-    First of all LinkedList here has relatively small size countOfStoredMeasurements / DEFAULT_CHUNK_SIZE.
-    And we are heavily rely on LinkedList implementation because:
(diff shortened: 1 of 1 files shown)


storozhukBM:
Hi @arteam,
It would be nice if you have a chance to look at this PR before release.

storozhukBM:
Here is great illustration of chunks reuse. [link](http://i.imgur.com/urPakEh.gifv)

storozhukBM:
Hi @ryantenney thank you for merge.
🎉

ryantenney:
I didn't merge, but only because I didn't want to screw up the release @arteam was planning! Just waiting for him to review and merge!

metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
```diff
@@ -22,11 +43,34 @@
         this.defaultChunkSize = chunkSize;
     }
 
+    private Chunk allocateChunk() {
+        SoftReference<Chunk> chunkRef = chunksCache.pollLast();
```
arteam: What do you think about simplifying this method to:
```java
while (true) {
  final SoftReference<Chunk> chunkRef = chunksCache.pollLast();
  if (chunkRef == null) {
    return new Chunk(defaultChunkSize);
  }
  final Chunk chunk = chunkRef.get();
  if (chunk != null) {
    chunk.cursor = 0;
    chunk.startIndex = 0;
    chunk.chunkSize = chunk.keys.length;
    return chunk;
  }
}
```

metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
```diff
@@ -3,15 +3,36 @@
...
 import java.util.ListIterator;
 
 class ChunkedAssociativeLongArray {
     private static final long[] EMPTY = new long[0];
     private static final int DEFAULT_CHUNK_SIZE = 512;
+    private static final int MAX_CACHE_SIZE = 128;
 
     private final int defaultChunkSize;
+    /*
+    We use this ArrayDeque as cache to store chunks that are expired and removed from main data structure.
```
arteam: Please use multi-line comments with an asterix on every line. We are trying to follow the Google Java Style Guide. See section about [comments](https://google.github.io/styleguide/javaguide.html#s4.8.6-comments).

storozhukBM:
Hi @arteam,
Thank you for review. All comments fixed.

PR #1148, opened 2017-06-22
Fixing issue with Graphite re-connects behind a dynamic DNS endpoint …

alexmnyc:
…when hostname+port is used

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
@@ -20,7 +20,7 @@ public class Graphite implements GraphiteSender {
 
     private final String hostname;
     private final int port;
-    private final InetSocketAddress address;
+    private InetSocketAddress address;
     private final SocketFactory socketFactory;
     private final Charset charset;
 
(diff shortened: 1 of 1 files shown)


metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
```diff
@@ -181,6 +181,11 @@ public void close() throws IOException {
         } finally {
             this.socket = null;
         }
+        
+        // if hostname is used, re-resolve DNS in connect() by clearing the address
+        if (this.hostname != null) {
```
arteam: I think if you specify a hostname and a port in the constructor, `Graphite` doesn't cache `InetSocketAddress` and it will make a DNS lookup everytime the the `connect` method is invoked.

alexmnyc: @arteam It looks like connect() will re-use the old address if it is not cleared - https://github.com/alexmnyc/metrics/blob/16679f3e5fddcfc79bbe51bfab0ca8ab9c377088/metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java#L114 no?

ryantenney: The `address` thats referring to is the local variable `address`, not the shadowed field `address`. It's confusing. But we never write to the field `address` if the class was initialized with a hostname, so the hostname will always be re-resolved (at least we will attempt to reresolve it, Java's DNS caching could be screwing that up).

alexmnyc: ok, yep. it was the shadowing that I did not see, thanks. Closing

Issue #832, opened 2015-07-16
JMX Refactor + Add healthchecks


ryantenney:
Extract JMX into a new module, and add HealthCheck support.

Contributions for this are welcome, it may not happen if I have to find time for it.

See #92 #149 #830

jgangemi:
:+1: for continued jmx support.

it's used heavily at the day job and i find it extremely useful for random side projects.

bripkens:
Same here! Having health checks available via JMX would enable us to easily extend our monitoring.

ariens:
+1 for continued JMX support--came here specifically looking for health check support.

gdubya:
+1 to what @ariens said

ryantenney:
Alright, JMX definitely stays.

sivasamyk:
I am trying to fix this issue. There are two options
1. Extend current JMXReporter with a new static forRegistry method with HealthCheckRegistry argument  and a new builder
2. A new HeathCheckJMXReporter class which will handle health check reporting.

I prefer option 1. Let me know your comments.

CodingFabian:
on this high level both sound kind of like the same for me. The second one sounds a bit more decoupled.

sivasamyk:
On exploring further I see that I cannot implement option#1 since health check is a different module. I am going ahead with option#2.

bripkens:
Woah, this is great to hear @sivasamyk. Thank you very much! :)

ryantenney:
Hi @sivasamyk! Thanks for working on this! I agree with option #2. If you aren't already I'd love to see JMXReporter and JMXHealthCheckReporter extracted to their own module, metrics-jmx.

sivasamyk:
sure, let me try it

sivasamyk:
Also moving JmxAttributeGauge class from metrics-core to metrics-jmx module. Any comments?

ryantenney:
That's a tough one. Sure why not.

robertjchristian:
Hi All,

Just checking in to see what the current status of "JMX support for HealthChecks" is.

I have a use case - use the Healthcheck module to expose a GREEN/RED/YELLOW-like status in JMX, with ability to manual re-run Health Checks from JMX console.  Keeping in mind that it can be expensive to run these tests, especially on a production system under load.

I'm considering using a Cache Gauge to work around the existing limitations.  This was suggested in a google groups thread that I'm now unable to find...

Any and all advice greatly appreciated.

Thanks,

Rob

[...]

PR #1162, opened 2017-07-19
Introducing a health check filter

arkigabor:
Adding a health check filter similar to metric filter. This will allow registering multiple health check servlets on different endpoints exposing different checks from a single registry. It could be useful when someone wants to distinguish critical checks  - that would signal the application is non-functional - from the less important ones.

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheckFilter.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheckFilter.java
@@ -1,0 +1,25 @@
+package com.codahale.metrics.health;
+
+/**
+ * A filter used to determine whether or not a health check should be reported.
+ */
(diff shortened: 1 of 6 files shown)


arkigabor:
Will there be another 3.2 release this could fit into?
I'm thinking about looking into JMX health check as well ( #832 ) and this could be useful there too.

ryantenney:
I'm afraid this can't go into 3.2, we (mostly) follow semantic versioning, so we can't change the API without going to 3.3 or 4.0.

arkigabor:
Yeah, I meant 3.x just for some reason wrote 3.2 instead. Rephrasing the question: do you plan any other minor releases before 4.0 or that is the only branch for feature PRs?

arteam:
The next release will be 4.0.0 and we plan only maintenance releases on 3.2.*. I've merged the pull requests against the 4.0-development branch. Thank you!

PR #1163, opened 2017-07-19
Remove non Jdk6 compatible letter from date pattern

arkigabor:
Fix to restore JDK6 compatibility broken by #609 #1077


arteam:
I will merge the pull requests, because it seems like a right change, but I am bit reluctant to make a new 3.2.* release only because of it. Nevertheless, thank you!

PR #1167, opened 2017-07-24
Fixing GraphiteReporter rate reporting

mgolahi:
Fixing rate reporting when rates are converted to a different TimeUnit. Proper integration can be seen here: https://github.com/dropwizard/metrics/blob/d170988536d07beb68431cecca2a468e51ae4581/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java#L277

Added unit test for verifying rate conversion.


arteam:
You're absolutely right! It was a regression introduced in fbb266a3dbc670642f73460c1470a0da16c50e54. Unfortunately, we didn't have a unit test for conversion with a custom rate and this mistake slipped through. Thank you very much for submitting a bugfix and unit test.

Issue #1168, opened 2017-07-27
ScheduledReporter#stop() throws uncaught CancellationException


garsEstTemps:
### Given:

The following code which creates a new GraphiteReporter using a custom executor service which we not want to stop when the reporter should be stoped.

```java
final PickledGraphite graphite = new PickledGraphite(new InetSockedAddress("host", port));
this.graphiteReporter = 
GraphiteReporter.forRegistry(getTelemetryMetricsRegistry())
                               .scheduleOn(myCustomExecutorService)
                               .shutdownExecutorOnStop(false)
                               .build(graphite);
```

### When:
Stopping the reporter

```java
this.graphiteReporter.stop();
```

### Then:
We got an uncaught `CancellationException` thrown by `java.util.concurrent.FutureTask.java`:

```java
    private V report(int s) throws ExecutionException {
        Object x = outcome;
        if (s == NORMAL)
            return (V)x;
        if (s >= CANCELLED)
            throw new CancellationException();
        throw new ExecutionException((Throwable)x);
    }
```

The exception is thrown because the `ScheduledReporter` is calling `FutureTask#get()` after calling `FutureTask#cancel()` and expects an `ExecutionException` to be thrown but not a `CancellationException`:

```java
public void stop() {
        if (shutdownExecutorOnStop) {
            [...] // shutdownExecutorOnStop is always false in our case
        } else {
            // The external manager(like JEE container) responsible for lifecycle of executor
            synchronized (this) {
                if (this.scheduledFuture == null) {
                    // was never started
                    return;
                }
                if (this.scheduledFuture.isCancelled()) {
                    // already cancelled
                    return;
                }
                // just cancel the scheduledFuture and exit
                this.scheduledFuture.cancel(false);
                try {
                    // Wait a while for existing tasks to terminate
                    scheduledFuture.get(1, TimeUnit.SECONDS);
                } catch (ExecutionException e) {
                    // well, we should get this error when future is cancelled normally, just ignore it
                } catch (InterruptedException e) {
                    // The thread was interrupted while waiting future to complete
                    // Preserve interrupt status
                    Thread.currentThread().interrupt();
                    if (!this.scheduledFuture.isDone()) {
                        LOG.warn("The reporting schedulingFuture is not cancelled yet");
                    }
                } catch (TimeoutException e) {
                    // The last reporting cycle is still in progress, nothing wrong, just add log record
                    LOG.warn("The reporting schedulingFuture is not cancelled yet");
                }
            }
        }
[...]

PR #1170, opened 2017-07-28
#1168 Fix uncaught CancellationException when stopping reporter with …

vladimir-bukhtoyarov:
Fix uncaught CancellationException when stopping reporter with externally managed executor:

- Do not try to wait the end of current reporting phase, when stopping reporter with externally managed executor, because JDK ScheduledExecutorService does not provide this feature, and [handy implementation of this feature](https://stackoverflow.com/a/45418900) requires big amount of complex concurrent code.
- Do not mock executors(where sensible) in order to make tests more realistic.


arteam:
Thank you!

PR #1179, opened 2017-08-28
Exposed Healthcheck objects from the registry based on name

ashisamazin:
We would like to get access to the HealthCheck object that has been added by (for example extending InjectableHealthcheck) to do some tests on the Healthcheck object. 

The particular use case we have is checking for the existence of an annotation on the HealthCheck object, so we can treat some healthchecks differently. 

I can't see any reason why these aren't exposed currently so this would be a big help.


arteam:

PR #1180, opened 2017-08-31
Add tracking for idle time in instrumentedExecutorService

keze:
At this point I can't know how much time did it take for task to be executed. I can only know how much time is spent on execution, but I don't know how long task was in queue. So I have added new Timer metric called idle which starts when task is created and stop when execution starts.


arteam:
Looks good to me. Thank you for the contribution!

keze:
Why isn't this pull request in 3.2.5 release?

arteam:
Sorry, I didn't make myself it clear. The pull request  has been merged for the 3.2 branch, but I moved it to the 4.0 branch and the commit actually included in the latest alpha release. The reason for that is the desire to maintain the 3.2 branch only for bugfixes and introduce features and improvements in the 4.0 branch which is compiled again JDK8.

PR #1184, opened 2017-09-06
common metric filters

bentatham:
This adds some common filters to the suite, besides just ``ALL``. Seems like these would be useful to many.  

I've left ``MetricName`` and ``String`` versions, because it seems many modules still use ``String`` rather than the new typed version.


arteam:
Could you please  re-submit your pull requests against the 4.0-development branch? The master branch is not maintained, I need to delete it.

bentatham:
@arteam : ok -- I'll work on that.  What happened to the support for "tags" (like was in ``MetricName``, etc), which I assume was designed to handle different data models, like influxdb.  Is that still on the plans on some branch?  Is there a roadmap document somewhere about the different versions?

PR #1189, opened 2017-09-12
Javadoc clarification: initialDelay time unit

timvlaer:
If you use the method start on ScheduledReporter with an initialDelay, it is not clear which time unit is used for this initial delay. I added a clarification in the JavaDoc.


arteam:
Thank you!

Issue #1176, opened 2017-08-18
InstrumentedScheduledExecutorService schedules FixedDelay tasks as FixedRate


ghost:
It means we're getting the wrong period semantics when scheduling some of our tasks using the wrapper

PR #1192, opened 2017-09-15
Fixing the scheduledFixedDelay to call the correct method

mveitas:
Fixes https://github.com/dropwizard/metrics/issues/1176
cc: at-alexb


arteam:
Thank you! I will backport it to 3.2.* and will make a new maintenance release.

PR #1195, opened 2017-09-15
Minor style cleanups (java 8 lambda, unused imports)

mveitas:
@arteam 

What are your thoughts on introducing something like [checkstyle](http://checkstyle.sourceforge.net/) into the project? Starting with some simple rules such as preventing unused imports and such?


arteam:
Thank you! 

> What are your thoughts on introducing something like checkstyle into the project? Starting with some simple rules such as preventing unused imports and such?

I think that's a good idea. We can copy the checkstyle [config](https://github.com/dropwizard/dropwizard/blob/master/checkstyle.xml) from the parent Dropwizard project and set up an integration with [Code Climate](https://codeclimate.com/github/dropwizard/dropwizard) to detect such issues automatically during a pull request.

PR #1196, opened 2017-09-15
Warnings cleanup

dpursehouse:


mveitas:
Looks like we crossed paths with #1195

dpursehouse:
@mveitas yes, looks like it, although mine also fixes some other things.

arteam:
Thank you both for contributing!

Issue #1174, opened 2017-08-11
CsvReporter should set separator in dependence of locale-dependent decimal separator


siom79:
Using the CsvReporter for example with a German locale, results in a CSV file that uses the comma as separator between the columns but additionally also as decimal separator:
```
1502448484,100,1,081546,0,610664,0,441010,0,097851,0,596013,0,643630,0,794799,0,971580,1,080447,1,081546,11,095617,0,090692,24,310302,69,889724,calls/minute,milliseconds
```
In the line above it is not clear where each column starts.

As the JDK class [DecimalFormatSymbols](http://docs.oracle.com/javase/6/docs/api/java/text/DecimalFormatSymbols.html) provides the decimal separator derived from the given locale, it should be possible to use a different char for the column separator. 

In general it would be useful to be able to provide the separator char for columns as parameter to the CsvReporter builder.

ryanrupp:
Alternatively or additionally, double quote values that can potentially contain a comma. Can be fancy and just check the decimal formatter output on startup to see if it's going to generate comma producing values e.g. US locale wouldn't need double quote ever.

PR #1197, opened 2017-09-15
Adding a custom separator for the CsvReporter

mveitas:
Fixes https://github.com/dropwizard/metrics/issues/1174

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/CsvReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/CsvReporter.java
@@ -16,7 +16,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
  * A reporter which creates a comma-separated values file of the measurements for each metric.
  */
 public class CsvReporter extends ScheduledReporter {
-    private static final char DEFAULT_SEPARATOR = ',';
+    private static final String DEFAULT_SEPARATOR = ",";
     
     /**
      * Returns a new {@link Builder} for {@link CsvReporter}.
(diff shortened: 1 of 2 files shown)


metrics-core/src/main/java/com/codahale/metrics/CsvReporter.java
```diff
@@ -220,7 +239,10 @@ private void reportTimer(long timestamp, String name, Timer timer) {
         report(timestamp,
                 name,
                 "count,max,mean,min,stddev,p50,p75,p95,p98,p99,p999,mean_rate,m1_rate,m5_rate,m15_rate,rate_unit,duration_unit",
-                "%d,%f,%f,%f,%f,%f,%f,%f,%f,%f,%f,%f,%f,%f,%f,calls/%s,%s",
+                "%d" + separator + "%f" + separator + "%f" + separator + "%f" + separator + "%f" + separator + "%f" +
```
arteam: What do you think about using `String.join` to avoid repetition? This method uses a `StringBuilder` under the hood, so it should provide similar performance as concatenation.

arteam (approved):
Looks good. I've left a suggestion to improve joining.

metrics-core/src/main/java/com/codahale/metrics/CsvReporter.java
```diff
@@ -180,8 +202,13 @@ private CsvReporter(MetricRegistry registry,
         super(registry, "csv-reporter", filter, rateUnit, durationUnit, executor, shutdownExecutorOnStop);
         this.directory = directory;
         this.locale = locale;
+        this.separator = separator;
         this.clock = clock;
         this.csvFileProvider = csvFileProvider;
+
+        this.histogramFormat = String.join(separator, "%d","%d","%f","%d","%f","%f","%f","%f","%f","%f","%f");
```
mveitas, line 209: While it was nice to have the headers and the format right next to each other in the code, there isn't a need to construct the same value over and for each call to `report`

arteam:
Thank you for the fixing the issue!

PR #1201, opened 2017-09-15
Adding checkstyle plugin and fixing some issues found

mveitas:
Adding in the checkstyle configuration file that is based upon [config](https://github.com/dropwizard/dropwizard/blob/master/checkstyle.xml) found in the Dropwizard project. I saw that checkstyle was not being run as part of the local build in the Dropwizard project. At the very least it should be made available to be run locally instead of just having it run on CodeClimate.

I had to comment out 2 rules (for now), but fixed a bunch of small things that CheckStyle found

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -1,6 +1,17 @@
 package com.codahale.metrics.graphite;
 
-import com.codahale.metrics.*;
+import com.codahale.metrics.Clock;
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.Metered;
+import com.codahale.metrics.MetricAttribute;
+import com.codahale.metrics.MetricFilter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.ScheduledReporter;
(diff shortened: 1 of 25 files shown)


metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
```diff
@@ -15,7 +26,21 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import static com.codahale.metrics.MetricAttribute.*;
+import static com.codahale.metrics.MetricAttribute.COUNT;
```
dpursehouse, line 29: Shouldn't static imports typically be listed before the normal imports?

pom.xml
```diff
@@ -326,6 +326,13 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
```
dpursehouse: Specify the version?

dpursehouse (changes requested):

metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
mveitas: I agree that they should be listed before the normal imports, that can be tackled in a different PR. This is just setting things up to use the same configuration as the Dropwizard project.

http://checkstyle.sourceforge.net/config_imports.html#CustomImportOrder

PR #1203, opened 2017-09-15
Fix style issues in CsvReporter pointed by Checkstyle

arteam:
Let's check the Code Climate integration: https://codeclimate.com/github/dropwizard/metrics


arteam:
Looks like it works, nice!

PR #1204, opened 2017-09-15
Fixing test sources directory checkstyle issues

mveitas:
Enable checkstyle to include the test sources directory in the plugin configuration. Found and fixed a number of warnings


dpursehouse (commented):
👍

PR #1206, opened 2017-09-18
Fix "Missing Fail" warnings reported by error-prone

dpursehouse:
As described in [1] the test is broken if fail() is not called because
it will still pass if the exception didn't get thrown.

In some of the cases this can be fixed by using JUnit's ExpectedException
rule. The limitation of this is that it can only be used when there are
no more statements in the test method following the statement that is
expected to throw. Those subsequent statements will never get executed
because the framework exits the test.

For tests where it's not possible to use ExpectedException due to the
reason mentioned above, add a call to fail().

[1] http://errorprone.info/bugpattern/MissingFail

Change-Id: I75eabf0bcb325fe385742f48aedee2938439b9da

### DIFF
--- a/metrics-core/src/test/java/com/codahale/metrics/SharedMetricRegistriesTest.java
+++ b/metrics-core/src/test/java/com/codahale/metrics/SharedMetricRegistriesTest.java
@@ -2,12 +2,17 @@ package com.codahale.metrics;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.Rule;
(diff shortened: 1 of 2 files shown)


dpursehouse:
@arteam I have a couple more commits fixing error-prone warnings; I'll push them as a separate PR after this one is completed.

dpursehouse:
Ping @arteam

arteam:
Thank you!

PR #1209, opened 2017-09-30
InstrumentedThreadFactoryTest: Fix "Future return value ignored"

dpursehouse:
As reported by error-prone:

  Return value of methods returning Future must be checked.
  Ignoring returned Futures suppresses exceptions thrown from the code
  that completes the Future.

  (see http://errorprone.info/bugpattern/FutureReturnValueIgnored)

Save the return value in a local. To ensure that we don't get another
"unused variable" warning, add a simple assertion that it's not null.

Remove the now redundant Eclipse warning suppression, which was not
effective for error-prone anyway.

Change-Id: I0d15c6312b2baae547a478ba828769a32c986a15


dpursehouse:
Ping @arteam any chance to merge this?

PR #1210, opened 2017-10-04
Add some common metric filters

bentatham:
This adds some common filters to the suite, besides just ALL. Seems like these would be useful to many.


arteam:
Thank you.

PR #1211, opened 2017-10-06
Unnecessary `clear` operation removal

storozhukBM:
Hi
Recently I've discussed `ChunkedAssociativeArray` implementation with my colleague and he had noticed that `ChunkedAssociativeArray` contains whole bunch of complicated code `clear(long startKey, long endKey)` to support one supper rare case of long overflow, that can happen only after 1 year after class loading. And even then, it is not so helpful, we will remove almost all values. 
So I've decided to simplify handling of this case 😄 .

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ChunkedAssociativeLongArray.java
@@ -177,77 +177,8 @@ class ChunkedAssociativeLongArray {
         }
     }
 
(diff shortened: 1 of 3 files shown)


storozhukBM:
Also, what do you think about replacing of `SlidingTimeWindowReservoir` by this `SlidingTimeWindowArrayReservoir` in 4.0 branch?

arteam:
LGTM, I think that's justified. I'm still reluctant to make `SlidingTimeWindowArrayReservoir` as default, though.

PR #1212, opened 2017-10-12
Add gauge for process CPU load to JVM attribute gauge set

apirom9:
Add gauge for process CPU load to JVM attribute gauge set


arteam:
Thanks for the pull requests! Unfortunately, I think I will reject it. My motivation:

* Users can use `JmxAttributeGauge` to registers metrics from JMX. 
* I think CPU load is not an attribute of the JVM, but rather an attribute of the operation system, so I'm reluctant to include it by default. There's a ton of tool for capturing OS-level metrics and for application-level CPU profiling we provide `CpuProfileServlet` which provides pprof statistic.

Issue #1217, opened 2017-11-02
ActorReporter for Akka


zharvey-welltok:
I have a need for DW Metrics to publish/report custom metrics to an Akka `ActorSystem` so that actors can handle them intelligently. My idea is to have simple/similar functionality as `ConsoleReporter`, except instead of writing metrics to the console I'd be sending them to some "*Metrics Actor*" (e.g. `metricsActor.tell(new MetricTickUpdate(...), null);`).

I'm gonna go forward with this `ActorReporter` concept for my project but would be willing to contribute if this community feels it would be a welcomed & useful addition.

PR #1222, opened 2017-11-08
Add metrics-akka module

bitbythecron:
Adding support for a new metrics-akka module per Issues #1217  and accompanying `AkkaReporter` (`ScheduledReporter` subclass) that pumps reported metrics to an actor thread living inside an Akka actor system.

The intention of this `AkkaReporter` is to be a lightweight abstraction layer between this DW Metrics library and the Akka actor that will be doing anything meaningful with the metrics (`SortedMap` instances). Therefore I intentionally omitted functionality such as checking for null metric maps, checking for duplicate key names (e.g. a meter and a histogram both have the same key/name), etc.

I'm open to any-and-all criticism and feedback and will make any changes this community feels are necessary.


bitbythecron:
Ah forgot one *very* important item, adding Akka as a dependency causes the Maven Enforcer plugin (`DependencyConvergence` rule) to fail. I offered several workarounds in #1217 and am happy to take the charge on any of them or any others this community would consider.

arteam:
Thanks for the pull request. Unfortunately, I can't accept it. I'm trying to keep the core metrics modules focused without adding anything new. I'm not an expert in Scala and Akka and I think maintaining this module here would be a big burden. Maybe you could release it as an independent module?

bitbythecron:
Thanks @arteam I appreciate the feedback and no worries at all. Tell you what, at some point soon I'll publish my new Akka reporter as its own tiny project and I'll send you the links so you know where to look if you ever change you mind ;-)

Cheers!

PR #1224, opened 2017-11-24
Add a method for timing non-throwing functions.

hamnis:
Allow methods that does not throw exceptions to use this avoiding a unnecessary try/catch block.


arteam:
LGTM. Thanks for the contribution!

PR #1229, opened 2017-12-12
Adds a module for instrumentation of Jdbi 3

arteam:
jdbi3 has been released a couple weeks ago (https://groups.google.com/forum/#!topic/jdbi/zS4p0p83sOM) and they use a new model for instrumentation. This change adds support for setting strategies for instrumenting Jdbi apps.

Implementation-wise this is a port of the `metrics-jdbi` module, improved, upgraded to Java8, and with added support for the `@Timed` annotation (previously it was not possible to set a custom metric name for an SqlObject class and method).

### DIFF
--- a/metrics-jdbi3/src/main/java/com/codahale/metrics/jdbi3/strategies/DefaultNameStrategy.java
+++ b/metrics-jdbi3/src/main/java/com/codahale/metrics/jdbi3/strategies/DefaultNameStrategy.java
@@ -1,0 +1,59 @@
+package com.codahale.metrics.jdbi3.strategies;
+
+import com.codahale.metrics.MetricRegistry;
+import org.jdbi.v3.core.extension.ExtensionMethod;
+import org.jdbi.v3.core.statement.StatementContext;
+
+/**
+ * Default strategies which build a basis of more complex strategies
+ */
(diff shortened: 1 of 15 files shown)


metrics-jdbi3/src/main/java/com/codahale/metrics/jdbi3/strategies/DefaultNameStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+public enum DefaultNameStrategy implements StatementNameStrategy {
+
+    /**
+     * If no SQL in the context, returns `sql.empty`, otherwise falls through
+     */
+    CHECK_EMPTY {
+        @Override
+        public String getStatementName(StatementContext statementContext) {
+            final String rawSql = statementContext.getRawSql();
+            if (rawSql == null || rawSql.length() == 0) {
```
jplock: `rawSql.isEmpty()` ?

jplock (approved):

metrics-jdbi3/src/main/java/com/codahale/metrics/jdbi3/strategies/DefaultNameStrategy.java
arteam: Will change!

Issue #1173, opened 2017-08-10
mean metric of Timer returns Double.NaN in some cases


altery:
There is a bug in WeightedSnapshot/ExponentiallyDecayingReservoir that causes the mean to return a value of NaN instead of 0.0 in some cases. The error occurs, if there are only 0.0 weights in the snapshot (ie. very old samples) while calculating the normWeights in WeightedSnapshot. I created a small example, that reproduces the problem: https://gist.github.com/altery/0efd06c4f8d0c83a44bd718e8170a49e

estekhin:
Seems related to #1005 which was fixed in  3.1.3, but I am getting a lot of timer.mean=NaN with 3.2.3 version.

altery:
Forgot to mention it: The problem occurs with 3.2.3.

philmtd:
Looks to me like #1135 let this issue appear.

I added #1230 to fix this.

arteam:
Closed via #1230

PR #1230, opened 2017-12-14
Prevent NaN values when calculating mean

philmtd:
This addresses the issue raised in #1173. 

Credits go to @altery for providing this example to reproduce the issue (which I used to write the unit test): https://gist.github.com/altery/0efd06c4f8d0c83a44bd718e8170a49e


arteam:
Thanks for the fix! It looks good to me. I think we could also add a check to `WeightedSnapshot` to avoid 0/0 situations which lead to NaN values.

arteam:
I will cherry-pick the fix to the 3.2 branch.

philmtd:
Cool, thanks!

PR #1231, opened 2017-12-18
Add module 'metrics-collectd' - Metrics Integration for Collectd

g-fresh:
This is a reporter which sends metrics to Collectd.

The original code was created by @dernasherbrezon and had already been merged to the master branch.

I did these modifications:
- Adapt the code for the 3.2 API
- Close channel after metrics were sent
- Add tests
- Support for specifying the local host name via the builder. (As a reverse lookup for the hostname might not always work.)


metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+        send(id, toByteArray(value), timestamp, dataType, period);
+    }
+
+    public void send(Identifier id, double value, long timestamp, DataType dataType, long period) {
+        send(id, toByteArray(value), timestamp, dataType, period);
+    }
+
+    private void send(Identifier identifier, byte[] value, long timestamp, DataType dataType, long period) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream(Network.BUFFER_SIZE);
+        DataOutputStream buffer = new DataOutputStream(baos);
```
arteam, line 86: Because we need to build a `ByteBuffer` to pass to `DatagramChannel` , we could create it directly without using `DataOutputStream` and `ByteArrayOutputStream`. `ByteBuffer` also provides methods for adding data of different types to the underlined byte array. Something like:
```java
ByteBuffer buffer = ByteBuffer.allocate(Network.BUFFER_SIZE);
writeString(buffer, Network.TYPE_HOST, identifier.host);
...
buffer.flip();
channel.send(buffer, address);
```

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+        writeHeader(buffer, type, len);
+        buffer.writeLong(val);
+    }
+
+    private static void writeHeader(DataOutputStream buffer, int type, int len) throws IOException {
+        buffer.writeShort((short) type);
+        buffer.writeShort((short) len);
+    }
+
+    private static byte[] toByteArray(long value) {
```
arteam, line 143: If we use `ByteBuffer` as the builder, we could avoiding creating an intermediate buffer by adding the value directly to the the `ByteBuffer`. The biggest issue is that it should be done at the end of building (when writing a binary representation of the value) instead of at the start. I think, we could change the signature of the `send` method by marking `value` as `Number` and make an `instanceof` check before writing the value of the `ByteBuffer`.  
```java
if (value instanceof Long) {
   buffer.putLong(value.longValue());
 } else {
   buffer.order(ByteOrder.LITTLE_ENDIAN).putDouble(value.doubleValue());
 }
```

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            writeString(buffer, Network.TYPE_HOST, identifier.host);
+            writeNumber(buffer, Network.TYPE_TIME, timestamp / 1000);
+            writeString(buffer, Network.TYPE_PLUGIN, identifier.plugin);
+            writeString(buffer, Network.TYPE_PLUGIN_INSTANCE, identifier.pluginInstance);
+            writeString(buffer, Network.TYPE_TYPE, identifier.type);
+            writeString(buffer, Network.TYPE_TYPE_INSTANCE, identifier.typeInstance);
+            writeNumber(buffer, Network.TYPE_INTERVAL, period);
+
+            int len = Network.HEADER_LEN + Network.UINT16_LEN + Network.UINT8_LEN + Network.UINT64_LEN;
```
arteam, line 96: I think `len` could be extracted as constant, so we don't recalculate it at every call.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+            writeHeader(buffer, Network.TYPE_VALUES, len);
+            buffer.writeShort((short) 1);
```
arteam, line 98: It would be nice to extract `1` to a constant with a meaningful name. Is it the data type length?

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+            LOG.error("unable to write bytearray", e);
+            return;
+        }
+        try {
+            buffer.write(value);
+            buffer.close();
+        } catch (IOException e) {
+            LOG.error("unable to write bytearray", e);
+        }
+        if (buffer.size() > Network.BUFFER_SIZE) {
```
arteam, line 110: If we use a `ByteBuffer` this check will become superfluous, because `ByteBuffer` can't grow and will throw an `IllegalArgumentException` if metric data exceeds its capacity. The biggest issue is that this exception will propagated instead of logged, but I think it's a good think, because I presume such failures can't be easily fixed.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Collectd.java
```diff
@@ -0,0 +1,156 @@
...
+            writeString(buffer, Network.TYPE_PLUGIN_INSTANCE, identifier.pluginInstance);
+            writeString(buffer, Network.TYPE_TYPE, identifier.type);
+            writeString(buffer, Network.TYPE_TYPE_INSTANCE, identifier.typeInstance);
+            writeNumber(buffer, Network.TYPE_INTERVAL, period);
+
+            int len = Network.HEADER_LEN + Network.UINT16_LEN + Network.UINT8_LEN + Network.UINT64_LEN;
+            writeHeader(buffer, Network.TYPE_VALUES, len);
+            buffer.writeShort((short) 1);
+            buffer.write(dataType.getCode());
+        } catch (IOException e) {
```
arteam, line 100: If we use `ByteBuffer` as the builder, we can remove this try/catch block, because `ByteBuffer` doesn't throw an `IOException` in its API.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/CollectdReporter.java
```diff
@@ -0,0 +1,228 @@
...
+        } catch (IOException e) {
+            LOG.warn("Unable to report to Collectd", e);
+        } finally {
+            disconnect(collectd);
+        }
+    }
+
+    private void reportTimer(String name, Timer timer, long timestamp) {
+        final Snapshot snapshot = timer.getSnapshot();
+        Identifier.Builder id = MetricNameMapping.createIdentifier(hostName, name);
```
arteam, line 166: nit: `Identifier.Builder id` is not reassigned and can be declared as `final` in the all `report` methods. Not a requirement, it would be nice to do so, because the rest of the codebase follows this style.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/CollectdReporter.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private void reportCounter(String name, Counter value, long timestamp) {
+        Identifier.Builder id = MetricNameMapping.createIdentifier(hostName, name);
+        collectd.send(id.get(), value.getCount(), timestamp, DataType.COUNTER, period);
+    }
+
+    private void reportGauge(String name, Gauge<?> value, long timestamp) {
+        Object object = value.getValue();
+        if (object instanceof Number) {
+            LOG.debug("Dropping metric '{}', which provides a value of unsupported type {}.",
```
arteam, line 213: I believe this debug log originally was intended for not-numbers instead of numbers. Anyway, I think debug logs should be removed from production code.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/CollectdReporter.java
```diff
@@ -0,0 +1,228 @@
...
+        }
+        Identifier.Builder id = MetricNameMapping.createIdentifier(hostName, name);
+        collectd.send(id.get(), ((Number) object).doubleValue(), timestamp, DataType.GAUGE, period);
+    }
+
+    private void disconnect(Collectd collectd) {
+        try {
+            collectd.disconnect();
+        } catch (Exception e) {
+            LOG.debug("Error disconnecting from Collectd: " + e.getMessage(), e);
```
arteam, line 224: I think it's better to log this message as the `warn` level. A failure to close the socket could be a symptom of a bigger networking issue.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/DataType.java
```diff
@@ -0,0 +1,17 @@
+package com.codahale.metrics.collectd;
+
+enum DataType {
+
+    COUNTER(0), GAUGE(1);
+
+    private final int code;
```
arteam, line 7: `code` is written to the message as a `byte`, so I think it's better to declare it in the enum as `byte` as well to avoid a narrowing primitive conversion from `int` to `byte` in runtime.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/MetricNameMapping.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.StringTokenizer;
+
+final class MetricNameMapping {
+
+    private static final String SEPARATOR = ".";
+
+    static Identifier.Builder createIdentifier(String host, String metricName) {
+        Identifier.Builder builder = new Identifier.Builder(host);
+        StringTokenizer tokenizer = new StringTokenizer(metricName, SEPARATOR);
```
arteam, line 11: I believe the whole thing can be simplified by using the `split` method of `String` and the joining collector from the Stream API: Something like that
```java
        String[] parts = metricName.split("\\.");
        switch (parts.length) {
            case 1:
                return builder.plugin(parts[0]).type(parts[0];);
            case 2:
                return builder.plugin(parts[0]).type(parts[1]);
            default:
                return builder.plugin(parts[0])
                        .pluginInstance(parts[1])
                        .type(Arrays.stream(parts)
                                .skip(2)
                                .collect(Collectors.joining(".")));
}
```

arteam: Maybe it also make sense to add a case when it's not possible to split the metric name (zero parts) and throw a meaningful exception.

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Network.java
```diff
@@ -0,0 +1,24 @@
+package com.codahale.metrics.collectd;
+
+/**
+ * Constants from collectd/src/network.h and jcd.* property configuration.
+ */
+class Network {
+
+    public static final int TYPE_HOST            = 0x0000;
```
arteam, line 8: I think the constants could be declared without the hexadecimal representation (it seems to me it doesn't add anything useful).

metrics-collectd/src/main/java/com/codahale/metrics/collectd/Sanitize.java
```diff
@@ -0,0 +1,43 @@
...
+
+    static String name(String name) {
+        return sanitize(name, DASH, SLASH, NULL);
+    }
+
+    static String instanceName(String instanceName) {
+        return sanitize(instanceName, SLASH, NULL);
+    }
+
+    private static String sanitize(String string, char... reservedChars) {
```
arteam, line 23: Instead of using varags, I think it's more reasonable to declare the reserverd chars as constants:
```java
private static final List<Character> INSTANCE_RESERVED = Arrays.asList(SLASH, NULL);
private static final List<Character> NAME_RESERVED = Arrays.asList(DASH, SLASH, NULL);
```
As the advantage, we can reuse the `contains` method of `List` instead wring an own over a char array.

metrics-collectd/src/test/java/com/codahale/metrics/collectd/CollectdReporterTest.java
```diff
@@ -0,0 +1,159 @@
...
+    @Test
+    public void handlesCounter() throws Exception {
+        Counter counter = registry.counter("api.rest.requests");
+        counter.inc(42);
+
+        reporter.report();
+
+        ValueList values = receiver.next();
+        assertNotNull(values.getHost());
+        assertEquals("api", values.getPlugin());
```
arteam, line 42: AssertJ assertions are more generally used in the project than the JUnit one. We use `AssertJ` for verifying `ValueList`, so for the consitency reasons it would be nice to use them in the all assertions.

metrics-collectd/src/test/java/com/codahale/metrics/collectd/CollectdReporterTest.java
```diff
@@ -0,0 +1,159 @@
...
+        assertEquals("count", values.getTypeInstance());
+        assertThat(receiver.next()).hasType("requests", "m1_rate");
+        assertThat(receiver.next()).hasType("requests", "m5_rate");
+        assertThat(receiver.next()).hasType("requests", "m15_rate");
+        assertThat(receiver.next()).hasType("requests", "mean_rate");
+    }
+
+    @Test
+    public void handlesGauge() throws Exception {
+        Gauge<Double> gauge = registry.register("gauge", new Gauge<Double>() {
```
arteam, line 70: We can create a `Gauge` with a lambda expression as `(Gauge<Double>) () -> 0.25`. Should be more concise and readable.

metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    protected void before() throws Throwable {
+        queue = new LinkedBlockingQueue<ValueList>();
+        receiver = new UdpReceiver(new QueueDispatcher(queue));
+        receiver.setPort(port);
+        start();
+    }
+
+    private void start() {
+        receiverThread = new Thread(this);
```
arteam, line 33: Instead of implementing `Runnable` by `Receiver`, we could pass the action directly to the `Thread` as a lambda expression.
```java
receiverThread = new Thread(() -> {
            try {
                receiver.listen();
            } catch (Exception e) {
                e.printStackTrace();
            }
        });
```

metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java
```diff
@@ -0,0 +1,81 @@
...
+    private BlockingQueue<ValueList> queue;
+
+    public Receiver(int port) {
+        this.port = port;
+    }
+
+    @Override
+    protected void before() throws Throwable {
+        queue = new LinkedBlockingQueue<ValueList>();
+        receiver = new UdpReceiver(new QueueDispatcher(queue));
```
arteam, line 27: We can avoid creating the `QueueDispatcher` class and just create a new anonymous `Dispatcher` class, because it's used only in one place.
```java
       new Dispatcher() {
            @Override
            public void dispatch(ValueList values) {
                queue.offer(new ValueList(values));
            }

            @Override
            public void dispatch(Notification notification) {
                throw new UnsupportedOperationException();
            }
        }
```

metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    protected void after() {
+        stop();
+    }
+
+    private void stop() {
+        receiver.shutdown();
+        try {
+            receiverThread.join();
```
arteam, line 58: It doesn't seem useful to `join` with the receiver thread. The main thread doesn't expect any results from it. We can just call an `interrupt` to signal the receiver thread to exit if it got stuck listening on the socket.

arteam (approved):
Looks good! I've left feedback regarding the implementation, but generally very good work and contribution.

g-fresh:
Hi @arteam ,

thanks a lot for your feedback.
I'll try to address the items as soon as possible.

PR #1232, opened 2017-12-18
Jetty9: unhandled response should be counted as 404 and not 200

swallez:
In Jetty, a request that has not been handled by the handler chain keeps the default status code 200, and a 404 response is produced by `HttpChannel`. The Jetty9 `InstrumentedHandler` therefore has to count unhandled request as 404 and not 200.

To reveal the issue, this PR removes the 404 response from `InstrumentedHandlerTest.TestHandler` so that a request to `/hello` ends up not being handled.

This PR is against branch `3.2-development`. Should I open another PR for branch `4.0-development`?


arteam:
Thanks for spotting this bug! I agree that an unhandled request should be counted as an error (because there's no handler that handles the request). 

I'm still maintaining the 3.2 branch, so the fix is good enough to merge to it. I will also cherry-pick it to the 4.0 branch.

swallez:
Awesome! Thanks @arteam!

PR #1239, opened 2017-12-23
Add Bill of Materials for Metrics

### LINKED ISSUE #1169: BOM for metrics?
I noticed there is no bom for metric artefacts, it could be nice to have one.

For example latest dropwizard uses metrics 3.2.2 and 3.2.3 was released: if I want to use latest version I have to add lots of dependency management just to force the version…

arteam:
Metrics is a multi-module project with a lot of interproject dependencies. Also there's a lot of Metrics version in the wild (3.1.2, 3.1.5, 3.2.3, 3.2.5, 4.0.0). Many projects often depend on metrics as
a 3rd-party dependency. It would be nice if Metrics users could specify a metrics version for all dependencies consistently via a BOM dependency without manual twiddling with excludes.

Fixes #1169


metrics-bom/pom.xml
```diff
@@ -0,0 +1,286 @@
...
+                    <exclusion>
+                        <groupId>org.slf4j</groupId>
+                        <artifactId>slf4j-api</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>${http-client.version}</version>
```
scottcarey, line 167: this bom absolutely should not have included things that are not part of the build (dropwizard stuff)

Including all the third party things makes this inconsumable for external users if even one of those things clashes with another library, even if we aren't using the submodule that depends on it.

This pom should have been to help external users manage the metrics dependencies, not all transitive ones.

Useless.  I'll have to copy paste the top part.

PR #1247, opened 2018-01-04
Throw on existing healt check name, fixes #1245

### LINKED ISSUE #1245: Inconsistency registring gauges and health checks
When once registers a gauge under the name of an already registered metric, this fails with an exception. A similar operation; registering a health check under the name of an already registered health check fails silently, its a no-op.

You could argue either way but I think that `HealthCheckRegistry.register` should throw when the name is already in use.

erikvanoosten:
Throw an exception when registering a health check under an already existing name. This fixes #1245.

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheckRegistry.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheckRegistry.java
@@ -95,21 +95,18 @@ public class HealthCheckRegistry {
      * @param healthCheck the {@link HealthCheck} instance
      */
     public void register(String name, HealthCheck healthCheck) {
-        HealthCheck registered = null;
+        HealthCheck registered;
         synchronized (lock) {
-            if (!healthChecks.containsKey(name)) {
-                registered = healthCheck;
-                if (healthCheck.getClass().isAnnotationPresent(Async.class)) {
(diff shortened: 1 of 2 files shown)


metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheckRegistry.java
```diff
@@ -105,7 +105,9 @@ public void register(String name, HealthCheck healthCheck) {
                 healthChecks.put(name, registered);
             }
         }
-        if (registered != null) {
+        if (registered == null) {
+            throw new IllegalArgumentException("A health check named " + name + " already exists");
+        } else {
```
arteam: It seems to me that it would be better to move throwing exception near to the `containsKey` check. It looks more logical (we check that the health check doesn't exist, otherwise throw an exception). 
```java
if (!healthChecks.containsKey(name)) {
  ....
} else {
  throw new IllegalArgumentException("A health check named " + name + " already exists");
}
```

Also this allows to remove the not null check before invoking the listeners, since the compiler will guarantee that `registered` will be always not null.

arteam (approved):
Looks good! I've left one suggestion.

arteam:

erikvanoosten:
Thanks for merging so quickly 👍

PR #1250, opened 2018-01-05
Introduce MetricName to encapsulate more info about metrics

arteam:
Original contribution is in #561 by @udoprog.

> This is a large refactoring meant to change the way metric names are represented to better suit a wide variety of backends.
> 
> Quite a few metric backend systems are starting to use a general concept of 'tags' which is a step back from having too encode information in the (string) names of the metrics:
> 
> * KairosDB (http://code.google.com/p/kairosdb/),
> * OpenTSDB (http://opentsdb.net/overview.html)
> 
> Tags are meant to allow for natural grouping of similar timeseries allowing for the ability to easily correlate freely between the timeseries that incorporate them.
> 
> For example, take the following two timeseries.
> 
> * example.com.interface.traffic.eth0.in
> * example.com.interface.traffic.eth0.out
> 
> Using tags, these would be represented as;
> 
> * interface.traffic {host=example.com, interface=eth0, direction=in}
> * interface.traffic {host=example.com, interface=eth0, direction=out}

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java
@@ -223,18 +223,18 @@ public class ConsoleReporter extends ScheduledReporter {
 
     @Override
     @SuppressWarnings("rawtypes")
-    public void report(SortedMap<String, Gauge> gauges,
-                       SortedMap<String, Counter> counters,
-                       SortedMap<String, Histogram> histograms,
-                       SortedMap<String, Meter> meters,
-                       SortedMap<String, Timer> timers) {
+    public void report(SortedMap<MetricName, Gauge> gauges,
+                       SortedMap<MetricName, Counter> counters,
+                       SortedMap<MetricName, Histogram> histograms,
+                       SortedMap<MetricName, Meter> meters,
+                       SortedMap<MetricName, Timer> timers) {
         final String dateTime = dateFormat.format(new Date(clock.getTime()));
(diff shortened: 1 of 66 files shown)


qualidafial:
Looking at the `MetricName` class: not all tags out there are key/value pairs. Some are just labels--they are either present or not.

Datadog recognizes either of these concepts as "tags", which are all just strings, but it recognizes those formatted as "key:value" and lets you search on them and see comparative metrics for different values.

arteam:
Thanks for the observation. I'm wondering if we can add some specific marker for a value (to distinguish between a label and a key-value pair).

qualidafial:
Possibly `tags` could be key/value, and `labels` could be plain Strings?

udoprog:
I currently believe that there's nothing you can do with free-floating labels, that you can't do with key-value pairs _better_.

For example, `good`, `bad` are labels, they could instead be `status: good`, and `status: bad`. The latter communicates _what_ is good or bad (the status). Without this information the data is free-floating and prone to conflicts. Less descriptive.

I don't think free-floating labels are well supported in TSDBs (Prometheus, infuxdb, Heroic), or UIs (mainly Grafana). But this I'm less sure about now. Datadog might have supported them at some point.

@qualidafial Do you have an example of a label and TSDB combo where this is useful?

qualidafial:
I think you've convinced me.

I've also realized that `MetricName` could always get a new `labels` property in the future if needed, so there's no risk to leaving it out now.

arteam:
Okay, then I'm going to merge this PR.

Issue #1248, opened 2018-01-05
Change package for new major releases (4.x.x, 5.x.x)


udoprog:
Changing the package name when doing major releases is critical to avoid a special kind of dependency hell for libraries.

There are two concepts here which collide:
* A major change implies a [binary compatibility break](https://wiki.eclipse.org/Evolving_Java-based_APIs_2).
* Libraries are transitively embedded into the dependencies of any project.

Let's say you have a dependency graph which contains both `3.x.x` and `4.x.x`. Because the fully qualified name of the classes conflict, only one will be used due to how the classpath _currently_ works. If it attempts to do something which is not compatible with the given major version that they are using code interacting with this class will crash at runtime. This is especially egregious if it happens because one of your dependencies use one version or the other.

By not having separate packages for the different major versions, it is now exceedingly difficult to introduce compatibility layers as the ecosystem is moving from one version to another. Given a [hard enough incompatibility](https://groups.google.com/forum/#!topic/protobuf/cXbnth6QVnU), you'll end up in a catch 22 where no one is able to upgrade without resorting to something like shading every dependency and the entire ecosystem is stuck in a lock step of major upgrades.

With different package names we can do things like introduce a shim metric registry that provide an API for a previous version. This can then be used by the application providing the MetricRegistry to decide which version of the library to use.

For more detailed information, I've written about this topic in the past: https://udoprog.github.io/programming/2015-09-03/semantic-versioning-java.html

Note that this also applies to having a distinct *groupId*:*artifactId* combo for maven packages, since that is how it currently prunes its dependency graph. This did change from 3.x.x -> 4.x.x, but currently doesn't seem to be the direction with 5.x.x: https://github.com/dropwizard/metrics/blob/5.0-development/metrics-core/pom.xml#L6

RxJava is a good example of a project that took this seriously for their 1.x -> 2.x transition: https://github.com/ReactiveX/RxJava

P.S. It should be noted that some libraries are less offensive in this regard, since they primarily exist in the periphery in the dependency graphs, and types provided by them aren't shared as much.

arteam:
Thanks for the very insightful comment!

[...]

PR #1256, opened 2018-01-09
Change Maven group to io.dropwizard.metrics5

arteam:
This should provide an easier transition from Metrics 3/4 to Metrics 5 and prevent conflicts in case somebody by accident will include different version of metrics in the classpath.

See #1248


morpheus9999:
Doesn't make any sense...

PR #1268, opened 2018-01-26
Provide a type-safe way to create a custom counter

### LINKED ISSUE #1261: API improvements
I have a suggestion to improve your
`Counter counter(MetricName name, final MetricSupplier<Counter> supplier) `
style methods in MetricRegistry. I see two drawbacks for a consumer:
1) I would naturally tend to write `MetricSupplier<MyCounter>` inner class rather than `MetricSupplier<Counter>`
2) I have to cast the result to MyCounter (MyCounter might extend Counter with some extra methods I need to call).

You could consider
`<T extends Counter> T counter(String name, final MetricSupplier<T> supplier)`
instead. This overcomes both of the above mentioned limitations. Further, you can have a stricter `isInstance()` method in your anon MetricBuilder than just `Counter.class.isInstance(metric)`. It is likely that the actual supplier impl will have a literal generic type, e.g. `MetricSupplier<MyCounter>`. Using reflection, it is possible to extract this from the declaration, you could then do `declaredCounterType.isAssignableFrom(metric.getClass())`
In the worst case, you have to fallback to Counter.class.

arteam:
Practically, the only way to customize a counter is too extend it. That means users who provide a new supplier will probably use a new class that extends `Counter`. It would be nice if the `counter(MetricName name, MetricSupplier<Counter> supplier)` method returned a counter with the same type as specified in the `MetricSupplier` implementation. This will save a redundant type cast for the user and avoid a potential compilation warning. Since it's a breaking change, it can be done only in the 5.0 version.

Fixes #1261.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics5/MetricRegistry.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics5/MetricRegistry.java
@@ -144,10 +144,10 @@ public class MetricRegistry implements MetricSet {
      * @param supplier a MetricSupplier that can be used to manufacture a counter.
      * @return a new or pre-existing {@link Counter}
      */
-    public Counter counter(MetricName name, final MetricSupplier<Counter> supplier) {
-        return getOrAdd(name, new MetricBuilder<Counter>() {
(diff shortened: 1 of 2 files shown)


pulquero:
Looks good, but how about the other metric types?

PR #1269, opened 2018-02-02
InfluxDB reporter for metrics 5.0

mbrannstrom:
Metric reporter for influxDB over UDP.

- Native **support for tags** in MetricName.
- **Low pressure on the GC** in the reporter, i.e. avoid creating garbage.

With the addition of tag support in MetricName 5.0.0 I think it is reasonable to add a reporter that is capable of using those tags natively. While there are a few 3rd party metrics-influxdb implementations out there, many of these fail to keep up to date with the latest metrics version.

### DIFF
--- a/metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
+++ b/metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
@@ -17,41 +17,11 @@ import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
+abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
 
-    private MetricRegistry registry;
-    private RegistryMirror mirror;
+    private final MetricRegistry registry;
+    private final RegistryMirror mirror;
     
-    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
-            String name, 
-            MetricFilter filter, 
-            TimeUnit rateUnit, 
-            TimeUnit durationUnit) {
-        super(registry, name, filter, rateUnit, durationUnit);
-        init(registry, filter);
-    }
-
-    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
-            String name, 
-            MetricFilter filter, 
-            TimeUnit rateUnit, 
-            TimeUnit durationUnit, 
-            ScheduledExecutorService executor) {
-        super(registry, name, filter, rateUnit, durationUnit, executor);
-        init(registry, filter);
-    }
-
-    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
-            String name, 
-            MetricFilter filter, 
-            TimeUnit rateUnit, 
-            TimeUnit durationUnit, 
-            ScheduledExecutorService executor, 
-            boolean shutdownExecutorOnStop) {
-        super(registry, name, filter, rateUnit, durationUnit, executor, shutdownExecutorOnStop);
-        init(registry, filter);
-    }
-
     protected GarbageFreeScheduledReporter(MetricRegistry registry, 
             String name,
             MetricFilter filter, 
@@ -61,13 +31,9 @@ public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
             boolean shutdownExecutorOnStop, 
             Set<MetricAttribute> disabledMetricAttributes) {
         super(registry, name, filter, rateUnit, durationUnit, executor, shutdownExecutorOnStop, disabledMetricAttributes);
-        init(registry, filter);
-    }
-    
-    private void init(MetricRegistry registry, MetricFilter filter) {
         this.registry = registry;
         this.mirror = new RegistryMirror(filter);
-        mirror.register(registry);
+        registry.addListener(mirror);
     }
 
     @Override
@@ -75,7 +41,7 @@ public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
         try {
             super.stop();
         } finally {
-            mirror.unregister(registry);
+            registry.removeListener(mirror);
         }
     }
 
@@ -91,10 +57,10 @@ public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
         }
     }
     
+    @SuppressWarnings("rawtypes") // because of signature (for Gauge) in ScheduledReporter#report(..)
     private static class RegistryMirror implements MetricRegistryListener {
 
         private final MetricFilter filter;
-        @SuppressWarnings("rawtypes") // because of signature in ScheduledReporter#report(..)
         private final ConcurrentSkipListMap<MetricName, Gauge> gauges = new ConcurrentSkipListMap<>();
         private final ConcurrentSkipListMap<MetricName, Counter> counters = new ConcurrentSkipListMap<>();
         private final ConcurrentSkipListMap<MetricName, Histogram> histograms = new ConcurrentSkipListMap<>();
@@ -105,14 +71,6 @@ public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
             this.filter = filter;
         }
         
-        public void register(MetricRegistry registry) {
-            registry.addListener(this);
-        }
-        public void unregister(MetricRegistry registry) {
-            registry.removeListener(this);
-        }
-
-        @SuppressWarnings("rawtypes") // because of signature in ScheduledReporter#report(..)
         SortedMap<MetricName, Gauge> gauges() {
             return gauges;
         }
--- a/metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
+++ b/metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
@@ -1,0 +1,199 @@
+package io.dropwizard.metrics5.influxdb;
+
+import io.dropwizard.metrics5.Counter;
+import io.dropwizard.metrics5.Gauge;
+import io.dropwizard.metrics5.Histogram;
+import io.dropwizard.metrics5.Meter;
+import io.dropwizard.metrics5.MetricAttribute;
+import io.dropwizard.metrics5.MetricFilter;
+import io.dropwizard.metrics5.MetricName;
+import io.dropwizard.metrics5.MetricRegistry;
+import io.dropwizard.metrics5.MetricRegistryListener;
+import io.dropwizard.metrics5.ScheduledReporter;
+import io.dropwizard.metrics5.Timer;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public abstract class GarbageFreeScheduledReporter extends ScheduledReporter {
+
+    private MetricRegistry registry;
+    private RegistryMirror mirror;
+    
+    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
+            String name, 
+            MetricFilter filter, 
+            TimeUnit rateUnit, 
+            TimeUnit durationUnit) {
+        super(registry, name, filter, rateUnit, durationUnit);
+        init(registry, filter);
+    }
+
+    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
+            String name, 
+            MetricFilter filter, 
+            TimeUnit rateUnit, 
+            TimeUnit durationUnit, 
+            ScheduledExecutorService executor) {
+        super(registry, name, filter, rateUnit, durationUnit, executor);
+        init(registry, filter);
+    }
+
+    protected GarbageFreeScheduledReporter(MetricRegistry registry, 
+            String name, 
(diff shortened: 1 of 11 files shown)


mbrannstrom:
The memory turnover caused by calling `MetricRegistry#getGauges(MetricFilter)` etc (copying maps) from `ScheduledReporter` have now been eliminated by the class `GarbageFreeScheduledReporter`.

The assumption is that metrics are reported more often than they are added/removed.

Added in commit 71a4321.

arteam:
At a first glance looks great! I'm going to review the PR during the weekend.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
```diff
@@ -0,0 +1,199 @@
...
+            MetricFilter filter, 
+            TimeUnit rateUnit, 
+            TimeUnit durationUnit, 
+            ScheduledExecutorService executor, 
+            boolean shutdownExecutorOnStop) {
+        super(registry, name, filter, rateUnit, durationUnit, executor, shutdownExecutorOnStop);
+        init(registry, filter);
+    }
+
+    protected GarbageFreeScheduledReporter(MetricRegistry registry,
```
arteam: I think it should be enough to have only this constructor implemented in `GarbageFreeScheduledReporter`. All other constructors can be derived from it. At any case, reporters are built using builders which invoke only the most detailed constructor. Another plus that we can declare `registry` and `mirror` as final.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
```diff
@@ -0,0 +1,199 @@
...
+                    mirror.histograms(), 
+                    mirror.meters(), 
+                    mirror.timers());
+        }
+    }
+    
+    private static class RegistryMirror implements MetricRegistryListener {
+
+        private final MetricFilter filter;
+        @SuppressWarnings("rawtypes") // because of signature in ScheduledReporter#report(..)
```
arteam: I guess we could be less specific and just to place this annotation over the class.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/GarbageFreeScheduledReporter.java
```diff
@@ -0,0 +1,199 @@
...
+        private final ConcurrentSkipListMap<MetricName, Counter> counters = new ConcurrentSkipListMap<>();
+        private final ConcurrentSkipListMap<MetricName, Histogram> histograms = new ConcurrentSkipListMap<>();
+        private final ConcurrentSkipListMap<MetricName, Meter> meters = new ConcurrentSkipListMap<>();
+        private final ConcurrentSkipListMap<MetricName, Timer> timers = new ConcurrentSkipListMap<>();
+
+        public RegistryMirror(MetricFilter filter) {
+            this.filter = filter;
+        }
+        
+        public void register(MetricRegistry registry) {
```
arteam: I don't see much advantage to implement `register` and `unregister` methods. ` registry.removeListener(mirror)` and `mirror.unregister(registry)` looks semantically same to me.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbReporter.java
```diff
@@ -0,0 +1,517 @@
...
+        }
+        writeTimestampMillis(timestamp);
+        reportLine();
+    }    
+    
+    private void reset() {
+        str.delete(0, str.length());
+    }
+    
+    private void writeMeasurement(MetricName name) {
```
arteam: I think it makes to extract all the methods and buffer to a new class like `InfluxDbLineBuilder` or so. Like
```java
class InfluxDbLineBuilder {

    private StringBuilder str = new StringBuilder();
    private boolean firstField;

    private final Set<MetricAttribute> disabledMetricAttributes;
    private final MetricName prefix;
    private final Map<MetricName, String> encodedNameCache = new WeakHashMap<>();

    InfluxDbLineBuffer(Set<MetricAttribute> disabledMetricAttributes, MetricName prefix) {
        this.disabledMetricAttributes = disabledMetricAttributes;
        this.prefix = prefix;
    }

    InfluxDbLineBuffer writeMeasurement(MetricName name) {
        str.delete(0, str.length());
        str.append(encodedNameCache.computeIfAbsent(name, this::writeMeasurementNoCache));
        str.append(' ');
        firstField = true;
        return this;
    }

    public InfluxDbLineBuffer reset() {
        str.delete(0, str.length());
        return this;
    }

    InfluxDbLineBuffer writeField(MetricAttribute key) {
        if (!firstField) {
            str.append(',');
        }
        str.append(key.getCode()).append('=');
        firstField = false;
        return this;
    }

    InfluxDbLineBuffer writeField(String key) {
        if (!firstField) {
            str.append(',');
        }
        appendName(key, str);
        str.append('=');
        firstField = false;
        return this;
    }

    InfluxDbLineBuffer writeFieldValue(double value) {
        str.append(value);
        return this;
    }

    InfluxDbLineBuffer writeFieldValue(long value) {
        str.append(value).append('i');
        return this;
    }

    InfluxDbLineBuffer writeFieldValue(String value) {
        str.append('"');
        appendString(value, str);
        str.append('"');
        return this;
    }

    InfluxDbLineBuffer writeFieldValue(boolean value) {
        str.append(value ? 't' : 'f');
        return this;
    }

    InfluxDbLineBuffer writeTimestampMillis(long utcMillis) {
        str.append(' ').append(utcMillis).append("000000\n");
        return this;
    }

    boolean hasValues() {
        return !firstField;
    }

    InfluxDbLineBuffer writeFieldIfEnabled(MetricAttribute key, double value) {
        if (!disabledMetricAttributes.contains(key)) {
            writeField(key);
            writeFieldValue(value);
        }
        return this;
    }

    InfluxDbLineBuffer writeFieldIfEnabled(MetricAttribute key, long value) {
        if (!disabledMetricAttributes.contains(key)) {
            writeField(key);
            writeFieldValue(value);
        }
        return this;
    }

    public StringBuilder get() {
        return str;
    }
```

arteam: The idea to not mix the logic of the reporter (reading information about gauges, metrics, timers) from the InfluxDb protocol implementation. The reporter should now know about caches and whether we reuse `StringBuilder` or create a new one. I think it would be nice to see building an InfluxDb metric line as a sequence of calls:
```java
 builder.writeMeasurement(name)
            .writeFieldIfEnabled(COUNT, meter.getCount())
            .writeFieldIfEnabled(SUM, meter.getSum())
            .writeFieldIfEnabled(M1_RATE, convertRate(meter.getOneMinuteRate()))
            .writeFieldIfEnabled(M5_RATE, convertRate(meter.getFiveMinuteRate()))
            .writeFieldIfEnabled(M15_RATE, convertRate(meter.getFifteenMinuteRate()))
            .writeFieldIfEnabled(MEAN_RATE, convertRate(meter.getMeanRate()))
            .writeTimestampMillis(timestamp)
```

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbReporter.java
```diff
@@ -0,0 +1,517 @@
...
+            nameChars = new char[str.length()];
+            str.getChars(0, str.length(), nameChars, 0);
+            encodedNameCache.put(name, nameChars);
+        }
+        
+        str.append(' ');
+        firstField = true;
+    }
+    
+    private void writeMeasurementNoCache(MetricName name) {
```
arteam: If we cache names of metrics, I think it's better to store them as strings instead of char arrays. Also I don't think it makes much sense to cache a value and at the time write it to the current `StringBuilder`. I would expect that we just read the data from the cache and if the key is not present, we load it: `str.append(encodedNameCache.computeIfAbsent(name, this::toInfluxMetricName))`

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbReporter.java
```diff
@@ -0,0 +1,517 @@
...
+        MetricName prefixedName = MetricName.join(prefix, name);
+        
+        appendName(prefixedName.getKey(), str);
+        Map<String, String> tags = prefixedName.getTags();
+        if (!tags.isEmpty()) {
+            // InfluxDB Performance and Setup Tips:
+            // Sort tags by key before sending them to the database. 
+            // The sort should match the results from the Go bytes.Compare function.
+            TreeMap<String, String> sortedTags = new TreeMap<>(tags);
```
arteam: I think we could take advantage of the Stream API and use it for building the metric name in a declarative fashion:
```java
MetricName prefixedName = MetricName.join(prefix, name);
return prefixedName.getTags()
          .entrySet()
          .stream()
          .sorted(Comparator.comparing(Map.Entry::getKey))
          .reduce(appendName(prefixedName.getKey(), new StringBuilder()), (sb, e) -> {
                    sb.append(',');
                    appendName(e.getKey(), sb);
                    sb.append('=');
                    return appendName(e.getValue(), sb);
          }, (l, r) -> l)
          .toString();
```

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbReporter.java
```diff
@@ -0,0 +1,517 @@
...
+            } else {
+                writeFieldValue(value.toString());
+            }
+        }
+        writeTimestampMillis(timestamp);
+        reportLine();
+    }    
+    
+    private void reset() {
+        str.delete(0, str.length());
```
arteam: Looks we could just use `str.setLength(0)` to reset the counter in `StringBuilder`.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbReporter.java
```diff
@@ -0,0 +1,517 @@
...
+        appendString(value, str);
+        str.append('"');
+    }
+    
+    private void writeFieldValue(boolean value) {
+        str.append(value ? 't' : 'f');
+    }
+    
+    private static void appendName(CharSequence field, StringBuilder dst) {
+        int len = field.length();
```
arteam: Looks like that the `len` variable is redundant. We could directly use `field.length` in the loop condition.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbUdpSender.java
```diff
@@ -0,0 +1,148 @@
...
+            byteBuf.clear();
+        }
+    }
+    
+    private void sendBuffer() throws IOException {
+        byteBuf.flip();
+        datagramChannel.send(byteBuf, address);
+    }
+        
+    private void encode(StringBuilder str) {
```
arteam, line 119: I wonder if all these low-level manipulation are worth. Looks like a lot of code instead of:
```
datagramChannel.send(encoder.encode(CharBuffer.wrap(str)), address)
```

I know this creates a new byte array every time, but the question is whether it's such a big issue. Or maybe we could build a line as a `ByteBuffer` right away instead of `StringBuilder`.

arteam (commented):
Generally looks good, I've left some comments.

arteam:
I'm wondering if we could get some ideas from https://github.com/brettwooldridge/influx4j for a zero-garbage reporter implementation.

metrics-influxdb/src/main/java/io/dropwizard/metrics5/influxdb/InfluxDbUdpSender.java
mbrannstrom: In latency sensitive applications, low memory turnover is important. Including a 3rd party library that creates lots of garbage is often not acceptable, since it affects the GC pauses of the entire application.

It would be nice to be able to use (parts of) the metrics library in such an applications. Especially, since collecting metrics is (also) important in latency and performance critical applications.

See for example, the [garbage free implementation of log4j2](https://logging.apache.org/log4j/2.x/manual/garbagefree.html). Their [`TextEncoderHelper`](https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/TextEncoderHelper.java) has more code... as a comparison. I tried to keep this method simple.

mbrannstrom:
Apart from the comment on `InfluxDbUdpSender#encode`, I can fix the code according to your comments. They all sound reasonable.

Regarding https://github.com/brettwooldridge/influx4j, I don't see how it is GC free nor proper. Maybe I've misunderstood something. For example, see [line 324 in `Point.java` ](https://github.com/brettwooldridge/influx4j/blob/master/src/main/java/com/zaxxer/influx4j/Point.java#L324) for serialization. `String#getBytes()` at line 326 creates a new array and `String#getBytes(int, int, byte[], int)` at line 341 is deprecated (does not properly encode chars). Likely the author has only tested one code path, that happens to be garbage free (but uses deprecated methods).

mbrannstrom:
@arteam I hope you're satisfied with the changes :-)

arteam:
Thanks! I will take another look.

arteam:
Looks good! Thank you very much for the contribution!

arteam:
I've released Metrics 5.0-rc1: http://central.maven.org/maven2/io/dropwizard/metrics5/metrics-core/5.0.0-rc1/

PR #1270, opened 2018-02-06
Refactoring of MetricName for Metrics 5.0

mbrannstrom:
This pull request is a collection of various bug-fixes and improvements to MetricName. Either accept as is, or use it as a basis for discussion.

NOTE: non-backwards compatible changes, but since MetricName has not yet been released this is not an issue (if this pull request is accepted before the release of 5.0.0).

**Major changes** (each in separate commits):

- Bug-fix: Defensive copy of tags in constructor.
- Bug-fix: `tagged` should replace any previous tags (compare with implementation of join).
- Harmonization of `resolve`, `tagged`, `join` and `build`. 
  - `resolve` changed to vararg, to also replace `build(String...)`
  - `static join` (vararg) changed to `append(MetricName)`. All usages of `join` (vararg) were 2 args, and were replaced with e.g. `prefix.append(name)`.
  - `static empty()` is now exposed.
- Changed comparator to follow conventions, i.e. shorter string/sequence is smaller than longer string/sequence.
- Keep tags sorted internally, and optionally expose that in the API as well.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics5/MetricName.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics5/MetricName.java
@@ -14,7 +14,7 @@ import java.util.TreeSet;
 public class MetricName implements Comparable<MetricName> {
 
     private static final String SEPARATOR = ".";
-    private static final Map<String, String> EMPTY_TAGS = Collections.unmodifiableMap(new HashMap<String, String>());
+    private static final Map<String, String> EMPTY_TAGS = Collections.emptyMap();
     static final MetricName EMPTY = new MetricName("");
 
     private final String key;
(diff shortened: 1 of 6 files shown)


arteam:
Thank you for taking a look at the tags feature and improving it! Backward compatibility is not an issue for the features introduced in the 5.0, because we haven't had a stable release yet. I will review the PR shortly.

arteam:
LGTM, thanks for the contribution.

mbrannstrom:
@arteam, version 5.0.0 of dropwizard metrics is already on Maven central. Is that a mistake?
For example, `io.dropwizard.metrics5:metrics-core:5.0.0` is on maven central (01-Feb-2018)

BTW, when is metrics 5.0.0 planned to be released?

arteam:
Yes, it has been released by mistake. I mentioned it here: https://github.com/dropwizard/metrics/releases/tag/v5.0.0-rc0

I don't have a specific ETA for 5.0.0, sorry. I think there's still some work for providing better tag support for the Prometheus reporter. After it's merged, I think I can make a release.

Issue #1273, opened 2018-02-20
Expansion filter for Slf4jReporter


fabienrenaud:
The slf4j reporter is too verbose by default and I'd like to be able to filter which expansions (as defined by [coursera-metrics-datadog](https://github.com/coursera/metrics-datadog)) are logged.

Proposed api:

```java
var expansions = EnumSet.of(COUNT, RATE_1_MINUTE, RATE_15_MINUTE, MEDIAN, P95, P99);
Slf4jReporter.forRegistry(registry)
                .outputTo(LoggerFactory.getLogger(Foo.class))
                .withExpansions(expansions)
                .convertRatesTo(TimeUnit.SECONDS)
                .convertDurationsTo(TimeUnit.MILLISECONDS)
                .build()
```

If there is no objection to this proposition, I will submit a pull request within the next few days.

arteam:
Hi. The proposal looks good and a PR will be welcomed. We have implemented this ability for [GraphiteRepoter](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java#L177) and [ConsoleReporter](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/ConsoleReporter.java#L171), would be nice to have for `Slf4jReporter` too. The enum is called `MetricAttribute` and we stick to the name convention `disabledMetricAttributes`.

fabienrenaud:
I submitted a PR. Let me know of any feedback.

arteam:
Thanks, I will take a look shortly.

arteam:
Closed via #1275

fabienrenaud:
You introduced a defect with https://github.com/dropwizard/metrics/commit/4384903a1f170a6be83e571afa6a203ad9a83218

The StringBuilder is reused across calls to the log* methods, so `setLength(0)` is required. This helps reducing the frequency of buffer resize/copy and generates less garbage (which is good since we want logging to be as cheap as possible).
There is no test in `Slf4jReporterTest` verifying logging multiple metrics in a row outputs the expected result.

arteam:
Ah, sorry for that. I had presumed we build one message they and log it. I've brought resetting of the builder back and added a unit test for logging several metrics in 010c2dfd8d09911dbfa0e5764577148bdb9b16da. Thank you for spotting the regression!

PR #1275, opened 2018-02-21
4.1 dev slf4j metric attributes

fabienrenaud:
https://github.com/dropwizard/metrics/issues/1273

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
@@ -12,7 +12,21 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
-import static com.codahale.metrics.MetricAttribute.*;
+import static com.codahale.metrics.MetricAttribute.COUNT;
+import static com.codahale.metrics.MetricAttribute.M15_RATE;
+import static com.codahale.metrics.MetricAttribute.M1_RATE;
+import static com.codahale.metrics.MetricAttribute.M5_RATE;
+import static com.codahale.metrics.MetricAttribute.MAX;
+import static com.codahale.metrics.MetricAttribute.MEAN;
+import static com.codahale.metrics.MetricAttribute.MEAN_RATE;
+import static com.codahale.metrics.MetricAttribute.MIN;
+import static com.codahale.metrics.MetricAttribute.P50;
+import static com.codahale.metrics.MetricAttribute.P75;
+import static com.codahale.metrics.MetricAttribute.P95;
+import static com.codahale.metrics.MetricAttribute.P98;
+import static com.codahale.metrics.MetricAttribute.P99;
+import static com.codahale.metrics.MetricAttribute.P999;
+import static com.codahale.metrics.MetricAttribute.STDDEV;
 
 /**
  * A reporter class for logging metrics values to a SLF4J {@link Logger} periodically, similar to
@@ -172,7 +186,7 @@ public class Slf4jReporter extends ScheduledReporter {
          * Don't report the passed metric attributes for all metrics (e.g. "p999", "stddev" or "m15").
          * See {@link MetricAttribute}.
          *
-         * @param disabledMetricAttributes a {@link MetricFilter}
+         * @param disabledMetricAttributes a set of {@link MetricAttribute}
          * @return {@code this}
          */
         public Builder disabledMetricAttributes(Set<MetricAttribute> disabledMetricAttributes) {
@@ -212,7 +226,6 @@ public class Slf4jReporter extends ScheduledReporter {
     private final LoggerProxy loggerProxy;
     private final Marker marker;
     private final String prefix;
-    private final ThreadLocal<StringBuilder> threadLocalStringBuilder = ThreadLocal.withInitial(StringBuilder::new);
 
     private Slf4jReporter(MetricRegistry registry,
                           LoggerProxy loggerProxy,
@@ -238,15 +251,13 @@ public class Slf4jReporter extends ScheduledReporter {
                        SortedMap<String, Meter> meters,
                        SortedMap<String, Timer> timers) {
         if (loggerProxy.isEnabled(marker)) {
-            StringBuilder b = threadLocalStringBuilder.get();
+            StringBuilder b = new StringBuilder();
             for (Entry<String, Gauge> entry : gauges.entrySet()) {
                 logGauge(b, entry.getKey(), entry.getValue());
             }
 
-            if (!getDisabledMetricAttributes().contains(COUNT)) {
-                for (Entry<String, Counter> entry : counters.entrySet()) {
-                    logCounter(b, entry.getKey(), entry.getValue());
-                }
+            for (Entry<String, Counter> entry : counters.entrySet()) {
+                logCounter(b, entry.getKey(), entry.getValue());
             }
 
             for (Entry<String, Histogram> entry : histograms.entrySet()) {
@@ -342,15 +353,15 @@ public class Slf4jReporter extends ScheduledReporter {
         }
     }
 
-    private void appendLongIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Long> durationSupplier) {
+    private void appendLongIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Long> valueSupplier) {
         if (!getDisabledMetricAttributes().contains(metricAttribute)) {
-            append(b, metricAttribute.getCode(), durationSupplier.get());
+            append(b, metricAttribute.getCode(), valueSupplier.get());
         }
     }
 
-    private void appendDoubleIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Double> durationSupplier) {
+    private void appendDoubleIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Double> valueSupplier) {
         if (!getDisabledMetricAttributes().contains(metricAttribute)) {
-            append(b, metricAttribute.getCode(), durationSupplier.get());
+            append(b, metricAttribute.getCode(), valueSupplier.get());
         }
     }
 
--- a/metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
@@ -4,10 +4,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.Marker;
 
+import java.util.Collections;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import static com.codahale.metrics.MetricAttribute.*;
 
 /**
  * A reporter class for logging metrics values to a SLF4J {@link Logger} periodically, similar to
@@ -44,6 +49,7 @@ public class Slf4jReporter extends ScheduledReporter {
         private MetricFilter filter;
         private ScheduledExecutorService executor;
         private boolean shutdownExecutorOnStop;
+        private Set<MetricAttribute> disabledMetricAttributes;
 
         private Builder(MetricRegistry registry) {
             this.registry = registry;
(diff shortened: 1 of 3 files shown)


metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -186,13 +205,14 @@ public Slf4jReporter build() {
...
             }
-            return new Slf4jReporter(registry, loggerProxy, marker, prefix, rateUnit, durationUnit, filter, executor, shutdownExecutorOnStop);
+            return new Slf4jReporter(registry, loggerProxy, marker, prefix, rateUnit, durationUnit, filter, executor, shutdownExecutorOnStop, disabledMetricAttributes);
         }
     }
 
     private final LoggerProxy loggerProxy;
     private final Marker marker;
     private final String prefix;
+    private final ThreadLocal<StringBuilder> threadLocalStringBuilder = ThreadLocal.withInitial(StringBuilder::new);
```
arteam: I feel that using a `ThreadLocal` is a premature optimization. It's not clear for me that it will provide a big performance boost over allocating a new `StringBuilder`. Also,  using `ThreadLocal` can cause strange classloader leaks if one fail to clear a reference to it.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -217,94 +238,155 @@ public void report(SortedMap<String, Gauge> gauges,
...
         if (loggerProxy.isEnabled(marker)) {
+            StringBuilder b = threadLocalStringBuilder.get();
             for (Entry<String, Gauge> entry : gauges.entrySet()) {
-                logGauge(entry.getKey(), entry.getValue());
+                logGauge(b, entry.getKey(), entry.getValue());
             }
 
-            for (Entry<String, Counter> entry : counters.entrySet()) {
-                logCounter(entry.getKey(), entry.getValue());
+            if (!getDisabledMetricAttributes().contains(COUNT)) {
```
arteam: `count` is not used to filtering counters. It's used only to filter the count attribute of histograms, meters, and timers. Users can filter counters by `MetricFilter`.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -217,94 +238,155 @@ public void report(SortedMap<String, Gauge> gauges,
...
+        appendRateIfEnabled(b, MEAN_RATE,  meter::getMeanRate);
+    }
+
+    private void appendRateIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Double> rateSupplier) {
+        if (!getDisabledMetricAttributes().contains(metricAttribute)) {
+            append(b, metricAttribute.getCode(), convertRate(rateSupplier.get()));
+        }
+    }
+
+    private void append(StringBuilder b, String key, long value) {
```
arteam, line 387: It seems to me that  perfomance benefits of not conveting primtives to wrappers are slim to have 4 `apennd` methods. I think one method accepting an `Object` should be enough, especially considering that in most cases we use `Supplier` which forces auto-boxing.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -4,10 +4,15 @@
...
 
+import java.util.Collections;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import static com.codahale.metrics.MetricAttribute.*;
```
arteam: Please don't use star imports, in the codebase we tend to use direct imports.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -161,6 +168,18 @@ public Builder withLoggingLevel(LoggingLevel loggingLevel) {
             return this;
         }
 
+        /**
+         * Don't report the passed metric attributes for all metrics (e.g. "p999", "stddev" or "m15").
+         * See {@link MetricAttribute}.
+         *
+         * @param disabledMetricAttributes a {@link MetricFilter}
```
arteam: s/MetricFilter/a set of MetricAttribute

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
```diff
@@ -4,66 +4,89 @@
...
+import java.util.EnumSet;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static com.codahale.metrics.MetricAttribute.*;
```
arteam: Please replace the star imports with direct imports.

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
```diff
@@ -4,66 +4,89 @@
...
-    private final Slf4jReporter errorReporter = Slf4jReporter.forRegistry(registry)
-            .outputTo(logger)
-            .markWith(marker)
-            .convertRatesTo(TimeUnit.SECONDS)
-            .convertDurationsTo(TimeUnit.MILLISECONDS)
-            .withLoggingLevel(Slf4jReporter.LoggingLevel.ERROR)
-            .filter(MetricFilter.ALL)
-            .build();
+
+    private Set<MetricAttribute> disabledMetricAttributes = null;
```
arteam: Instead of declaring disabledMetricAttributes as a field, I think we should it make a parameter to the `infoReporter` and `errorReporter` methods. Othwerise it's not clear how what's a relationship between it and the actual reporter.

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
```diff
@@ -313,33 +310,54 @@ public void reportsTimerValues() {
...
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS)
+                .withLoggingLevel(Slf4jReporter.LoggingLevel.ERROR)
+                .filter(MetricFilter.ALL)
+                .disabledMetricAttributes(disabledMetricAttributes)
+                .build();
+    }
+
+    private void verifyError(String expectedLog) {
+        if (expectedLog == null) {
```
arteam: Do we really need conditional checks if we use it only in one place?

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
```diff
@@ -313,33 +310,54 @@ public void reportsTimerValues() {
...
+            verify(logger, never()).error(marker, expectedLog);
+        } else {
+            verify(logger).error(marker, expectedLog);
+        }
+    }
+
+    private void verifyInfo(String expectedLog) {
+        if (expectedLog == null) {
+            verify(logger, never()).info(marker, expectedLog);
```
arteam: It looks like this branch of the execution isn't used at all.

arteam (commented):
Looks good, I've left some comments.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
fabienrenaud: fixed.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
fabienrenaud: fixed. fixed GraphiteReporter as well.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
fabienrenaud: removed ThreadLocal. StringBuilder has to be a local variable to the `report` method since there is no guarantee the method won't be executed in parallel.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
fabienrenaud: removed if.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
fabienrenaud: I think they are worth keeping:
 - they are very simple methods, easy to read, understand, change. no hidden code complexity here.
 - convertDuration and convertRate return `double` so only having an `append(...,Object)` could cause extra auto-boxing
 - the JVM can likely optimize `primitive -> box(primitive) -> primitive` call chains
 - and yes, StringBuilder.append(long) is slightly more efficient than StringBuilder.append(Object) called with a Long (unless the JVM un-erases types for some core classes? -- Hotspot JVM does some surprising optimizations some times :) ). 

I can remove them if you feel strongly about it but now that the code is there, I think it's worth keeping.

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
fabienrenaud: fixed.

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
fabienrenaud: the idea was to be consistent with how tests were written.
removed.

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
fabienrenaud: this is no longer necessary since attribute filtering doesn't apply to counters

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -217,94 +238,155 @@ public void report(SortedMap<String, Gauge> gauges,
...
+        }
+    }
+
+    private void appendDoubleDurationIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Double> durationSupplier) {
+        if (!getDisabledMetricAttributes().contains(metricAttribute)) {
+            append(b, metricAttribute.getCode(), convertDuration(durationSupplier.get()));
+        }
+    }
+
+    private void appendLongIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Long> durationSupplier) {
```
fabienrenaud: renamed durationSupplier here since this isn't a duration.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
```diff
@@ -217,94 +238,155 @@ public void report(SortedMap<String, Gauge> gauges,
...
+        if (!getDisabledMetricAttributes().contains(metricAttribute)) {
+            append(b, metricAttribute.getCode(), durationSupplier.get());
+        }
+    }
+
+    private void appendDoubleIfEnabled(StringBuilder b, MetricAttribute metricAttribute, Supplier<Double> durationSupplier) {
```
fabienrenaud: ditto

metrics-core/src/test/java/com/codahale/metrics/Slf4jReporterTest.java
fabienrenaud: I like to use this pattern to override default values passed to the constructor of the main unit tested component. I think in the context of unit tests it makes sense since there is only 1 class that is supposed to be tested. In the context of Slf4jReporter, it should be clear that the only way of testing Slf4jReporter is to use its builder and that its builder takes an optional `Set<MetricAttribute> disabledMetricAttributes`.
This pattern becomes particularly handy when there might be more than 1 default value to overwrite for some tests. The alternatives would be to either invoke the full builder with all the proper default values set but this leads to code duplication; or to add more arguments to the `infoReporter` and `errorReporter` factory methods to overwrite the default value but this doesn't scale well when more than a couple of variables needs their default values overwritten and the factory method may end up having as many parameters as the builder...
I'll add a comment clarifying this is used to build the reporter in each test and can be overridden. I will also move at the top of the infoReporter and errorReporter methods to hep identify the relationship.

metrics-core/src/main/java/com/codahale/metrics/Slf4jReporter.java
arteam: Fair enough

arteam:

fabienrenaud:
@arteam Could you please release this change? It's been 8 months...

arteam:
Sure, will do a 4.1.0 release this week.

PR #1282, opened 2018-03-10
Collectd tests failed with java.net.SocketException: Can't assign requested address

chids:
When running tests locally collectd tests failed with `java.net.SocketException: Can't assign requested address`. [This StackOverflow](https://stackoverflow.com/questions/18747134/getting-cant-assign-requested-address-java-net-socketexception-using-ehcache) seemed plausble as I'm also on a Mac with wireless so I added `-Djava.net.preferIPv4Stack=true` to the Surefire configuration which fixed the problem On My Machine™ 🤷‍♂️.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,6 +252,7 @@
                 <version>2.14.1</version>
                 <configuration>
                     <parallel>classes</parallel>
+                    <argLine>-Djava.net.preferIPv4Stack=true</argLine>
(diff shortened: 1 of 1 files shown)


arteam:
LGTM. I'm all in favor of the build portability between different OS. Alas, we don't have a CI server for MacOS machines (only for Linux and Windows).

Issue #846, opened 2015-08-03
Create Interface for HealthCheck Operations Providers


jwcarman:
We are using an OSGi environment and we have created a "wrapper" interface for HealthCheck objects, so that they can be exposed as OSGi services and picked up using the "whiteboard" pattern.  It would be great if there were a native interface bundled with this library that folks could implement in order to provide health check results.

jillesvangurp:
We use Java 8 and it would be nice if HealthCheck was a functional interface instead of an abstract class. That way we could use lambda functions to quickly create health checks without a lot of boiler plate for the inner class. 

The execute method in the class could move elsewhere probably. Unfortunately it is public and non-final, which suggest that somebody might be overriding this. Or if you want to keep the abstract class, it could be renamed to `AbstractHealthCheck` and implement a new `HealthCheck` interface. This would be a backwards incompatible change. Alternatively, the interface could have a different name. 

Another interface that would be nice is something similar to the `MetricSet` interface to indicate that a class provides a `HealthCheck`. I actually created one for our spring setup along with a spring context listener that registers all health checks.

```java
public interface HealthCheckProvider {
    HealthCheck healthCheck();
}

public class HealthCheckContextListener implements ApplicationListener<ContextRefreshedEvent> {
    private static final Logger LOG = LoggerFactory.getLogger(HealthCheckContextListener.class);

    private final HealthCheckRegistry healthCheckRegistry;

    public HealthCheckContextListener(HealthCheckRegistry healthCheckRegistry) {
        this.healthCheckRegistry = healthCheckRegistry;
    }

    @Override
    public void onApplicationEvent(ContextRefreshedEvent event) {
        // register health checks provided by other beans
        Map<String, HealthCheckProvider> healthCheckProvidingBeans = event.getApplicationContext().getBeansOfType(HealthCheckProvider.class);
        for(Entry<String, HealthCheckProvider> e: healthCheckProvidingBeans.entrySet()) {
            LOG.info("health check " + e.getKey() + " " + e.getValue().getClass().getName());
            HealthCheck check = e.getValue().healthCheck();
            healthCheckRegistry.register(e.getKey(), check);
        }
        // register any health check beans registered directly
        Map<String, HealthCheck> healthCheckBeans = event.getApplicationContext().getBeansOfType(HealthCheck.class);
        for(Entry<String, HealthCheck> e: healthCheckBeans.entrySet()) {
[...]

PR #1283, opened 2018-03-10
HealthCheck as an interface

### LINKED ISSUE #697: HealthCheck should really be an interface, not an abstract class
`HealthCheck` would be much more useful as an interface, then classes could mix it in to be monitorable.
The `execute` method could be a default method as of Java 8, or it could move to a helper class.

chids:
Been longing for `HealthCheck` to be an interface for some time and had som recent code that'd been better of if it were so decided to take a stab at it. If for nothing else to allow easier assessment of the impact of such a change.

(_wasn't able to run all tests locally due to networking weirdnesses, partly addressed in #1282, but hey it complies_ 😬)

Closes #697 
Partly addresses #846

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/AsyncHealthCheckDecorator.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/AsyncHealthCheckDecorator.java
@@ -8,7 +8,7 @@ import java.util.concurrent.ScheduledFuture;
 /**
  * A health check decorator to manage asynchronous executions.
  */
-class AsyncHealthCheckDecorator extends HealthCheck implements Runnable {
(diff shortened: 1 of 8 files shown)


arteam:
Thanks for the PR! The biggest setback for turning `HealthCheck` into an interface from an abstract class was backwards compatibility between Metrics 3.2 and Metrics 4.0. Ideally, users shouldn't rewrite their health checks when upgrading to Metrics 4.0.

Nevertheless, it seems reasonable to implement this change in Metrics 5.0 along with other breaking changes.

chids:
Makes sense. Do you want me to create a PR against the `5.0-development` branch instead?

arteam:
Yes, please. That would be great!

PR #1285, opened 2018-03-13
Enable customization of MetricsServlet.

### LINKED ISSUE #1281: Customizable MetricsServlet
I would like to propose the change of accepting a customized MetricsModule into MetricsServlet.
There are some customization issues on MetricsModule itself,  since members such as rateUnit, durationUnit, showSamples and filter are private, it is not possible to override. My suggestions are:

- Make members of MetricsModule protected so it is easier to add in new serializers in the setupModule function. This is particularly helpful if we want to limit the rollups being provided for each metric type such as Counters, Timers etc. I think this should be owned by the metric type itself, however it is fine for now if we can add in customizable serializers since it would be the same effect.
- Do the same for MetricsServlet or let a MetricsModule *like* class be injectable in the constructor. This should make it possible to customize the response from MetricsServlet.

I would be happy to work on a PR if my suggestions look good.

rahulravindran0108:
**Summary**
`MetricsModule` should now be easier to customize with other serializers. `setupMetricsModule` can be overridden to facilitate introduction of new `MetricsModule` like objects.

**Issues**
Closes #1281 
Addresses #1281

### DIFF
--- a/metrics-json/src/main/java/com/codahale/metrics/json/MetricsModule.java
+++ b/metrics-json/src/main/java/com/codahale/metrics/json/MetricsModule.java
@@ -216,10 +216,10 @@ public class MetricsModule extends Module {
         }
     }
 
-    private final TimeUnit rateUnit;
-    private final TimeUnit durationUnit;
-    private final boolean showSamples;
-    private final MetricFilter filter;
+    protected final TimeUnit rateUnit;
+    protected final TimeUnit durationUnit;
+    protected final boolean showSamples;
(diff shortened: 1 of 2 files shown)


rahulravindran0108:
@arteam did you want me to add this into #1240 or incorporate that on this PR ?

arteam:
No, that's not needed. Thank you for the pull request, I will review it shortly!

arteam:
@rahulravindran0108 Thanks for the contribution!

PR #1288, opened 2018-03-16
Add support for signing and verifying requests to CollectD

### LINKED ISSUE #1284: Use collectd-metrics with network cryptography
Hi, we want to use collectd-metrics in our project, but looks like there is no support for network cryptography.
For example, as described in collectd wiki: https://collectd.org/wiki/index.php/Networking_introduction
In order to send metrics to our collectd server, we need to provide credentials and `SecurityLevel "Encrypt"`. Do you have plans on this?
Thanks!

arteam:
Starting with version 4.7 the CollectD provides the possibility to cryptographically sign or encrypt the network traffic. The type of protection is configured by the securityLevel option:

* None (plain traffic)
* Sign (packets should provide a digital signature)
* Encrypt (packets should be encrypted)

This changes adds possibility to set up a username, password, and a securityLevel in the CollectD reporter, so it can send metrics data  in the format expected by the CollectD server.

Resolves #1284


arteam:
Verified signing and encryption locally with https://github.com/collectd/go-collectd.

PR #1289, opened 2018-03-16
Healthceck as an interface 5.0

### LINKED ISSUE #697: HealthCheck should really be an interface, not an abstract class

chids:
_(#1283 repurposed for `5.0-development`)_
> 
> Been longing for `HealthCheck` to be an interface for some time and had som recent code that'd been better of if it were so decided to take a stab at it. If for nothing else to allow easier assessment of the impact of such a change.
> 
> (_wasn't able to run all tests locally due to networking weirdnesses, partly addressed in #1282, but hey it complies_ 😬)
> 
> Closes #697 
> Partly addresses #846

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ For more information, please see [the documentation](http://dropwizard.github.io
 
 #### Version 4.x.x ([Javadoc](https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.0.1))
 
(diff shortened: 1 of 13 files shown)


chids:
The Travis build failed  on `io.dropwizard.metrics5.jvm.CpuTimeClockTest.cpuTimeClock(CpuTimeClockTest.java:17)` for both 4e1e164 and c747ccd due to...
```
Expecting:
  <1.521299482153E12>
to be close to:
  <1.521299482361E12>
by less than <200.0> but difference was <208.0>.
```
...so I winged it 👉 57ed131.

arteam:
LGTM, thank you for the contribution!

joschi:
Closes #846

PR #1291, opened 2018-03-30
Add counter for active connections

jhaber:
Way back in the day, the Jetty 8 instrumentation included a counter for active connections:
https://github.com/dropwizard/metrics/blob/5eb6730346cff7414229ab77a7edc074564d235c/metrics-jetty/src/main/java/com/yammer/metrics/jetty/InstrumentedSelectChannelConnector.java#L47-L49

We've had mixed results trying to add this instrumentation ourselves to Jetty 9, but I was just curious if there's a reason that it was removed? Example PR included for how it might look in `InstrumentedConnectionFactory`

### DIFF
--- a/metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
+++ b/metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
@@ -15,22 +15,16 @@ import io.dropwizard.metrics5.Timer;
 public class InstrumentedConnectionFactory extends ContainerLifeCycle implements ConnectionFactory {
     private final ConnectionFactory connectionFactory;
     private final Timer timer;
-    private final Optional<Counter> counterMaybe;
+    private final Counter counter;
 
     public InstrumentedConnectionFactory(ConnectionFactory connectionFactory, Timer timer) {
-        this(connectionFactory, timer, Optional.empty());
+        this(connectionFactory, timer, null);
     }
 
     public InstrumentedConnectionFactory(ConnectionFactory connectionFactory, Timer timer, Counter counter) {
-        this(connectionFactory, timer, Optional.of(counter));
-    }
-
-    private InstrumentedConnectionFactory(ConnectionFactory connectionFactory,
-                                          Timer timer,
-                                          Optional<Counter> counterMaybe) {
         this.connectionFactory = connectionFactory;
         this.timer = timer;
-        this.counterMaybe = counterMaybe;
+        this.counter = counter;
         addBean(connectionFactory);
     }
 
@@ -53,13 +47,17 @@ public class InstrumentedConnectionFactory extends ContainerLifeCycle implements
             @Override
             public void onOpened(Connection connection) {
                 this.context = timer.time();
-                counterMaybe.ifPresent(Counter::inc);
+                if (counter != null) {
+                    counter.inc();
+                }
             }
 
             @Override
             public void onClosed(Connection connection) {
                 context.stop();
-                counterMaybe.ifPresent(Counter::dec);
+                if (counter != null) {
+                    counter.dec();
+                }
             }
         });
         return connection;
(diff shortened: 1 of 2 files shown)


arteam:
I think it's gone because `SocketConnector` was removed from Jetty 9. I will take a look at the metric and the others from the Jetty8 connector. The amount of active connections can be an interesting metrics in conjunction with `io.dropwizard.jetty.MutableServletContextHandler.active-requests` to detect idle clients.

metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
```diff
@@ -1,21 +1,36 @@
...
 import org.eclipse.jetty.util.component.ContainerLifeCycle;
 
-import java.util.List;
+import io.dropwizard.metrics5.Counter;
+import io.dropwizard.metrics5.Timer;
 
 public class InstrumentedConnectionFactory extends ContainerLifeCycle implements ConnectionFactory {
     private final ConnectionFactory connectionFactory;
     private final Timer timer;
+    private final Optional<Counter> counterMaybe;
```
arteam: I think a nullable `Counter` would be a better option. We don't do any transformation with the counter and don't expose it, so the advantages of `Optional` are not clear. There's another point: I believe the JVM can optimize out null checks if a final field is null, but I'm not so sure about optionals.

metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
```diff
@@ -1,21 +1,36 @@
...
 
     public InstrumentedConnectionFactory(ConnectionFactory connectionFactory, Timer timer) {
+        this(connectionFactory, timer, Optional.empty());
+    }
+
+    public InstrumentedConnectionFactory(ConnectionFactory connectionFactory, Timer timer, Counter counter) {
+        this(connectionFactory, timer, Optional.of(counter));
+    }
+
+    private InstrumentedConnectionFactory(ConnectionFactory connectionFactory,
```
arteam: If we removed `Optional`, I believe the private constructor would be unnecessary.

metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
```diff
@@ -38,11 +53,13 @@ public Connection newConnection(Connector connector, EndPoint endPoint) {
             @Override
             public void onOpened(Connection connection) {
                 this.context = timer.time();
+                counterMaybe.ifPresent(Counter::inc);
```
arteam: Could you add an integration test to `InstrumentedConnectionFactoryTest` to verify that the counter gets incremented when a connection opens, and get decremented when the client close it.

arteam (commented):
Generally LGTM, I've left a couple of comments. Could you also rebase your changes onto the `4.1-development` branch? I think this improvement would be a good fit in 4.1.0.

metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
jhaber: Sure thing

metrics-jetty9/src/main/java/io/dropwizard/metrics5/jetty9/InstrumentedConnectionFactory.java
jhaber: `counterMaybe.ifPresent(Counter::inc);` is quite satisfying compared to null-checks, but I'll update

jhaber:
Should I send a separate PR to `4.1-development` branch? Also, not sure if it's too late for 5.0 but it could be nice to have `InstrumentedConnectionFactory` accept a `MetricRegistry` that way any future refactorings don't keep changing the constructor signature

PR #1294, opened 2018-04-09
Added InstrumentedHttpClientConnectionManager constructor with HttpClientConnectionOperator

jkytomaki:
I'd like to be able to define a custom HttpClientConnectionOperator when creating a InstrumentedHttpClientConnectionManager (to handle some SSL SNI errors). This is supported by a constructor of the parent class PoolingHttpClientConnectionManager, but InstrumentedHttpClientConnectionManager does not offer a matching constructor. I added a new constructor matching the parent class.

Could something like this be added to the project?


arteam:
LGTM. Thanks fotr your contribution!

PR #1307, opened 2018-04-27
Make the relationship between the Reporter and Closeable interfaces explicit.

### LINKED ISSUE #1305: All Reporters are Closeable?
It seems #743 was closed without a merge because the forked repository was deleted? I cloned this repo to verify that all classes which implement `Reporter` (directly or indirectly) also implement `Closeable`. I'd be happy to submit a PR like #743 and provide attribution to @obfischer if it's an acceptable improvement.

tjcelaya:
Resolves #1305 and reimplements #743. Credit to @obfischer for the initial implementation.

This makes it easier to denote that any `Reporter` is automatically `Closeable` when passing around the generic `Reporter` interface. Making the relationship explicit eases situations where a user may receive either a `JmxReporter` or some subclass of `ScheduledReporter`.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -3,7 +3,6 @@ package com.codahale.metrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.Closeable;
 import java.util.Collections;
 import java.util.Locale;
 import java.util.Set;
@@ -23,7 +22,7 @@ import java.util.concurrent.atomic.AtomicInteger;
  * @see CsvReporter
  * @see Slf4jReporter
  */
-public abstract class ScheduledReporter implements Closeable, Reporter {
+public abstract class ScheduledReporter implements Reporter {
 
     private static final Logger LOG = LoggerFactory.getLogger(ScheduledReporter.class);
 
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -18,6 +18,8 @@ import java.util.concurrent.atomic.AtomicInteger;
  * The abstract base class for all scheduled reporters (i.e., reporters which process a registry's
  * metrics periodically).
  *
+ * Since {@link Reporter} extends {@link java.io.Closeable} this class also satisfies that interface.
+ *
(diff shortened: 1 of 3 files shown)


tjcelaya:
Pushed a fix for the codeclimate issue. There seems to be an unrelated test feailure in metrics-httpasyncclient though, not sure what to do about that.

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -19,11 +18,13 @@
...
  * metrics periodically).
  *
+ * Since {@link Reporter} extends {@link java.io.Closeable} this class also satisfies that interface.
+ *
  * @see ConsoleReporter
  * @see CsvReporter
  * @see Slf4jReporter
  */
-public abstract class ScheduledReporter implements Closeable, Reporter {
+public abstract class ScheduledReporter implements Reporter {
```
arteam: Please don't remove `Closeable` from `ScheduledReporter`, because the change is binary incompatible. That means we will force all 3rd reporters to upgrade to the latest version of `metrics-core`. I think it should be fine to just add `Closeable` to `Reporter`. It should be backward-compatible because `ScheduledReporter` and `JmxReporter` implement it already.

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -30,8 +29,9 @@
 
 /**
  * A reporter which listens for new metrics and exposes them as namespaced MBeans.
+ * Since {@link Reporter} extends {@link java.io.Closeable} this class also satisfies that interface.
  */
-public class JmxReporter implements Reporter, Closeable {
+public class JmxReporter implements Reporter {
```
arteam: The same comment as for `ScheduledReporter`

arteam (commented):
Thanks for the pull request! I've left a suggestion to take a more defensive approach to evolving `Reporter`.

arteam (commented):
Thanks for the pull request! I've left a suggestion to take a more defensive approach to evolving `Reporter`.

tjcelaya:
I hadn't considered that removing the interface would cause problems but it makes sense. Updatd the PR to address the issue.

Also, the same metrics-httpasyncclient error seems to have occurred again.

arteam:
Thanks, I will probably ignore `httpasyncclient` tests becase they fail constantly on Travis CI.

arteam:
Thanks for the contribution!

PR #1309, opened 2018-05-02
Reporters: report Timer sums in the Timer duration units

michaelpj:
Reporting sums for `Timer`s is very useful, however they are currently always reported in nanoseconds, instead of in the `Timer`'s duration units. This is because we inherit the handling of `getSum` from `Meter`, where it has a slightly different meaning.

There are a couple of unsatisfying bits:
- `JmxTimer` loses some precision in reporting the sum, because we want to be able to report fractional values, but `JmxMeter` assumes that it can be represented as a `long`. We could probably just change `JmxMeter` to report a `double`.
- `GraphiteReporter.reportTimer` delegates to a `reportMetered`, which will do the wrong thing with the sum.

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics5/ConsoleReporter.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics5/ConsoleReporter.java
@@ -317,7 +317,7 @@ public class ConsoleReporter extends ScheduledReporter {
     private void printTimer(Timer timer) {
         final Snapshot snapshot = timer.getSnapshot();
         printIfEnabled(MetricAttribute.COUNT, String.format(locale, "             count = %d", timer.getCount()));
-        printIfEnabled(MetricAttribute.SUM, String.format(locale, "               sum = %d", timer.getSum()));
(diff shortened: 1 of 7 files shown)


arteam:
Thanks! This is a very nice PR and I agree that for Timers we should report the sum in a reporter's duration unit. Regarding your questions:

* I think it makes sense to change `getSum` of `JmxMeterMBean` to `double`, because it will allow to report sums more precise in the reporter's duration unit. It's a new field, so we don't have any requirements to keep compatibility.

* In `GraphiteReporter` we could add a new parameter to the `reportMetered` method and check whether we shoud convert it:
```java
 private void reportMetered(MetricName name, Metered meter, long timestamp, boolean isSumADuration) throws IOException {
        sendIfEnabled(COUNT, name, meter.getCount(), timestamp);
        sendIfEnabled(SUM, name, isSumADuration ? convertDuration(meter.getSum()) : meter.getSum(), timestamp);
``` 

I would gladly review both changes as separate PRs.

michaelpj:
Thanks for the quick review! I've put up the suggested follow-up PRs.

PR #1310, opened 2018-05-02
DefaultObjectNameFactory: use type, be a bit more conservative about quoting

michaelpj:
This makes `DefaultObjectNameFactory` a little cleverer. In particular, it uses the `type` argument as the `type` property of the `ObjectName`. This is used by tools like `visualvm` to give another layer of grouping, which is quite helpful.

In addition, we can be a little more conservative about which part of the name we need to quote if we need to quote something.

### DIFF
--- a/metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
+++ b/metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
@@ -1,5 +1,7 @@
 package com.codahale.metrics.jmx;
 
+import java.util.Hashtable;
+
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 
@@ -13,10 +15,28 @@ public class DefaultObjectNameFactory implements ObjectNameFactory {
     @Override
     public ObjectName createName(String type, String domain, String name) {
         try {
-            ObjectName objectName = new ObjectName(domain, "name", name);
-            if (objectName.isPattern()) {
-                objectName = new ObjectName(domain, "name", ObjectName.quote(name));
+            ObjectName objectName;
+            Hashtable<String, String> properties = new Hashtable<>();
+
+            properties.put("name", name);
+            properties.put("type", type);
+            objectName = new ObjectName(domain, properties);
+
+            /*
+             * The only way we can find out if we need to quote the properties is by
+             * checking an ObjectName that we've constructed.
+             */
+            if (objectName.isDomainPattern()) {
+                domain = ObjectName.quote(domain);
             }
+            if (objectName.isPropertyValuePattern("name")) {
+                properties.put("name", ObjectName.quote(name));
+            }
+            if (objectName.isPropertyValuePattern("type")) {
+                properties.put("type", ObjectName.quote(type));
+            }
+            objectName = new ObjectName(domain, properties);
+
             return objectName;
         } catch (MalformedObjectNameException e) {
             try {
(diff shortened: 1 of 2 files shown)


metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
```diff
@@ -13,10 +15,35 @@
...
+                domain = ObjectName.quote(domain);
+                retry = true;
+            }
+            if (objectName.isPropertyValuePattern("name")) {
+                properties.put("name", ObjectName.quote(name));
+                retry = true;
             }
+            if (objectName.isPropertyValuePattern("type")) {
+                properties.put("type", ObjectName.quote(type));
+                retry = true;
```
arteam: I think `objectName = new ObjectName(domain, properties)` is small enough to not warrant of of creating a retry variable to control its execution. I think it would be more clear just do use it in every branch.

metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
```diff
@@ -13,10 +15,35 @@
...
+            /*
+             * The only way we can find out if we need to quote the properties is by
+             * checking an ObjectName that we've constructed.
+             */
+            boolean retry = false;
+            if (objectName.isDomainPattern()) {
+                domain = ObjectName.quote(domain);
+                retry = true;
+            }
+            if (objectName.isPropertyValuePattern("name")) {
```
arteam: I think `else if` would be better here because we don't want to check the other branches as soon as we found that we need to create new object name.

metrics-jmx/src/test/java/com/codahale/metrics/jmx/JmxReporterTest.java
```diff
@@ -1,16 +1,19 @@
...
-import com.codahale.metrics.Histogram;
-import com.codahale.metrics.Meter;
-import com.codahale.metrics.MetricFilter;
-import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.Snapshot;
-import com.codahale.metrics.Timer;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
+import static org.assertj.core.api.Assertions.assertThat;
```
arteam: Could you please not reorder the imports? It will make easier to port this change to the 5.0 branch, which has another namespace.

arteam (commented):
LGTM, I've left some comments. I've checked the change and it works in visualvm on my machine, but to be honest, I haven't used JMX reporting for a while and I'm not sure whether we can break any 3rd party tools with it, but let's hope not.

metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
michaelpj: We still need to see which of the three things we need to quote - it might not be all of them.

metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
michaelpj: Sure.

metrics-jmx/src/test/java/com/codahale/metrics/jmx/JmxReporterTest.java
michaelpj: Argh, sorry, I thought I got rid of that and then my IDE put it back in :)

michaelpj:
Updated.

michaelpj:
FWIW I do find the JMX reporting useful for ad-hoc inspection of a running process, so I think it's worth keeping in working condition, even if it's not the primary interface.

arteam:
Thanks for the contribution!

PR #1311, opened 2018-05-03
JMXReporter: report meter/timer sums as doubles

michaelpj:
As discussed in #1309. `double` is slightly less sensical for `Meter` sums, but it allows better precision for `Timer`s.


arteam:
LGTM

PR #1315, opened 2018-05-11
feat: builder pattern for InstrumentedHttpClientConnectionManager

snakeru:
For my project I needed to specify only the name (and metrics registry, of course) and leave all the rest  at defaults. So here is the more flexible approach.

Also, the code was auto-formatted with IntelliJ.

### DIFF
--- a/metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
+++ b/metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
@@ -38,16 +38,28 @@ public class InstrumentedHttpClientConnectionManager extends PoolingHttpClientCo
     private final MetricRegistry metricsRegistry;
     private final String name;
 
+    /**
+     * @deprecated Use {@link #builder(MetricRegistry)} instead.
+     */
+    @Deprecated
     public InstrumentedHttpClientConnectionManager(MetricRegistry metricRegistry) {
         this(metricRegistry, getDefaultRegistry());
     }
 
+    /**
+     * @deprecated Use {@link #builder(MetricRegistry)} instead.
+     */
+    @Deprecated
     public InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
                                                    Registry<ConnectionSocketFactory> socketFactoryRegistry) {
         this(metricsRegistry, socketFactoryRegistry, -1, TimeUnit.MILLISECONDS);
     }
 
 
+    /**
+     * @deprecated Use {@link #builder(MetricRegistry)} instead.
+     */
+    @Deprecated
     public InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
                                                    Registry<ConnectionSocketFactory> socketFactoryRegistry,
                                                    long connTTL,
@@ -56,6 +68,10 @@ public class InstrumentedHttpClientConnectionManager extends PoolingHttpClientCo
     }
 
 
+    /**
+     * @deprecated Use {@link #builder(MetricRegistry)} instead.
+     */
+    @Deprecated
     public InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
                                                    Registry<ConnectionSocketFactory> socketFactoryRegistry,
                                                    HttpConnectionFactory<HttpRoute, ManagedHttpClientConnection>
@@ -73,6 +89,10 @@ public class InstrumentedHttpClientConnectionManager extends PoolingHttpClientCo
              name);
     }
 
+    /**
+     * @deprecated Use {@link #builder(MetricRegistry)} instead.
+     */
+    @Deprecated
     public InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
                                                    HttpClientConnectionOperator httpClientConnectionOperator,
                                                    HttpConnectionFactory<HttpRoute, ManagedHttpClientConnection>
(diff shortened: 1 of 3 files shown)


metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -30,80 +30,48 @@
 
     protected static Registry<ConnectionSocketFactory> getDefaultRegistry() {
         return RegistryBuilder.<ConnectionSocketFactory>create()
-            .register("http", PlainConnectionSocketFactory.getSocketFactory())
-            .register("https", SSLConnectionSocketFactory.getSocketFactory())
-            .build();
+                .register("http", PlainConnectionSocketFactory.getSocketFactory())
```
arteam: Please don't add unrelated whitespace changes to your PR. This makes porting commits across different versions harder, because it increases the change of a possible merge conflict.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -30,80 +30,48 @@
...
+                .register("https", SSLConnectionSocketFactory.getSocketFactory())
+                .build();
     }
 
     private final MetricRegistry metricsRegistry;
     private final String name;
 
-    public InstrumentedHttpClientConnectionManager(MetricRegistry metricRegistry) {
```
arteam: We try to maintain backward compatibility with the previous version of Metrics. Please don't remove the public constructors, otherwise anybody who uses them will be forced to migrate their code to `Builder`, which should be optional.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -30,80 +30,48 @@
...
-    public InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
+    private InstrumentedHttpClientConnectionManager(MetricRegistry metricsRegistry,
                                                    HttpClientConnectionOperator httpClientConnectionOperator,
                                                    HttpConnectionFactory<HttpRoute, ManagedHttpClientConnection>
                                                            connFactory,
                                                    long connTTL,
                                                    TimeUnit connTTLTimeUnit,
                                                    String name) {
         super(httpClientConnectionOperator, connFactory, connTTL, connTTLTimeUnit);
+        if (metricsRegistry==null) {
```
arteam: This is a good defensive pattern, but across the codebase we don't use it. It's expected that the user will provide a not-null `MetricsRegistry`. Otherwise, one can ask why do we check the metrics registry for nullability, but not `httpClientConnectionOperator` or `connTTLTimeUnit`? I think in the long run a better approach for the project would be to use `@Nullable` annotations from the checker framework for marking nullable fields and parameters, but that's out of the scope of this PR.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -30,80 +30,48 @@
...
+        }
         this.metricsRegistry = metricsRegistry;
         this.name = name;
 
         metricsRegistry.register(name(HttpClientConnectionManager.class, name, "available-connections"),
-            (Gauge<Integer>) () -> {
-                // this acquires a lock on the connection pool; remove if contention sucks
-                return getTotalStats().getAvailable();
-            });
+                (Gauge<Integer>) () -> {
```
arteam: The same comment about whitespacing changes. Also, please take a look at the warnings produced by `checkstyle`.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -114,4 +82,76 @@ public void shutdown() {
...
+        private SchemePortResolver schemePortResolver;
+        private DnsResolver dnsResolver = SystemDefaultDnsResolver.INSTANCE;
+        private long connTTL = -1;
+        private TimeUnit connTTLTimeUnit = TimeUnit.MILLISECONDS;
+        private String name;
+
+        Builder() {
+        }
+
+        public Builder metricsRegistry(final MetricRegistry metricsRegistry) {
```
arteam: To be honest, I don't see much value in declaring the method parameters as `final`. It's clear from the semantic of `Builder` that they are not going to change and we don't force final method parameters in the codestyle. I would remove them unless there's a specific reason to keep.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClients.java
```diff
@@ -28,8 +28,7 @@ public static HttpClientBuilder custom(MetricRegistry metricRegistry,
                                            HttpClientMetricNameStrategy metricNameStrategy) {
         return HttpClientBuilder.create()
                 .setRequestExecutor(new InstrumentedHttpRequestExecutor(metricRegistry, metricNameStrategy))
-                .setConnectionManager(new InstrumentedHttpClientConnectionManager(metricRegistry));
+                .setConnectionManager(InstrumentedHttpClientConnectionManager.builder(metricRegistry).build());
```
arteam, line 31: This change doesn't seem to be very useful, at the end we invoke the same constructor anyway...

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
```diff
@@ -114,4 +82,76 @@ public void shutdown() {
         metricsRegistry.remove(name(HttpClientConnectionManager.class, name, "max-connections"));
         metricsRegistry.remove(name(HttpClientConnectionManager.class, name, "pending-connections"));
     }
+
+    public static Builder builder(final MetricRegistry metricsRegistry) {
+        return new Builder().metricsRegistry(metricsRegistry);
+    }
+
+    public static class Builder {
```
arteam: Could you please write a unit test for `Builder`? Just with specifying a `MetricsRegistry` and a name as you described in the issue.

arteam (changes requested):
Hi @snakeru, thanks for contributing! I like the idea of adding a builder to `InstrumentedHttpClientConnectionManager`, its constructors indeed are rather cumbersome. I've left some comments regarding the implementation, though.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: ok, will undo

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: Arteam, my understanding was that this was just added to the development branch and did not yet made it to any stable release. In this case it is probably ok to make backwards-incompatible changes? Or do you still think that I should add these constructors back?

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: ok, will remove the check.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: The whitespacing was changed by the IntelliJ (which, I think, is a good thing, as long as all users have consistently configured autoformatters). But I'll undo it, no problem.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: There are two notes from my side here:
1) I think these improve readability a bit, though I might be mistaken
2) TBH, this is an autogenerated code. I first did it with lombok and then delombokize the code to get the template out.

Again - since this is your repo, not mine, will remove "final", no problem.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClients.java
snakeru: If we go for the builder pattern, we should discourage usage of the constructor. Ideally - make it private, which consequently makes this change mandatory. So even if we end up not making the constructor private - this would still be a good change as we showcase, how the library is to be used.

Performance-wise this is not a big deal, as setting up metrics usually happens only once per binary execution.

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
snakeru: Sure.

snakeru:
I pushed out some updates, not all of your comments are addressed (please see my comments above).
Standing by for your feedback.

Also, a quick question - once we settle up on some version - do you want me to squash commits or do you prefer to have the complete history with all the small changes?

metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
arteam: Please add them back. Yes, it's a development branch, but we still don't want to break the compatibility across major updates.

arteam:
@snakeru Please add the constructors and it's good to merge. You can organize your commits either way, Github for some time has been providing an option to do a squash and a fast-forward merge automatically.

snakeru:
Ah, @arteam, I thought that this was just added, but now looking at the file history I see that it spans back at least to 2014 and probably more. Adding the constructors back.

Still - should I may be add @Deprecated to them at least?

arteam:
I agree, `@Deprecated` would make sense.

snakeru:
Done

arteam:

PR #1317, opened 2018-05-30
[jvm] make OSGi Package-Import for com.sun.management optional

vorburger:
requested in https://jira.opendaylight.org/browse/INFRAUTILS-41

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,6 +252,7 @@
                             javax.servlet*;version="[2.5.0,4.0.0)",
                             org.slf4j*;version="[1.6.0,2.0.0)",
                             sun.misc.*;resolution:=optional,
+                            com.sun.management.*;resolution:=optional,
                             *
                         ]]>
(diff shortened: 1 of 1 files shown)


arteam:
LGTM. I believe we use the `com.sun.management` package only in `FileDescriptorRatioGauge`, which is part of `metrics-jvm`, so it should be fine to make it optional.

arteam:
@vorburger If you want, I can cut a 4.0.3 release with this change.

vorburger:
@arteam it would be very cool and much appreciated if you could cut a 4.0.3 release - we would test and use it!

splatch:
@arteam, @vorburger any plans to cut 4.0.3 any time soon? ;-) I've just hit this one.

arteam:
Thanks for the reminder, I will cut a release today.

splatch:
@arteam thank you very much for cutting release :hand:!

Issue #1313, opened 2018-05-07
InstrumentedExecutorService: no idle time for callables


mziccard:
https://github.com/dropwizard/metrics/pull/1180 introduced idle time (time spent in the executor queue) for instances of `Runnable` submitted to an `InstrumentedExecutorService` (via `InstrumentedRunnable`). Idle time was not added the same way to `Callable` (i.e. to `InstrumentedCallable`).

What this done on purpose?

If not, I can send a PR, should be an easy fix.

arteam:
Nice catch! I don't know any specific reason why we don't track `idleTime` for `Callable` tasks, I guess it was just an oversight. I would gladly review a PR for adding it.

PR #1326, opened 2018-07-16
added idle time for callable in Instrumented ExecutorService

thiyagu06:
Fix for #1313


thiyagu06:
@arteam can you please look at this? Also which branch you want me to raise PR? The UT failure are not because the change I made.

jplock:
Fixed in 4.1.x by https://github.com/dropwizard/metrics/pull/1407

PR #1337, opened 2018-08-29
Upgrade maven-compiler-plugin to 3.8.0

dpursehouse:
Also define a property for the version, so that future updates
only need to change one place.


dpursehouse:
Hi @arteam any chance to merge this?

PR #1339, opened 2018-09-12
Fix NPE when user trying to pass null value in point tag

SgrAlpha:
Since we allow user to pass Map<String, String> as point tags, we should expect user might pass null as the value. If that happens, the compareTags function will throw an NPE which will eventually block metrics reporting.

The fix is do null check before compare values.


SgrAlpha:
I think your dev branch is broken, I can reproduce the test failure even without my change.

arteam:
Hi @SgrAlpha! Thank you for the PR and thanks for pointing out the broken dev branch. I've fixed the build, could you please rebase your PR?

SgrAlpha:
@arteam I just synced the latest changes, PR build looks good now.

Issue #1125, opened 2017-04-19
Feature request: Record and query the past n-minutes rate of an event.


behrangsa:
My app has an operating that should happen at least once (possibly thousands of times) per minute.

I can't use `com.codahale.metrics.Meter#getOneMinuteRate` as it still takes into account older occurrences, albeit with a lower weight.

This FR is to add a mechanism to record and query the past n-minute rate of an event, ignoring all the events before that.

storozhukBM:
Hi @behrangsa,
I think you can use `com.codahale.metrics.SlidingTimeWindowReservoir` for such purposes.
But in this case should manually create `Timer` and/or `Histogram` with specified reservoir and then register it.

vladimir-bukhtoyarov:
@storozhukBM reporting of rates, does not depend from type of reservoir, timer configured with SlidingTimeWindowReservoir will report absolutely same rates as timer configured with ExponentiallyDecayingReservoir.

vladimir-bukhtoyarov:
@behrangsa  Why do you interest only for rates and do not interest for latency distribution(histogram) of last n-minute events? I ask, because I am creator of [rolling-metrics](https://github.com/vladimir-bukhtoyarov/rolling-metrics) extension for Dropwizzards/Metrics, and request for rates without percentiles looks very strange for me, could you describe your use case s little bit?

behrangsa:
@vladimir-bukhtoyarov I think my requirements got lost in translation. 

What I need is to ensure that a given activity has occurred at least once in the past `n`-minutes, like a heartbeat. 

Anything beyond that (e.g. how many times did the even occur in the past event) is nice to have but not a requirement for my particular case.

Issue #143, opened 2012-01-17
JMX reset operation on metrics


kgs:
It would be really great to have operation "reset" on every metric exposed via JMX.

roytruelove:
+1 on this.  Very sad but I can't use this library without the ability to reset.

wrprice:
+1

amanhigh:
After reading all responses i see only way to reset is using some or other hack. A Simple Counter which resets after reporting data or fixed time interval becomes difficult to realize.

One way is described here.
http://evertrue.github.io/blog/2013/10/14/resettable-counters-with-codahale-metrics/

jmason:
There's another workaround, BTW.

If the desire is to zero Histograms and Timers after each report, this can be accomplished by using the HdrHistogramResetOnSnapshotReservoir from https://bitbucket.org/marshallpierce/hdrhistogram-metrics-reservoir/ ; this is pretty clean.

Alternatively, if you want to reset counters, another way is to subclass the desired Reporter class, use that subclass in your own app, and add an override processCounter() method which uses:

```
        long count = counter.getCount();
        // report the value in the usual way
        counter.dec(count);  // reset the counter for the next iteration
```

(pretty similar to @amanhigh 's suggestion, but without the creation of a Gauge.)

jan-molak:
@jmason - thanks for your suggestion, Justin. We're facing the same difficulty with gauges and counters being reported to the db even though their state hasn't been updated (ie. the client reporting them is down).

I can't see the `processCounter()` you're referring to in the `3.1.0` code base, could you please point me to the right place?
There is for example the `ScheduledReporter::report` method, but then resetting the state of stats at that stage would mess up any other reporters listening on the same `MetricRegistry`...

It's possible to add a `ResettingReporter` that would affect the state of the `MetricRegistry` when it's called, but this would result in similar synchronisation issues as described above. 
It looks like every reporter is running on a separate thread, and with a separate polling schedule - which gives @codahale more flexibility (you can have as many reporters as you want and don't need to care about their schedule), but takes away the control over the process.

I think the reason why Etsy's statsd supports resetting metrics upon a flush is that because [the way they've implemented it](https://github.com/etsy/statsd/blob/master/stats.js#L91) it's the "`MetricRegistry`"that's in control, rather than the reporters. Their equivalent of `MetricRegistry` sends a `flush` event to all its listeners, and then resets the metrics.
[...]

PR #1350, opened 2018-11-19
Add different moving average implementations to Meter

pe-st:
I propose to add a different moving average to meter (in fact the change aims to allow to plug-in any moving average implementation, similar to the Histogram metrics where you can provide different Reservoirs).

My use case is to monitor an online transaction system with hundreds or thousands per seconds and a rate-per-minute metric is exactly what I need.
However I want exact numbers per minute and no influence from past minutes as the exponentially decaying default implementation has; when the number of transactions drops to zero this must be clearly visible.

I'm aware of https://github.com/vladimir-bukhtoyarov/rolling-metrics but would like to keep the Meter interface for easy integration into tools.
The Meter interface is a good model for what I'd expect, it's just the Unix `top` behaviour that doesn't fit my needs.

This feature has already been requested directly ([Feature request: Record and query the past n-minutes rate of an event](https://github.com/dropwizard/metrics/issues/1125)) and indirectly (by requesting resettable metrics; a simple moving average is nothing else than having three counter that are reset automatically after one, five and fifteen minutes):

- [JMX reset operation on metrics](https://github.com/dropwizard/metrics/issues/143)
- [Adds a reset() method to Meter and Timer](https://github.com/dropwizard/metrics/pull/193)
- https://stackoverflow.com/questions/38132551/dropwizard-metrics-how-to-reset-counters-after-reporting-interval
- https://stackoverflow.com/questions/39198572/codahale-metrics-counter-reset-and-count-everyday

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     environment:
       JAVA_TOOL_OPTIONS: -Xms512m -Xmx2g
     docker:
-      - image: circleci/openjdk:8-jdk
+      - image: circleci/openjdk:8u171-jdk
     steps:
       - checkout
       - restore_cache:
--- a/README.md
+++ b/README.md
@@ -11,35 +11,35 @@ For more information, please see [the documentation](http://dropwizard.github.io
 
 ### Versions
 
-#### Version 4.x.x ([Javadoc](https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.0.1))
+#### Version 4.x.x ([Javadoc](https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.0.4))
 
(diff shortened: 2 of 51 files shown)


pe-st:
I tried to reduce the codeclimate issues a bit. There are some issues left, but I'm not sure that it makes sense to address them, I would mark them as false positives (i.e. the legibility of the code could suffer by further refactorings) or legacy (i.e. they were in the code before my pull request).

If you find my PR useful, I can also update the documentation (`docs/source/manual/core.rst`)

wudan3551:
i have same demands, and i think a pluggable Meter is better!

pe-st:
The current travis build with Java 11.0.2 fails due to a problem with javadoc:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:jar (attach-javadocs) on project metrics-annotation: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
[ERROR] 
[ERROR] Command line was: /home/travis/openjdk11/bin/javadoc @options @packages
```

This problem seems to be a problem with JDK or the maven-javadoc-plugin and was already reported elsewhere without known workaround, e.g.

* https://github.com/oracle/opengrok/issues/2629
* https://bugs.openjdk.java.net/browse/JDK-8212233
* https://issues.apache.org/jira/browse/MJAVADOC-555

pe-st:
After merging the fix for Javadoc with Java 11 (Commit 9894180c) the build is green again.

The PR didn't get much attention from the maintainers yet (either positive or negative). Can I improve something? Should I squash the commits to have a shorter history? Or is proposing a PR not the preferred way to contribute?

Kind Regards, Peter

Issue #1347, opened 2018-10-18
Support tagged metrics with JMX 


dlsrb6342:
I'm using metrics-core:5.0.0 and metrics-jmx:5.0.0

From v5, I know that MetricsName is added to support tags.
But with JmxReporter in metrics-jmx, if use the default setting, DefaultObjectNameFactory, all the tags added to MetricName are gone and only domain and name remain. 
It doesn't report tagged metrics.

I think DefaultObjectNameFactory implementation should be changed to support tags.
Is there any plan for this issue?

arteam:
Closed via #1352.

PR #1352, opened 2018-11-27
Edit DefaultObjectNameFactory to support tagged metrics as default

dlsrb6342:
As #1347 I opened, metrics-jmx's DefaulfObjectNameFactory doesn't support tagged metrics.

This fix make jmxReporter can report tagged metrics


arteam:
LGTM, thanks for contributing.

PR #1354, opened 2018-12-04
Avoid boxing some Longs

alexet:
`comparing` requires us to box the long but `comparingLong` doesn't.


arteam (approved):

arteam:
LGTM, thanks for the contribution!

PR #1360, opened 2018-12-22
Adding in a duration field to the healthcheck result

### LINKED ISSUE #1359: Add support for timing metrics in healthchecks
I have had scenarios where my service has a number of healthchecks and unfortunately one of that number acts up and takes a long time to return (i.e. it gets blocked during execution for some reason).  When this happens it causes the entire healthcheck call to wait and respond slower.  During troubleshooting this, it would be helpful to know which of the many healthchecks was the culprit by adding a duration attribute to the healthcheck that is populated with the processing time.

I'm attaching a PR for your review and consideration.

Thank you

chrisrohr:
Adding in a duration field to the healthcheck result that is populated with the time it took to execute the given healthcheck.

This will aide in troubleshooting slow healthchecks.

Fixes #1359

### DIFF
--- a/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
+++ b/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java
@@ -101,7 +101,7 @@ public abstract class HealthCheck {
         private final Throwable error;
         private final Map<String, Object> details;
         private final String timestamp;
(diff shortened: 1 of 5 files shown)


arteam:
This looks like a reasonable improvement, thanks for the contributing! It's a bit unfortunate that `duration` is not final, but I guess it's not very important and can be addressed later if needed.

PR #1366, opened 2018-12-27
Include timestamp in serialized JSON output for HealthCheck.Result (#1365)

### LINKED ISSUE #1365: Consider including timestamp in health check result JSON
`HealthCheck.Result` contains a `timestamp` field, but it is not included in the serialized JSON. There have been times when it would have been very useful to have this information.

I will shortly create a pull request with proposed changes for your consideration.

sleberknight:
- Include timestamp in JSON output from HealthCheckModule
- Update HealthCheckModuleTest
- Update HealthCheckServletTest (since actual timestamp is from a
  ZonedDateTime.now() call in HealthCheck.Result's private constructor,
  the test cannot easily match the exact timestamp. More refactoring
  would need to be done to permit using a mock Clock in order to
  do that. As a result the tests now verify the presence of the
  timestamp but not the actual value)

Closes #1365

### DIFF
--- a/metrics-json/src/main/java/com/codahale/metrics/json/HealthCheckModule.java
+++ b/metrics-json/src/main/java/com/codahale/metrics/json/HealthCheckModule.java
@@ -43,6 +43,7 @@ public class HealthCheckModule extends Module {
                 }
             }
 
+            json.writeStringField("timestamp", result.getTimestamp());
             json.writeEndObject();
         }
 
--- a/metrics-json/src/test/java/com/codahale/metrics/json/HealthCheckModuleTest.java
+++ b/metrics-json/src/test/java/com/codahale/metrics/json/HealthCheckModuleTest.java
@@ -18,26 +18,33 @@ public class HealthCheckModuleTest {
 
     @Test
     public void serializesAHealthyResult() throws Exception {
-        assertThat(mapper.writeValueAsString(HealthCheck.Result.healthy()))
-            .isEqualTo("{\"healthy\":true,\"duration\":0}");
+        HealthCheck.Result result = HealthCheck.Result.healthy();
+        assertThat(mapper.writeValueAsString(result))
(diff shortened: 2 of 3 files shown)


sleberknight:
I looked at the 3 failed code climate issues. The first one which increased the LOC in the test from 25 to 27...I could certainly collapse a few lines in the `isEqualTo` concatenated string but I don't think that makes the code better, and possibly less readable.

"Fixing" the other two issues by extracting the similar couple of lines in those tests also doesn't seem like it would make the code more readable, but it would be easy enough to extract a separate method.

arteam:
No problem, I've marked them as false positives.

arteam:
LGTM! Thanks for the contribution. It would be nice if `Result` had a `Clock`, so it would be possible to pin down the current time, but I guess this could be improved later.

sleberknight:
@arteam I took an initial stab at adding a `Clock` to `Result`. See:

https://github.com/sleberknight/metrics/commit/38f4ae2adce456d2736890b317f5e8ed3cef9825

Without modifying the existing `Result`factory methods like `healthy`, `unhealthy`, etc. I was able to add a `usingClock` method to `HealthCheck.ResultBuilder` which lets you customize the `Clock` _when using a builder_. This let me update the `HealthCheckTest` and `HealthCheckServletTest` to include the actual timestamp in the test assertions.

The `Result` static factory methods will still use the default clock (`Clock.defaultClock()`). I could not immediately think of a non-invasive way (i.e. not significantly changing the public API) to add the `Clock` when using those factory methods.

If you think this change makes sense I can submit directly as a PR or first create an issue and then a PR. Let me know if you think these changes make sense.

sleberknight:
I went ahead and created a PR https://github.com/dropwizard/metrics/pull/1372 for your consideration. It would solve issue https://github.com/dropwizard/metrics/issues/1371 which I also created.

PR #1367, opened 2018-12-27
Specify in BOM only artifacts managed by metrics

### LINKED ISSUE #1336: The Metrics BOM causes issues for projects depending transitively on Logback
Hi,

I have a project depending on the cassandra-all artifact (e.g. https://mvnrepository.com/artifact/org.apache.cassandra/cassandra-all/3.11.3). It has dependencies on logback-classic and logback-core.

When I add the Metrics BOM to my project, the logback-classic version is overridden to 1.2.3, but logback-core is left at 1.1.3.

I think either the BOM should only lock versions for the Metrics artifacts (as e.g. JUnit and Camel do https://mvnrepository.com/artifact/org.junit/junit-bom/5.2.0, https://mvnrepository.com/artifact/org.apache.camel/camel-bom/2.22.0) and not all dependencies, or the BOM should set logback versions for all 3 artifacts (logback-core, logback-classic and logback-access) from the Logback project.

arteam:
Fixes #1336


jplock:
@arteam we should probably fix this in dropwizard too

arteam:
I think it's not a huge problem in Dropwizard because it's usually a standalone app. For example Spring-boot pom also pulls a lot of 3rd-party dependencies: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml

scottcarey:
Oh good!  The former form was unusable.  This is great.

PR #1373, opened 2018-12-30
Made javadoc mention of precision consistent with the implementation.

Sounie:
Also Updated README to refer to latest 4.0.x version, and adjusted some grammar.

- If the implementation should be adjusted to match the existing documentation, then feel free to apply that change instead.


arteam:
That looks much better, thank you!

PR #1404, opened 2019-01-29
Collectd reporter supports disabled metric attributes.

g-fresh:
Added support for disabling metric attributes to the collectd reporter.


g-fresh:
Code Climate complains about a test exceeding a total line/method count limit. 
Do I have to split the tests into multiple files?

The Travis CI build with openjdk11 failed for a reason which is, I believe, not related to my changes.

Kind regards

PR #1409, opened 2019-02-24
Expire @Async healthchecks results after a period of time.

### LINKED ISSUE #1335: Log4j2 MetricsAppender is not detected
We can apply the `log4j-core` module as an annotation processor to `metrics-log4j2` in order to avoid setting a package in the log4j2 configuration.

This creates a discoverable `META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat` resource to allow plugins to be discovered at runtime.

dhatanian:
Fixes #1335. Typical use case is thread starvation preventing the next async healthcheck to run.


dhatanian:
Looking at the build failures:
* Travis build seems to be broken on `4.1-development`. It does not seem related to this PR
* Codeclimate complains about duplication, however I believe I'm following the general style of unit tests in this project. I can refactor and deduplicate more if needed, but it will come at the cost of readability for the unit tests.

Issue #1375, opened 2019-01-03
Better example for Timer


Dabuna:
On https://metrics.dropwizard.io/4.0.0/getting-started.html#timers there is an example for a timer.
The better example is to use the [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) which always closes the timer since it implements Closeable:

```
private final Timer responses = metrics.timer(name(RequestHandler.class, "responses"));

public String handleRequest(Request request, Response response) {
    
    try (final Timer.Context context = responses.time()) {
        doSomeHandling()
    } finally {
        doSomeFinalStuffAfterHandling()   }
}
```

The timer will be stopped by the try-with-resources by calling close()

PR #1410, opened 2019-02-25
Updated the documentation with better implementation of timer

rajcspsg:
Updated the documentation with better implementation of Timer. 
Addresses [issue](https://github.com/dropwizard/metrics/issues/1375).

Please review and kindly let me know if any changes are required

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -264,13 +264,10 @@ duration.
     private final Timer responses = metrics.timer(name(RequestHandler.class, "responses"));
 
     public String handleRequest(Request request, Response response) {
-        final Timer.Context context = responses.time();
-        try {
+        try(final Timer.Context context = responses.time()) {
             // etc;
             return "OK";
-        } finally {
-            context.stop();
-        }
+        } // catch and final logic goes here
     }
 
 This timer will measure the amount of time it takes to process each request in nanoseconds and


docs/source/getting-started.rst
```diff
@@ -264,8 +264,7 @@ duration.
     private final Timer responses = metrics.timer(name(RequestHandler.class, "responses"));
 
     public String handleRequest(Request request, Response response) {
-        final Timer.Context context = responses.time();
-        try {
+        try(final Timer.Context context = responses.time()) {
```
dennyac, line 267: You don't require ```context.stop()``` in the finally block, since you are using try-with-resources

rajcspsg: @dennyac at some point context needs to stopped for timer. 
So I think it should be in either catch block or finally, right?

dennyac: It will be closed automatically. Please read up on [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) and the [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) interface

rajcspsg: Thanks @dennyac. I have updated the PR. Could you please check now?

docs/source/getting-started.rst
```diff
@@ -264,12 +264,11 @@ duration.
...
             // etc;
             return "OK";
         } finally {
-            context.stop();
+            // doSomething();
```
dennyac: nit: Can we remove the finally block altogether?

rajcspsg: @dennyac I have replaced finally with comment. Is that fine?

dennyac: Not super opinionated on it. Will let the maintainers decide (I'm not authorized to merge or make those decisions).

PR #1414, opened 2019-03-29
Make getHealthCheck public

keirlawson:
This is useful for anyone who might have additional annotations on their underlying healthcheck that they want to be able to combine with @Async but still test for in a `HealthCheckRegistryListener`


keirlawson:
Build failure appears unrelated?

arteam (approved):

PR #1436, opened 2019-05-09
Avoid NoSuchElementException updating ExponentiallyDecayingReservoir

### LINKED ISSUE #1278: ExponentiallyDecayingReservoir empty list check?
Hi just wondering why there can't be a NoSuchElementException in ExponentiallyDecayingReservoir line 105. Since it seems possible that the list is still empty - assuming all 1028 threads before it got preempted before inserting a value to the list.

Another issue with this method - Isn't it more efficient to only create the object WeightedSample if it is selected to be stored?

kevinherron:
fixes #1278


arteam:
LGTM, thanks for contributing!

PR #1444, opened 2019-05-31
Add Gauges to measure CPU metrics to metrics-jvm

nchinmay:


joschi:
@dropwizard/dropwizard-metrics Any opinions on adding these metrics by default?

PR #1451, opened 2019-07-04
Expose numeric values (JmxGaugeMBean)

### LINKED ISSUE #1408: JmxReporter's JmxGauge returns Object which does not allow graphing
I have some caches I'd like to report metrics to JMX.  I'm using a gauge which is a great way to grab the metrics from the cache.  However, in Java Mission Control the "Visualize..." menu option is disabled.  This means a gauge can't be visualised.  Obviously not a show stopper, but would be nice if this could be supported.

guillermocalvo:
The purpose of this PR is to allow for easy visualization of gauges in Java Mission Control.

Class `JmxGaugeMBean` was updated so that it will now expose numeric values as `Number`. If the gauge does not represent a number, then the exposed value will be `0`. The actual value can be accessed through `Value` in any case.

- Fixes #1408


guillermocalvo:
Anybody home? CC @dropwizard-committers

arteam:
LGTM, thanks for contributing! I am going to make a 4.0.* release with this change.

PR #1457, opened 2019-07-14
Fixes issue-1455 MetricRegistryListener should be invoked before adding meter to the registry

asarkar:
Each `on*Added` method returns a meter corresponding to the one passed in, making it possible to wrap/proxy a meter before it's added to the registry. A common use case for this is interoperability with another metrics library.

Also got rid of `MetricRegistryListener.Base` since we can use `default` methods in Java 8+, and don't need an `abstract` adapter class.

Added unit tests.

### DIFF
--- a/metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
+++ b/metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
@@ -2,51 +2,46 @@ package com.codahale.metrics;
 
 import org.junit.Test;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verifyZeroInteractions;
 
 public class MetricRegistryListenerTest {
-    private final Counter counter = mock(Counter.class);
-    private final Histogram histogram = mock(Histogram.class);
-    private final Meter meter = mock(Meter.class);
-    private final Timer timer = mock(Timer.class);
-    private final MetricRegistryListener listener = new MetricRegistryListener.Base() {
-
+    private final MetricRegistryListener listener = new MetricRegistryListener() {
     };
 
     @Test
-    public void noOpsOnGaugeAdded() {
-        listener.onGaugeAdded("blah", () -> {
-            throw new RuntimeException("Should not be called");
-        });
+    public void returnsSameGauge() {
+        Gauge<?> gauge = mock(Gauge.class);
+        assertThat(listener.onGaugeAdded("blah", gauge))
+                .isEqualTo(gauge);
     }
 
     @Test
-    public void noOpsOnCounterAdded() {
-        listener.onCounterAdded("blah", counter);
-
-        verifyZeroInteractions(counter);
+    public void returnsSameCounter() {
+        Counter counter = mock(Counter.class);
+        assertThat(listener.onCounterAdded("blah", counter))
+                .isEqualTo(counter);
     }
 
     @Test
-    public void noOpsOnHistogramAdded() {
-        listener.onHistogramAdded("blah", histogram);
-
-        verifyZeroInteractions(histogram);
+    public void returnsSameHistogram() {
+        Histogram histogram = mock(Histogram.class);
+        assertThat(listener.onHistogramAdded("blah", histogram))
+                .isEqualTo(histogram);
     }
 
     @Test
     public void noOpsOnMeterAdded() {
-        listener.onMeterAdded("blah", meter);
-
-        verifyZeroInteractions(meter);
+        Meter meter = mock(Meter.class);
+        assertThat(listener.onMeterAdded("blah", meter))
+                .isEqualTo(meter);
     }
 
     @Test
     public void noOpsOnTimerAdded() {
-        listener.onTimerAdded("blah", timer);
-
-        verifyZeroInteractions(timer);
+        Timer timer = mock(Timer.class);
+        assertThat(listener.onTimerAdded("blah", timer))
+                .isEqualTo(timer);
     }
 
     @Test
--- a/metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
+++ b/metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
@@ -511,7 +511,7 @@ public class JmxReporter implements Reporter, Closeable {
             this.objectNameFactory = objectNameFactory;
         }
 
-        private void registerMBean(Object mBean, ObjectName objectName) throws InstanceAlreadyExistsException, JMException {
+        private void registerMBean(Object mBean, ObjectName objectName) throws JMException {
             ObjectInstance objectInstance = mBeanServer.registerMBean(mBean, objectName);
             if (objectInstance != null) {
                 // the websphere mbeanserver rewrites the objectname to include
@@ -533,7 +533,7 @@ public class JmxReporter implements Reporter, Closeable {
         }
 
         @Override
-        public void onGaugeAdded(String name, Gauge<?> gauge) {
+        public Gauge<?> onGaugeAdded(String name, Gauge<?> gauge) {
             try {
                 if (filter.matches(name, gauge)) {
                     final ObjectName objectName = createName("gauges", name);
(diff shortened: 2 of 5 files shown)


asarkar:
Seems like `codeclimate` check is failing on existing code, and in a way that doesn't make sense.

baharclerode:
How do you handle when a listener is added _after_ some metrics have already been added to the registry? It looks like you ignore the return result form the listener.... which creates a weird behavior in that the return value of the listener may or may not actually be used, and there's no way for the listener to know.

Similar with `register()` and `registerAll()`.... both of these allow an application to pass in a metric that they've already created and are using. The listeners wouldn't be able to wrap these either.

asarkar:
@baharclerode How’s the case of adding a listener after adding a metrics handled in the existing code?

baharclerode:
Currently, when you add a listener, it is invoked for all existing metrics already added to the registry. This ensures that any listener registered with a `MetricRegistry` always knows about every metric in the registry: https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L367-L369

This greatly simplifies the API usage / avoids timing issues about when metrics are registered vs when listeners are registered.

metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
```diff
@@ -2,51 +2,46 @@
...
 
     @Test
-    public void noOpsOnGaugeAdded() {
-        listener.onGaugeAdded("blah", () -> {
-            throw new RuntimeException("Should not be called");
-        });
+    public void returnsSameGauge() {
+        Gauge<?> gauge = mock(Gauge.class);
+        assertThat(listener.onGaugeAdded("blah", gauge))
+                .isEqualTo(gauge);
```
joschi, line 16: Shouldn't it use the `isSameAs()` annotation to match the test name?

metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
```diff
@@ -2,51 +2,46 @@
...
 
     @Test
-    public void noOpsOnCounterAdded() {
-        listener.onCounterAdded("blah", counter);
-
-        verifyZeroInteractions(counter);
+    public void returnsSameCounter() {
+        Counter counter = mock(Counter.class);
+        assertThat(listener.onCounterAdded("blah", counter))
+                .isEqualTo(counter);
```
joschi, line 23: Shouldn't it use the `isSameAs()` annotation to match the test name?

metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
```diff
@@ -2,51 +2,46 @@
...
 
     @Test
-    public void noOpsOnHistogramAdded() {
-        listener.onHistogramAdded("blah", histogram);
-
-        verifyZeroInteractions(histogram);
+    public void returnsSameHistogram() {
+        Histogram histogram = mock(Histogram.class);
+        assertThat(listener.onHistogramAdded("blah", histogram))
+                .isEqualTo(histogram);
```
joschi, line 30: Shouldn't it use the `isSameAs()` annotation to match the test name?

metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
```diff
@@ -2,51 +2,46 @@
...
     }
 
     @Test
     public void noOpsOnMeterAdded() {
```
joschi, line 34: Could we make the test names consistent?

metrics-core/src/test/java/com/codahale/metrics/MetricRegistryListenerTest.java
```diff
@@ -2,51 +2,46 @@
...
-        listener.onMeterAdded("blah", meter);
-
-        verifyZeroInteractions(meter);
+        Meter meter = mock(Meter.class);
+        assertThat(listener.onMeterAdded("blah", meter))
+                .isEqualTo(meter);
     }
 
     @Test
     public void noOpsOnTimerAdded() {
```
joschi, line 41: Could we make the test names consistent?

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -544,6 +544,7 @@ public void onGaugeAdded(String name, Gauge<?> gauge) {
             } catch (JMException e) {
                 LOGGER.warn("Unable to register gauge", e);
             }
+            return null;
```
joschi, line 547: Why don't we return the `Gauge<?>` which has been passed into the method when it was added?

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -570,6 +571,7 @@ public void onCounterAdded(String name, Counter counter) {
             } catch (JMException e) {
                 LOGGER.warn("Unable to register counter", e);
             }
+            return null;
```
joschi, line 574: Why don't we return the `Counter` which has been passed into the method when it was added?

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -596,6 +598,7 @@ public void onHistogramAdded(String name, Histogram histogram) {
             } catch (JMException e) {
                 LOGGER.warn("Unable to register histogram", e);
             }
+            return null;
```
joschi, line 601: Why don't we return the `Histogram` which has been passed into the method when it was added?

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -622,6 +625,7 @@ public void onMeterAdded(String name, Meter meter) {
             } catch (JMException e) {
                 LOGGER.warn("Unable to register meter", e);
             }
+            return null;
```
joschi, line 628: Why don't we return the `Meter` which has been passed into the method when it was added?

metrics-jmx/src/main/java/com/codahale/metrics/jmx/JmxReporter.java
```diff
@@ -648,6 +652,7 @@ public void onTimerAdded(String name, Timer timer) {
             } catch (JMException e) {
                 LOGGER.warn("Unable to register timer", e);
             }
+            return null;
```
joschi, line 655: Why don't we return the `Timer` which has been passed into the method when it was added?

joschi (changes requested):
@asarkar Thanks for your contribution!

In general I like the PR with some minor change requests and questions.

joschi:
@asarkar Please rebase this PR on the current `4.1-development` branch. Thanks!

baharclerode:
How do you control the order in which listeners are invoked? Previously this did not matter since listeners can't modify the metrics when they're added, but now they can so order is important/matters. It is often difficult to control all the points where listeners are added (some may be added by bundles, some may be added by the application code itself, some may be added dymanically at runtime, etc., which makes it difficult to control the order in which listeners are added to a registry.

I'm still curious how you plan to handle adding a listener after metrics are already added to the registry. Right now it means that listeners have two different behaviors, and there's no way to know which will be used-- which makes for a very bad API.

IMHO the proper way to achieve what you're looking for is to create a separate metric registry B, install a metric listener into registry A, and wrap all the metrics before adding them to registry B, and then expose registry B however you want; Or subclass the `MetricRegistry` and override `register(String,T)` to wrap all metrics as they're added. Changing `MetricListener` into what amounts to a `MetricMutator` is a bad design choice.

baharclerode:
I'm actually also curious about this:

> A common use case for this is interoperability with another metrics library.

How does this help interoperability with another metrics library? Wrapping a metric when it is added to the registry does not, for example, allow you to reliably intercept calls or method invocations on that metric, or know when that metric has been updated.

You would actually need a pretty massive overhaul of all of the current metrics to support something like that (each metric would have to have it's own listener type); Even with wrapped metrics, the best you can do right now is poll the metrics periodically and export their aggregate state to another metrics library, which can be done with the existing `MetricListener` API.

joschi:
@dropwizard/metrics @dropwizard/committers How should we continue with this pull request?

PR #1458, opened 2019-07-21
Update pom.xml

phax:
Switched to new profiler version that has no dependency on joda-time - see https://github.com/phax/profiler and http://repo2.maven.org/maven2/com/helger/profiler/1.1.1/ for details and binaries.
Since this project already requires Java 8, getting rid of joda here makes sense to me.

The internal package names and the API are unchanged and it can be considered a drop-in replacement (except for those, who transitively expect joda-time and won't have it anymore).


jplock (approved):

jplock:
Thanks!

PR #1476, opened 2019-08-25
support for Atomic* counters in Graphite Reporter

mazurkin:
support for `AtomicLong`, `AtomicInteger` and any other class that support `Number` interface


arteam:
Nice one!

PR #1483, opened 2019-09-17
Add remove function for metric set,fix #1007 

### LINKED ISSUE #1007: `MetricsRegistry.remove` is not the inverse of `MetricsRegistry.register`
If I register a metric and then remove it, I expect that I should then be able to register it again. (Think of unit tests where you are testing setup and teardown of an app.) However, this does not work without writing additional recursive code on top of this library.

Specifically, `MetricRegistry.register` can [accept a MetricSet](https://github.com/dropwizard/metrics/blob/ed96867/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L84) and recursively add the component metrics, deriving the names as it goes. On the other hand, `.remove` [only removes a single metric](https://github.com/dropwizard/metrics/blob/ed96867/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java#L158).

(This is at the bottom of a pile of yak-shaving for me, so I'm not sure I can offer a PR in good faith.)

mostafakashif:
Adding a new function removeAll inverse to registerAll to remove all the metrics in the metrics set.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -304,12 +304,12 @@ public class MetricRegistry implements MetricSet {
      * Object will be instance of Boolean in case single metric,
      * if metric set it will be returned as instance of Hashmap with status as Boolean for each metric in the set
      */
-    public Map<String,Object> removeAll(MetricSet metrics) {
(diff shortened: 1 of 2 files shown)


joschi:
@mostafakashif Thanks for your contribution!

Could you please describe your use case for these methods?

joschi:
No response from original author for over 6 months. Closing PR.

PR #1490, opened 2019-10-10
Jetty Update to 9.4.21

cstamas:
It contains some notable internal changes (most importantly related to
HttpChannel.State):
https://github.com/eclipse/jetty.project/commit/bde86467f4e5df595773ab11ed5e80c615b741f3#diff-68e74a60841dd5ec1f7edc4bd736ce18

Fix InstrumentedHandler to compile and work with pre-9.4.21 and
post-9.4.21 Jetty instances, but using internal API is not a recipe
for longevity.

Also, created a new InstrumentedHttpChannelListener that in turn
does use public API of Jetty, and should not break with minor
releases.

### DIFF
--- a/metrics-jetty9/pom.xml
+++ b/metrics-jetty9/pom.xml
@@ -18,7 +18,7 @@
 
     <properties>
         <javaModuleName>com.codahale.metrics.jetty9</javaModuleName>
-        <jetty9.version>9.4.20.v20190813</jetty9.version>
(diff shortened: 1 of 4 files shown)


arteam (approved):

arteam:
That's really cool change, thank you very much @cstamas! `InstrumentedHttpChannelListener` could be a good candidate for a 4.2.* release and the dispatcher fix can be backported to the 4.0 and 4.1 branches.

PR #1492, opened 2019-10-13
fix counting suspended requests

nikita-b:
There is a small bug with async requests, metrics dec even if requests no-suspended.


jplock:
@nikita-b mind rebasing?

nikita-b:
@jplock done

nikita-b:
Hm, seems like there is some internal change in last version of Jetty. I will check it.

PR #1496, opened 2019-10-25
Add checking not null metric variable when register metric

apirom9:
Add checking not null metric variable when register metric

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -89,7 +89,7 @@ public class MetricRegistry implements MetricSet {
     public <T extends Metric> T register(String name, T metric) throws IllegalArgumentException {
 
         if (metric == null) {
-            throw new IllegalArgumentException("metric variable must not be null !!!");
+            throw new NullPointerException("metric == null");
         }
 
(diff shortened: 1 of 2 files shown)


metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
```diff
@@ -83,10 +83,15 @@ public MetricRegistry() {
...
      * @param <T>    the type of the metric
      * @return {@code metric}
-     * @throws IllegalArgumentException if the name is already registered
+     * @throws IllegalArgumentException if the name is already registered or metric variable is null
      */
     @SuppressWarnings("unchecked")
     public <T extends Metric> T register(String name, T metric) throws IllegalArgumentException {
+
+        if (metric == null) {
+            throw new IllegalArgumentException("metric variable must not be null !!!");
```
jplock: I'd make this a `throw new NullPointerException("metric == null");` instead

jplock (changes requested):

metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
apirom9: Fixed as request

jplock (approved):

PR #1497, opened 2019-10-25
Prevent creation of ScheduledReporter with null metric registry

apirom9:
Prevent creation of ScheduledReporter with null metric registry

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -130,6 +130,11 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
                                 ScheduledExecutorService executor,
                                 boolean shutdownExecutorOnStop,
                                 Set<MetricAttribute> disabledMetricAttributes) {
+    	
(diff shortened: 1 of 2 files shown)


metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -130,6 +130,11 @@ protected ScheduledReporter(MetricRegistry registry,
                                 ScheduledExecutorService executor,
                                 boolean shutdownExecutorOnStop,
                                 Set<MetricAttribute> disabledMetricAttributes) {
+
+        if (registry == null) {
+            throw new IllegalArgumentException("MetricRegistry must not be null !!!");
```
jplock: I'd make this a `throw new NullPointerException("registry == null");` instead

jplock (changes requested):

metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
apirom9: Fixed as request

jplock (approved):

PR #1504, opened 2019-11-05
Use proper logger in ScheduledReporter

Matzz:
Use logger instead of System.err for logging ScheduledExecutorService termination failure.
I think the current code use System.err because this piece was copied from java documentation:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html
```
The following method shuts down an ExecutorService in two phases, first by calling shutdown to reject incoming tasks, and then calling shutdownNow, if necessary, to cancel any lingering tasks:
 void shutdownAndAwaitTermination(ExecutorService pool) {
   pool.shutdown(); // Disable new tasks from being submitted
   try {
     // Wait a while for existing tasks to terminate
     if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
       pool.shutdownNow(); // Cancel currently executing tasks
       // Wait a while for tasks to respond to being cancelled
       if (!pool.awaitTermination(60, TimeUnit.SECONDS))
           System.err.println("Pool did not terminate");
     }
   } catch (InterruptedException ie) {
     // (Re-)Cancel if current thread also interrupted
     pool.shutdownNow();
     // Preserve interrupt status
     Thread.currentThread().interrupt();
   }
 }
```

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -273,7 +273,6 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
 
     protected double convertDuration(double duration) {
         return duration / durationFactor;
-
     }
 
     protected double convertRate(double rate) {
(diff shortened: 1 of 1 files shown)


metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -272,7 +272,8 @@ protected String getDurationUnit() {
     }
 
     protected double convertDuration(double duration) {
-        return duration / durationFactor;
+        return duration / durationFactor;                LOG.error("Exception thrown from {}#report. Exception was suppressed.", ScheduledReporter.this.getClass().getSimpleName(), ex);
```
joschi: Accidental copy & paste error?

joschi (changes requested):

Matzz:
@joschi Yes, that was c&p issue. Fixed in the last commit

joschi (approved):

Issue #1439, opened 2019-05-14
Add metrics-collectd to metrics-bom


ghaase:
I noticed that pull request #https://github.com/dropwizard/metrics/pull/1239 does not include the metrics-collectd module that was added in pull request  #https://github.com/dropwizard/metrics/pull/871

It would be nice to get this added.

jplock:
@ghaase can you submit a PR?

joschi:
Closed in #1507.

PR #1507, opened 2019-11-11
Add collectd to BOM

ghaase:
In response to issue #1439 

This is PR unnecessarily messy due to alpha ordering of dependencies in the bom and modules in the parent pom. I couldn't figure out where it made the most sense to put the collectd dependency and this also made it easier to check for additional missing modules.

If we don't like it, I can close this PR and resub with the collectd dependency appended to the bottom of the list.


joschi:
@ghaase Thanks for your contribution!

Issue #1513, opened 2019-11-29
Timer passes reservoir only for Histogram and not to Meter, causes discrepancies


rdsubhas:
## Issue

When constructing a `Timer`, it [accepts a Reservoir](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/Timer.java#L71). The docs say:

> reservoir: the {@link Reservoir} implementation the timer should use.

But the reservoir is passed only to the Histogram. Meter is still using default reservoir. So if we use a SlidingTimeWindowReservoir, we get discrepancies with the histogram reporting sliding time window, but Meter reporting the default exponential decay.

## Expectation

There is a way to achieve consistency between the histogram and meter – either both are using exponential decay, or both are using sliding time window.

vasilhsfoto:
Nice feature to have

I created a P.R for that https://github.com/dropwizard/metrics/pull/1517

PR #1517, opened 2019-12-04
Add Timer constructor to inject Metric and Histogram

vasilhsfoto:
This refers to the issue 
https://github.com/dropwizard/metrics/issues/1513


joschi:
@vasilhsfoto Thanks for your contribution!

PR #1544, opened 2020-02-26
Improve tick overflow handling in SlidingTimeWindowReservoir

beargummy:
The current implementation of `SlidingTimeWindowArrayReservoir` handles clock tick overflows much better, than `SlidingTimeWindowReservoir`. This is a change for `SlidingTimeWindowReservoir` to postpone tick overflow as much as possible, as it is done in `SlidingTimeWindowArrayReservoir`.

Maybe it would be even better to clear all measurements in case of overflow as it is done in `SlidingTimeWindowArrayReservoir`, but this may be treated as a backward-incompatible change from an end-user perspective.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowReservoir.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowReservoir.java
@@ -21,6 +21,7 @@ public class SlidingTimeWindowReservoir implements Reservoir {
     private final long window;
     private final AtomicLong lastTick;
     private final AtomicLong count;
+    private final long startTick;
 
     /**
      * Creates a new {@link SlidingTimeWindowReservoir} with the given window of time.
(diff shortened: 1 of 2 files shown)


storozhukBM:
The question is: why would you even want to use `SlidingTimeWindowReservoir` if you can use `SlidingTimeWindowArrayReservoir` since it is a full drop-in replacement.

`SlidingTimeWindowArrayReservoir` is used in production by many systems for years now, so maybe it is time to go one step further get fully replace `SlidingTimeWindowReservoir` with `SlidingTimeWindowArrayReservoir` implementation. What do you think @arteam, @joschi, @ryantenney?

beargummy:
I agree that `SlidingTimeWindowArrayReservoir` is a better choice, but from my perspective, it would be nice to at least backport the tick overflow handling to 3.x branch as it would not hurt anyone.

joschi (approved):
@beargummy Thank you very much!

PR #1558, opened 2020-03-23
Run builds with Java 14 on GitHub Actions and Travis CI

joschi:


pkwarren (approved):

pkwarren:
Looks like Java 6 bytecode is no longer supported by the compiler. We'd need to bump to Java 7 as a baseline to build on Java 14.

joschi:
@pkwarren I remember running into the same problem with Java 13. 😆 

Let's skip this change then. I'm a bit hesitant raising the minimum requirements in a patch version.

Issue #1519, opened 2019-12-16
NoClassDefFoundError on loading com/sun/management/UnixOperatingSystemMXBean


sammyhk:
We got `NoClassDefFoundError` on loading `com/sun/management/UnixOperatingSystemMXBean`  when accessing `FileDescriptorRatioGauge` on OpenJDK8 OpenJ9 build, suspected
https://github.com/dropwizard/metrics/blob/9438c91679eba570e3a7fa1aa43a82801af7a33f/metrics-jvm/src/main/java/com/codahale/metrics/jvm/FileDescriptorRatioGauge.java#L33
using Sun/Oracle internal class which should only works on Oracle/ HotSpot JDK. Suggest to check class accessibility before type casting.

Stacktrace:
```
java.lang.NoClassDefFoundError: com/sun/management/UnixOperatingSystemMXBean
	at com.codahale.metrics.jvm.FileDescriptorRatioGauge.getRatio(FileDescriptorRatioGauge.java:33)
	at com.codahale.metrics.RatioGauge.getValue(RatioGauge.java:64)
	at com.codahale.metrics.RatioGauge.getValue(RatioGauge.java:11)
	at com.codahale.metrics.json.MetricsModule$GaugeSerializer.serialize(MetricsModule.java:42)
	at com.codahale.metrics.json.MetricsModule$GaugeSerializer.serialize(MetricsModule.java:26)
	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeFields(MapSerializer.java:722)
	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:643)
	at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:33)
	at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider._serialize(DefaultSerializerProvider.java:480)
	at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:319)
	at com.fasterxml.jackson.databind.ObjectMapper.writeValue(ObjectMapper.java:2795)
	at com.fasterxml.jackson.core.base.GeneratorBase.writeObject(GeneratorBase.java:388)
	at com.fasterxml.jackson.core.JsonGenerator.writeObjectField(JsonGenerator.java:1769)
	at com.codahale.metrics.json.MetricsModule$MetricRegistrySerializer.serialize(MetricsModule.java:210)
[...]

PR #1562, opened 2020-04-09
#1519 fix NoClassDefFoundError on loading com/sun/management/UnixOper…

sammyhk:
…atingSystemMXBean when classloader restrict access to com.sun package


joschi (approved):
LGTM and there's even a test. ❤️

@sammyhk Thanks for your contribution!

Issue #711, opened 2014-12-01
CachedGauge's getValue() can return null if called concurrently before a value has been cached


cpopp:
The shouldLoad() method of the CachedGauge appears to attempt to be thread safe by checking whether or not the value should be loaded using an AtomicLong for the reload time.

```
private boolean shouldLoad() {
    for (; ; ) {
        final long time = clock.getTick();
        final long current = reloadAt.get();
        if (current > time) {
            return false;
        }
        if (reloadAt.compareAndSet(current, time + timeoutNS)) {
            return true;
        }
    }
}
```

However, if we take a look at the getValue() method...we see that if multiple threads call getValue(), it's possible for one thread to complete the compareAndSet on the reloadAt variable before actually setting the cached value on the field.  As a result, this means that other threads calling getValue() will see the updated reloadAt variable, determined they do not need to load the latest value, and as a result, they return a null value since this.value has not had a value set yet.

```
@Override
public T getValue() {
    if (shouldLoad()) {
        this.value = loadValue();
    }
    return value;
}
```

I encountered this in doing some load testing of an application.  As a temporary fix I just prime the value of the CachedGauge (call getValue()) before making the gauge available.

Here's some code I used to recreate it...when running it you can see that most of the time at least one of the threads will see a null value:

```
public static void main(String[] args) throws Exception {
    final CachedGauge<Integer> gauge = new CachedGauge<Integer>(1, TimeUnit.SECONDS) {
        public Integer loadValue() {
            return 10;
        }
    };

    for(int i = 0; i < 8; i++) {
        new Thread(){
            public void run() {
                System.out.println(gauge.getValue());
            }
        }.start();
    }
}
```

PR #1566, opened 2020-04-16
Make CachedGauge thread-safe

### LINKED ISSUE #1565: CachedGauge is not thread-safe
(I am not sure if `CachedGauge` is meant to be thread-safe. But the `shouldLoad` method indicates the intention of making it thread-safe.)

In https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/CachedGauge.java#L48-L54

```
    @Override
    public T getValue() {
        if (shouldLoad()) {
            this.value = loadValue();
        }
        return value;
    }
```
Imaging `thread-1` and `thread-2` invoking `getValue()` at the same time at very beginning and assume `thread-1` arrives a little bit earlier than thread-2. So according to the `shouldLoad` method, it returns `true` for  `thread-1` and `false` for `thread-2`.

Now `thread-1` is on ` this.value = loadValue();` and `thread-1` is on `return value;`.

The result for `thread-2` will be different depending on which one finishes first.  If `thread-1` finishes `this.value = loadValue();` first, the value returned for `thread-2` is the new value, which is correct. 
But if `return value` finishes for `thread-2` first, then `thread-2` gets the original value (which is `null`) and can cause NPE.  This is what we are seeing in `CachedThreadStatesGaugeSet`. And this is wrong because `thread-2` should get the new value (remember `shouldLoad` returned `false` for `thread-2`)

An example of using `CachedThreadStatesGaugeSet` on multiple threads:
```
Exception in thread "Thread-1" java.lang.NullPointerException
	at com.codahale.metrics.jvm.ThreadStatesGaugeSet.getThreadCount(ThreadStatesGaugeSet.java:95)
	at com.codahale.metrics.jvm.ThreadStatesGaugeSet.access$000(ThreadStatesGaugeSet.java:20)
	at com.codahale.metrics.jvm.ThreadStatesGaugeSet$1.getValue(ThreadStatesGaugeSet.java:56)
```

the-thing:
https://github.com/dropwizard/metrics/issues/1565

I was able to recreate failures with a unit tests and I came up with this non blocking implementation. The issue was already pointed out before - https://github.com/dropwizard/metrics/issues/711

**Changes**

- non blocking, thread safe cached gauge implemented
- added multithreaded unit test
- added cached gauge benchmark

**Benchmark**

Caption			: Intel64 Family 6 Model 60 Stepping 3
MaxClockSpeed	: 4001
Name			: Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
NumberOfCores	: 4 

**Non blocking**

```
Benchmark                           Mode  Cnt          Score         Error  Units
CachedGaugeBenchmark.perfGetValue  thrpt    5  171858323.080 ± 5082330.853  ops/s
```

**Synchronized**

```
Benchmark                           Mode  Cnt         Score        Error  Units
CachedGaugeBenchmark.perfGetValue  thrpt    5  15644559.909 ± 792061.174  ops/s
```

Seems that non blocking implementation is order of magnitude faster (based on timeout == 100ms and loading time == 10ms).

```
171858323 (non blocking)
 15644559 (synchronized)
```

### DIFF
--- a/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/CachedGaugeBenchmark.java
+++ b/metrics-benchmarks/src/main/java/com/codahale/metrics/benchmarks/CachedGaugeBenchmark.java
@@ -1,0 +1,46 @@
+package com.codahale.metrics.benchmarks;
+
+import com.codahale.metrics.CachedGauge;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class CachedGaugeBenchmark {
+
+    private CachedGauge<Integer> cachedGauge = new CachedGauge<Integer>(100, TimeUnit.MILLISECONDS) {
+        @Override
+        protected Integer loadValue() {
+            try {
+                Thread.sleep(10);
+            } catch (InterruptedException e) {
+                throw new RuntimeException("Thread was interrupted", e);
+            }
+            return 12345;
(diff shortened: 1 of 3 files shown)


joschi (approved):
LGTM, thanks for your contribution (even with benchmarks and tests)! :heart:

PR #1573, opened 2020-05-06
Support custom floating point formatter in Graphite reporter

### LINKED ISSUE #1556: It is not easy to  change double metric format in GraphiteReporter
Hi,

I am using GraphiteReporter in my java project, which has some cpu load metric, like 0.000045322 that is cut to zero. 

But I found it is hard to change double format method, I have to extend the  GraphiteReporter and override builder. the code is not reusable.  is there a better way  in  that case?

```

protected String format(double v) {
        // the Carbon plaintext format is pretty underspecified, but it seems like it just wants
        // US-formatted digits
        return String.format(Locale.US, "%2.2f", v);
    }
```

https://github.com/dropwizard/metrics/blob/d12d3deed42e31ee8e80d9766fa609bd4792d01a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java#L401

the-thing:
https://github.com/dropwizard/metrics/issues/1556

Added support for custom floating point formatter to graphite reporter and its builder. Formatting is not tied to `java.lang.String#format()` - any custom formatter can be used. Change is backwards compatible.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -23,6 +23,7 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.DoubleFunction;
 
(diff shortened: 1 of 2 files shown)


joschi (approved):
@the-thing LGTM, thank you very much for your contribution!

PR #1575, opened 2020-05-06
Ensure datagram socket in collectd tests is bound before sending

### LINKED ISSUE #1574: Flaky CollectdReporterTest due to datagram socket creation
Hi,

While working on pull request https://github.com/dropwizard/metrics/pull/1573 Travis has failed https://travis-ci.org/github/dropwizard/metrics/builds/683822186

This can happen anytime, because sometimes datagram socket write happens, before the socket is available. I was able to capture this with RawCap.

image

All `CollectdReporterTest` test cases are affected. To recreate run the same test method until failure.

image

The issue is between 

https://github.com/dropwizard/metrics/blob/c47f3bff64805e639f1eb0187f3c98200ccf517e/metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java#L40

and

https://github.com/dropwizard/metrics/blob/c47f3bff64805e639f1eb0187f3c98200ccf517e/metrics-collectd/src/main/java/com/codahale/metrics/collectd/Sender.java#L36

I don't see a way to synchronise these two, but adding a sleep call after

https://github.com/dropwizard/metrics/blob/c47f3bff64805e639f1eb0187f3c98200ccf517e/metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java#L44

should help.

the-thing:
https://github.com/dropwizard/metrics/issues/1574

I found a better way of ensuring that datagram socket is bound before test reporter starts sending data. `com.codahale.metrics.collectd.Receiver` now uses its own instance of datagram socket to ensure that it is created, before call to listen.

Unfortunately this is a flaky test fix so the only way to verify this is to do the same test as described in the issue.

### DIFF
--- a/metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java
+++ b/metrics-collectd/src/test/java/com/codahale/metrics/collectd/Receiver.java
@@ -6,6 +6,9 @@ import org.collectd.protocol.Dispatcher;
 import org.collectd.protocol.UdpReceiver;
 import org.junit.rules.ExternalResource;
 
+import java.net.DatagramSocket;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


joschi (approved):
@the-thing Thank you very much!

PR #1576, opened 2020-05-09
Support sending Graphite metric attributes as tags

fitzoh:
# What Problem does this PR solve?

I recently submitted a [pull request](https://github.com/micrometer-metrics/micrometer/pull/1806) to [micrometer](http://micrometer.io/) to enable [graphite tag support](https://graphite.readthedocs.io/en/latest/tags.html).

Apparently I failed to actually test that end to end, and as @workless discovered in https://github.com/micrometer-metrics/micrometer/issues/2069 the PR I submitted doesn't work due to the metric attributes (e.g. `p999`, `stddev` or `m15`) appended to the metric name by the dropwizard graphite reporter.

For example, a metric with a base name of `my.metric;thing=foo` would be sent to graphite as `my.metric;thing=foo.count`, modifying the user's intended tag value.

Instead, it should be sent to graphite as `my.metric;thing=foo;metricattribute=value.count` (or something similar).

# What does this PR do?

This PR updates the GraphiteReporter to optionally send metric attributes as tags instead of using the default dot delimited format.
Defaults to the existing behavior, and does not attempt to add tag usage anywhere else within the project.

Primarily intended to allow other projects that use the graphite reporter as a library (micrometer) to support graphite tags.

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -71,6 +71,7 @@ public class GraphiteReporter extends ScheduledReporter {
         private ScheduledExecutorService executor;
         private boolean shutdownExecutorOnStop;
         private Set<MetricAttribute> disabledMetricAttributes;
+        private boolean addMetricAttributesAsTags;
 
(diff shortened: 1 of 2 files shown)


joschi (approved):
@fitzoh LGTM. Thank you very much!

PR #1578, opened 2020-05-13
Adding PATH_AND_METHOD naming strategy

### LINKED ISSUE #1529: HttpClientMetricNameStrategy should also have PATH_AND_METHOD
**Problem:**
`com.codahale.metrics.httpclient.HttpClientMetricNameStrategies` should also provide `PATH_AND_METHOD` because most of the time we are interested only in path information.

Currently, to get path metrics we use `QUERYLESS_URL_AND_METHOD`. This is not a good practice because endpoints change frequently. Suppose if we have dashboards, alerting system or other dependents in place, we need to change all of them with new metric names each time endpoint changes, it will be really painful.

**Solution**
we need something like this
```
public static final HttpClientMetricNameStrategy PATH_AND_METHOD =
            (name, request) -> {
                final URIBuilder url = new URIBuilder(requestURI(request));
                return name(HttpClient.class,
                        name,
                        url.getPath(),
                        methodNameString(request));
            };
```

shashank-babu:


joschi (approved):
LGTM.

Issue #1582, opened 2020-05-17
ExponentiallyDecayingReservoir can still corrupt state after periods of inactvity.


the-thing:
Prepare for rambling.

`ExponentiallyDecayingReservoir` random factor can sometimes favour old values. I fully understand that this is an expected error that will manifest occasionally and it is dependent on timing, generated random values and the alpha parameter.

After a period of inactivity when inserting new value, rescale will be triggered and it will clear all previous entries (depending on alpha and time). The newly inserted value will get weight ~ 1.0 (the time between a rescale "startTime" and update will be close to 0). If the RNG creates a "small" double value, it can lead to priority of this sample to be exceptionally high and further updates will be evicting each other, but a lot older favoured sample will remain.

This means inserting 1000 values in 100ms delays into size of 10 EDR might not remove the oldest value and skew the results. Yes, `com.codahale.metrics.ExponentiallyDecayingReservoirTest#longPeriodsOfInactivityShouldNotCorruptSamplingState` can currently fail. To recreate, run until failure in IDE. It takes between tens to hundreds of runs to confirm, so it is very rare. Surprisingly it failed for me on the second run (lucky me).

The paper clearly states that for priority sampling the random values are to be uniformly generated between [0, 1]. The case I'm trying to make here is that the priority random factor might be too scattered. I'm not trying to undermine the validity of implementation or the paper itself. Especially when this implementation has been for years.

When I change the random range from [0, 1] to [0.1, 1] in EDR, I am not able to fail this test. Also, I wrote a simple random benchmark that compares percentiles between EDR and "infinite" reference histogram for old and new range. The difference in percentile values is similar between two ranges (either of them performs better/worse). This would require more sophisticated statistical tests to prove.

Can somebody with better statistical understanding elaborate? Do we even care?

BTW. There are some comments mistakes in `ExponentiallyDecayingReservoirTest`. Which branch should I use for pull requests these days?

PR #1587, opened 2020-05-22
ExponentiallyDecayingReservoirTest improvements

the-thing:
Improves some of the comments and test parameters in `com.codahale.metrics.ExponentiallyDecayingReservoirTest`.

I think that https://github.com/dropwizard/metrics/issues/1582 can be closed after this is merged as the implementation is correct and statistical error is expected.

The increase of alpha value for `longPeriodsOfInactivityShouldNotCorruptSamplingState` should eliminate/reduce failure. I couldn't recrate this value.


joschi (approved):
@the-thing Thanks!

PR #1589, opened 2020-05-24
Enable reproducible builds

joschi:
> Reproducible builds are a set of software development practices that create an independently-verifiable path from source to binary code. A build is reproducible if given the same source code, build environment and build instructions, any party can recreate bit-by-bit identical copies of all specified artifacts.

* https://maven.apache.org/guides/mini/guide-reproducible-builds.html
* https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318
* https://github.com/jvm-repo-rebuild/reproducible-central
* https://reproducible-builds.org/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,6 +252,7 @@
                         </Import-Package>
                         <!-- https://stackoverflow.com/a/29853163/49505 -->
                         <_removeheaders>Bnd-LastModified</_removeheaders>
+                        <_reproducible>true</_reproducible>
                     </instructions>
                 </configuration>
             </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,7 @@
     </modules>
 
     <properties>
+        <project.build.outputTimestamp>2011-02-24T06:12:03Z</project.build.outputTimestamp>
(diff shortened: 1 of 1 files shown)


joschi:
This PR is blocked until `maven-bundle-plugin-4.2.2` has been released:
* https://issues.apache.org/jira/browse/FELIX-6203
* https://issues.apache.org/jira/browse/FELIX-6269

joschi:
It looks like `org.apache.felix:maven-bundle-plugin:5.1.1` doesn't support specifying the output timestamp (i. e. timestamp of the files in the generated JAR archive) and will always use the current timestamp.

The `maven-jar-plugin` supports setting the timestamp via [`outputTimestamp`](https://maven.apache.org/plugins/maven-jar-plugin/jar-mojo.html#outputTimestamp) (`${project.build.outputTimestamp}`) to allow for reproducible builds.

https://issues.apache.org/jira/browse/FELIX-6304

joschi:
It seems to be possible after all when passing the `-reproducible` flag to BND via the `maven-bundle-plugin`. 🥳

arteam (approved):

PR #1590, opened 2020-05-28
Avoid queuing up report tasks in ScheduledReporter

### LINKED ISSUE #1524: ScheduledReporter queues up runnables when gauge is slow
### Problem Statement

The current implementation of `ScheduledReporter` relies on a `ScheduledExecutorService` scheduled at fixed rate:
```
this.scheduledFuture = executor.scheduleAtFixedRate(runnable, initialDelay, period, unit);
```

Every `period`, a thread will report, except when one of the Gauge implementations (`getValue()`) is slower than that `period` (for example, network latency). In that case, threads will start to accumulate. When the latency is resolved, there will be a spike of report requests.

### Reproducible Scenario
To reproduce, have:
- a gauge that sleeps 10 seconds the first three times, then 0 the upcoming times before returning a value
- a ScheduledReporter with a delay of 5 seconds that prints all gauges.

You will get reports at 0, 10, 20, 30, 30, 30, 30, 35, and 40.

### Possible solution

I suggest to replace `executor.scheduleAtFixedRate` with `executor.scheduleWithFixedDelay`, so that the next call to `report` is separated by `period` with respect to the previous call, instead of having an absolute `period`.

the-thing:
Fixes https://github.com/dropwizard/metrics/issues/1524

Discussion does not seem to be active, but it could be merged if no objections.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -166,7 +166,7 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
             throw new IllegalArgumentException("Reporter already started");
         }
 
-        this.scheduledFuture = executor.scheduleAtFixedRate(runnable, initialDelay, period, unit);
+        this.scheduledFuture = executor.scheduleWithFixedDelay(runnable, initialDelay, period, unit);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


joschi:
@the-thing Thanks for your contribution! Could you please rebase your PR against the `release/4.2.x` branch?

the-thing:
The branch is up to date with 4.2.x. Do you mean change PR to merge into 4.2.x?

joschi (approved):

Issue #1126, opened 2017-05-01
Gauge that allows setting value


xiaochuanyu:
I find the current API for gauges awkward to use in certain situations where I simply just to want to set the gauge to arbitrary values. It is not always convenient to specify upfront a lambda or inner class required for providing values to a Gauge.

To illustrate, I extend the contrived example from the [Getting Started page](http://metrics.dropwizard.io/3.2.2/getting-started.html) as follows:
```
public class QueueManager {
    private final Queue queue;
   
    public QueueManager(MetricRegistry metrics, String name) {
        this.queue = new Queue();
        sizeGauge = metrics.register(MetricRegistry.name(QueueManager.class, name, "size"), new SettableLongGauge());
    }

    public void doSomethingToQueue() {
         // ... 
         sizeGauge.setValue(something);
    }

    public void doSomethingElseToQueue() {
         // ... 
         sizeGauge.setValue(somethingElse);
    }
}
```
In order to have an API like above I currently copy-paste around the classes from here:
https://github.com/signalfx/signalfx-java/tree/master/signalfx-codahale/src/main/java/com/signalfx/codahale/metrics

I'd like to propose in include this in the main library itself.

emste:
To give an idea and/or help for someone who wants to work around this: If you are using spring boot actuator, their wrapper does support updating a gauge: [DropwizardMetricServices.java#L202](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/dropwizard/DropwizardMetricServices.java#L202)

If you are not using the spring boot framework, you can still take a look at the implementation: There is a custom SimpleGauge class that implements dropwizards Gauge interface and provides an API to directly set the current Gauge value.

joschi:
@xiaochuanyu Would you be able to provide a pull request for this feature?

xiaochuanyu:
@joschi 
I'm not familiar with the code base here but I can give it a try.
Would appreciate if there is guidance on where to start.

the-thing:
@xiaochuanyu 

Look around the code first - it's not complex at all. All things have their own place. Here is my suggestion, how you could go about it.

[...]

PR #1599, opened 2020-06-07
Add SettableGauge interface and an implementation

xiaochuanyu:
This PR attempts to address https://github.com/dropwizard/metrics/issues/1126.

Example usage:
```
MetricRegistry registry = new MetricRegistry();
SettableGauge<Integer> settable = registry.register("example1", new SimpleSettableGauge<>(0));
settable.setValue(123);
```
Some things I want to point out that I'm not sure about:
1. Whats a better name than `SimpleSettableGauge`?
2. I'm not sure if we need a `MetricRegistry.settableGauge`.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -310,10 +310,10 @@ public class MetricRegistry implements MetricSet {
      * @return a new or pre-existing {@link Gauge}
      */
     @SuppressWarnings("rawtypes")
-    public Gauge gauge(String name, final MetricSupplier<Gauge> supplier) {
-        return getOrAdd(name, new MetricBuilder<Gauge>() {
+    public <T extends  Gauge> T gauge(String name, final MetricSupplier<T> supplier) {
+        return getOrAdd(name, new MetricBuilder<T>() {
             @Override
-            public Gauge newMetric() {
+            public T newMetric() {
                 return supplier.newMetric();
             }
 
--- a/metrics-core/src/main/java/com/codahale/metrics/SimpleSettableGauge.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/SimpleSettableGauge.java
@@ -1,0 +1,44 @@
+package com.codahale.metrics;
+
+/**
+ * <p>
+ * Similar to {@link Gauge}, but metric value is updated via calling {@link #setValue(T)} instead.
+ * Thread-safety is ensured via storing metric value as a volatile member variable.
+ * </p>
+ */
+public class SimpleSettableGauge<T> implements SettableGauge<T> {
+    /**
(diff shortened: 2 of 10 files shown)


the-thing:
Just a suggestion. Changing the metrics registry signature seems to be backwards compatible, but it will require some unit tests.

`public <T extends Gauge> T gauge(String name, final MetricSupplier<T> supplier);`

xiaochuanyu:
@the-thing I updated `guage()`'s signature per your suggestion 99c888f.

joschi:
@xiaochuanyu Please change the base branch to `release/4.2.x` and rebase the PR if necessary.

metrics-core/src/main/java/com/codahale/metrics/SimpleSettableGauge.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * <p>
+ * Similar to {@link Gauge}, but metric value is updated via calling {@link #setValue(T)} instead.
+ * Thread-safety is ensured via storing metric value as a volatile member variable.
+ * </p>
+ */
+public class SimpleSettableGauge<T> implements SettableGauge<T> {
+    /**
+     * Current value.
+     * Volatile so that assignment is thread-safe.
```
the-thing, line 12: You don't need to explain why volatile is there (people should know it). Probably the whole field comment is unnecessary.

metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
```diff
@@ -310,10 +310,10 @@ public boolean isInstance(Metric metric) {
      * @return a new or pre-existing {@link Gauge}
      */
     @SuppressWarnings("rawtypes")
-    public Gauge gauge(String name, final MetricSupplier<Gauge> supplier) {
-        return getOrAdd(name, new MetricBuilder<Gauge>() {
+    public <T extends  Gauge> T gauge(String name, final MetricSupplier<T> supplier) {
```
the-thing, line 313: Formatting (extra space).

metrics-core/src/main/java/com/codahale/metrics/SimpleSettableGauge.java
```diff
@@ -0,0 +1,44 @@
+package com.codahale.metrics;
+
+/**
+ * <p>
+ * Similar to {@link Gauge}, but metric value is updated via calling {@link #setValue(T)} instead.
+ * Thread-safety is ensured via storing metric value as a volatile member variable.
```
the-thing, line 6: The thread safety aspect doesn't need to be explained.

xiaochuanyu:
Hmm, I can't edit the source branch without opening a separate PR which is current `xiaochuanyu:release/4.1.x` but I want `xiaochuanyu:settable-gauge`.
I think I'm just gonna make another PR and close this one.

PR #1607, opened 2020-06-16
Add SettableGauge interface and an implementation

### LINKED ISSUE #1693: Generify MetricRegistry.gauge()
Typical usage of the `gauge` method is as follows:

    class MyCustomGauge implements Gauge<Long> { ... }
    ...
    MyCustomGauge myGauge = (MyCustomGauge) metrics.gauge("name", MyCustomGauge::new);

It would be great if the cast were unnecessary so that we could drop the repeated class name:

    MyCustomGauge myGauge = metrics.gauge("name", MyCustomGauge::new);

This could be made possible if the signature of `gauge` were:

    public <T extends Gauge<?>> T gauge​(String name, MetricRegistry.MetricSupplier<T> supplier)

The added advantage of this is that with newer versions of Java, we can then use the following syntax:

    var myGauge = metrics.gauge("name", MyCustomGauge::new);

Where `myGauge` would have the required compile-time type, and no repetition at all (DRY principle).

xiaochuanyu:
This PR was re-made from #1599 to change source (and target branch).

This PR attempts to address https://github.com/dropwizard/metrics/issues/1126.

Example usage:
```
MetricRegistry registry = new MetricRegistry();
SettableGauge<Integer> settable = registry.register("example1", new SimpleSettableGauge<>(0));
settable.setValue(123);
```


xiaochuanyu:
@joschi I've made this new PR to target `release/4.2.x` per your comment from the original PR.

@the-thing I've addressed the comments from the original PR in 9fe9105 here.

the-thing (approved):
LGTM. The name of the default implementation might not suit some people, because the existing naming convention favours "Default" prefix such as `com.codahale.metrics.jdbi3.strategies.DefaultNameStrategy` or `com.codahale.metrics.jmx.DefaultObjectNameFactory`.

joschi (approved):

joschi:
Merged in f4d84b6a83333e5e795dbfb10203307c1b81b7ef.

PR #1608, opened 2020-06-17
Metrics from Caffeine caches

john-karp:
Add ability to gather metrics from Caffeine cache library. An instance of `MetricsStatsCounter` should be passed to the Caffeine cache builder, resulting in Caffeine using that class to publish metrics to the desired metrics registry. (An example of this usage is in the UT.)

Code was taken from example given by Caffeine library author provided at https://github.com/ben-manes/caffeine/blob/master/examples/stats-metrics/src/test/java/com/github/benmanes/caffeine/examples/stats/metrics/MetricsStatsCounterTest.java

Adding it into official dropwizard metrics repo should make it easier for dependent projects to take advantage of.

It has the Apache Public License 2.0, same as this project, so I assume it is okay to integrate?

Changes made to copied sources:
 - Package names made consistent rest of project
 - Switched UT to use JUnit, added more UT
 - Used Histogram/Timer where possible
 - Break down evictions by RemovalCause
 - Documentation

### DIFF
--- a/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
+++ b/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
@@ -17,13 +17,16 @@ package com.codahale.metrics.caffeine;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.EnumMap;
 import java.util.concurrent.TimeUnit;
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import com.github.benmanes.caffeine.cache.stats.StatsCounter;
+import org.checkerframework.checker.index.qual.NonNegative;
 
 /**
  * A {@link StatsCounter} instrumented with Dropwizard Metrics.
@@ -38,6 +41,8 @@ public final class MetricsStatsCounter implements StatsCounter {
   private final Timer totalLoadTime;
   private final Counter evictionCount;
   private final Counter evictionWeight;
+  private final EnumMap<RemovalCause, Counter> evictionCountWithCause;
+  private final EnumMap<RemovalCause, Counter> evictionWeightWithCause;
 
   /**
    * Constructs an instance for use by a single cache.
@@ -54,6 +59,17 @@ public final class MetricsStatsCounter implements StatsCounter {
     loadFailureCount = registry.counter(MetricRegistry.name(metricsPrefix, "loads-failure"));
     evictionCount = registry.counter(MetricRegistry.name(metricsPrefix, "evictions"));
     evictionWeight = registry.counter(MetricRegistry.name(metricsPrefix, "evictions-weight"));
+
+    evictionCountWithCause = new EnumMap<>(RemovalCause.class);
+    evictionWeightWithCause = new EnumMap<>(RemovalCause.class);
+    for (RemovalCause cause : RemovalCause.values()) {
+      evictionCountWithCause.put(
+          cause,
+          registry.counter(MetricRegistry.name(metricsPrefix, "evictions", cause.name())));
+      evictionWeightWithCause.put(
+          cause,
+          registry.counter(MetricRegistry.name(metricsPrefix, "evictions-weight", cause.name())));
+    }
   }
 
   @Override
@@ -89,6 +105,13 @@ public final class MetricsStatsCounter implements StatsCounter {
   public void recordEviction(int weight) {
     evictionCount.inc();
     evictionWeight.inc(weight);
+  }
+
+  @Override
+  public void recordEviction(@NonNegative int weight, RemovalCause cause) {
+    evictionCountWithCause.get(cause).inc();
+    evictionWeightWithCause.get(cause).inc(weight);
+    recordEviction(weight);
   }
 
   @Override
--- a/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
+++ b/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
@@ -19,6 +19,7 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.EnumMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.LongAdder;
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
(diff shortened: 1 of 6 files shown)


john-karp:
@ben-manes see any reason not to?

ben-manes:
The only limitation is sometimes users add `cache.estimatedSize`, e.g. `CaffeineCacheMetrics` in Micrometer. I am not sure why that's desired unless one is using an unbounded cache, at which point there's inherent risk in production. The Dropwizard Ehcache2 metrics have an `objects` for the cache size, whereas the JCache one does not. Ideally size would be flat when it reaches the logical or bounded size threshold and not offer a meaningful devops data point.

Based on that, I think the push model is nice though polling like Guava's `CacheStats` is also an option. You are welcome to take that example code, or not, as you see fit.

pkwarren (approved):
Changes look good to me.

metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
```diff
@@ -0,0 +1,110 @@
...
+  public void recordEviction() {
+    // This method is scheduled for removal in version 3.0 in favor of recordEviction(weight)
+    recordEviction(1);
+  }
+
+  @Override
+  public void recordEviction(int weight) {
+    evictionCount.inc();
+    evictionWeight.inc(weight);
+  }
```
ben-manes: fyi, the example might be built on a older version of caffeine (2.7.0). 

In 2.8.0 (via https://github.com/ben-manes/caffeine/issues/304) we added the `RemovalCause` as a parameter. You might want to update to that delegate, though it has no direct bearing on your usage so it will be identical behavior.

```java
@Deprecated
default void recordEviction(@NonNegative int weight) {
  recordEviction();
}

/**
 * Records the eviction of an entry from the cache. This should only been called when an entry is
 * evicted due to the cache's eviction strategy, and not as a result of manual
 * {@link Cache#invalidate invalidations}.
 *
 * @param weight the weight of the evicted entry
 * @param cause the reason for which the entry was removed
 */
default void recordEviction(@NonNegative int weight, RemovalCause cause) {
  // This method will be abstract in version 3.0.0
  recordEviction(weight);
}
```

john-karp: Yeah, I saw that. I think assigning causes would be a use case for tagged metrics, but that's not supported in the 4.1.x branch this will be merged into. I don't know the dropwizard release management flow, but perhaps after this is merged, it could also be applied to the 5.x branch, and then there could be a followup PR against 5.x to tag with RemovalCause.

joschi: @john-karp The 5.x branch of Dropwizard Metrics is currently on ice and is not actively developed.

What do you think about having a counter for the total number of evictions (`metricsPrefix + ".evictions"` and `metricsPrefix + ".evictions-weight"`) and one counter for each removal cause (`metricsPrefix + ".evictions." + cause` and `metricsPrefix + ".evictions-weight." + cause`, for example `mycache.evictions.size` and `mycache.evictions-weight.explicit`)?

metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
```diff
@@ -0,0 +1,110 @@
...
+   */
+  public MetricsStatsCounter(MetricRegistry registry, String metricsPrefix) {
+    requireNonNull(metricsPrefix);
+    hitCount = registry.counter(metricsPrefix + ".hits");
+    missCount = registry.counter(metricsPrefix + ".misses");
+    totalLoadTime = registry.timer(metricsPrefix + ".loads");
+    loadSuccessCount = registry.counter(metricsPrefix + ".loads-success");
+    loadFailureCount = registry.counter(metricsPrefix + ".loads-failure");
+    evictionCount = registry.counter(metricsPrefix + ".evictions");
+    evictionWeight = registry.counter(metricsPrefix + ".evictions-weight");
```
joschi: Please use [`MetricRegistry#name(String, String...)`](https://javadoc.io/static/io.dropwizard.metrics/metrics-core/4.1.9/com/codahale/metrics/MetricRegistry.html#name-java.lang.String-java.lang.String...-) to generate the names.

```suggestion
    hitCount = registry.counter(MetricRegistry.name(metricsPrefix, "hits"));
    missCount = registry.counter(MetricRegistry.name(metricsPrefix, "misses"));
    totalLoadTime = registry.timer(MetricRegistry.name(metricsPrefix, "loads"));
    loadSuccessCount = registry.counter(MetricRegistry.name(metricsPrefix, "loads-success"));
    loadFailureCount = registry.counter(MetricRegistry.name(metricsPrefix, "loads-failure"));
    evictionCount = registry.counter(MetricRegistry.name(metricsPrefix, "evictions"));
    evictionWeight = registry.counter(MetricRegistry.name(metricsPrefix, "evictions-weight"));
```

metrics-caffeine/src/test/java/com/codahale/metrics/caffeine/MetricsStatsCounterTest.java
```diff
@@ -0,0 +1,51 @@
...
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import org.junit.Test;
+
+/**
+ * An example of exporting stats to Dropwizard Metrics (http://metrics.dropwizard.io).
+ *
+ * @author ben.manes@gmail.com (Ben Manes)
+ */
+public final class MetricsStatsCounterTest {
```
joschi, line 34: Could we add a few more tests for the Caffeine integration?

joschi (commented):
LGTM in general.

Would you please add a page in the documentation similar to the EhCache documentation (https://metrics.dropwizard.io/4.1.2/manual/ehcache.html)?

arteam (approved):
LGTM! Looking forward for Caffeine metrics.

metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
john-karp: OK, added metrics breaking down evictions by cause.

metrics-caffeine/src/test/java/com/codahale/metrics/caffeine/MetricsStatsCounterTest.java
john-karp: Done

john-karp:
@joschi - added documentation

joschi (approved):
@john-karp @ben-manes LGTM, thanks a lot!

ben-manes:
Thank you @john-karp for taking my simple example and making it production worthy! :)

PR #1638, opened 2020-08-24
Remove locks from ExponentialDecayReservoir

spkrka:


spkrka:
Not sure which branch I should make this PR for.

Background: we have observed contention on this class during high load for multi-threaded usecases.
This is an attempt to reduce that.

This may slightly change the behaviour in that some update operations might get dropped during rescaling (so once per hour by default). I am thinking that may be an acceptable tradeoff for improved performance.

If necessary, I can try to create some JMH benchmarks later.
I could also create this as a new class instead of changing the existing one.

I would like to hear your overall thoughts before I spend more time on the change though.

joschi:
@spkrka Thank you very much for your contribution! <3

> Not sure which branch I should make this PR for.

`release/4.2.x` is perfectly fine. I wouldn't want to introduce a substantial change like this into the Metrics 4.1.x version line in order not to break consumers which expect a certain behavior.

> If necessary, I can try to create some JMH benchmarks later.

This would be great. I'd really like to see how much better the new implementation performs compared to the old implementation.

> I could also create this as a new class instead of changing the existing one.

If the behavior is (or should be) generally the same, we can improve the existing class.

spkrka:
@joschi Thanks for the feedback! I added some JMH benchmark results in this PR: https://github.com/dropwizard/metrics/pull/1640

spkrka:
I have tested this change in one of our codebases and observed a stable reductions in latency for the 99th percentile (from 2ms to 1ms)

metrics-core/src/main/java/com/codahale/metrics/ExponentiallyDecayingReservoir.java
```diff
@@ -92,26 +89,11 @@ public void update(long value) {
...
-                        first = values.firstKey();
-                    }
-                }
-            }
-        } finally {
-            unlockForRegularUsage();
+        final State localState = this.writerState;
+        localState.update(value, timestamp);
+        final State newLocalState = this.writerState;
+        if (localState != newLocalState) {
```
spkrka, line 95: I am not sure this is needed but the idea is that it might help us avoid losing events that happen at the same time as a rescale operation triggered by another thread

spkrka:
Bump - anything else I should do or would it make sense to merge it?

joschi:
@spkrka All good from your side. We just didn't have time to do an in-depth review yet. Sorry for the delay!

spkrka:
Thanks for the reply! I understand - it's not a trivial change so it may require some careful review.

No real urgency from my side right now. I have temporarily added the change in one of our libraries that depends on this library - https://github.com/spotify/semantic-metrics/pull/80
It has worked very well for us since then.

Eventually I want to revert that change and replace it with a version bump to this.

vladimir-bukhtoyarov:
> Not sure which branch I should make this PR for.
> 
> Background: we have observed contention on this class during high load for multi-threaded usecases.
> This is an attempt to reduce that.
> 
> This may slightly change the behaviour in that some update operations might get dropped during rescaling (so once per hour by default). I am thinking that may be an acceptable tradeoff for improved performance.
> 
> If necessary, I can try to create some JMH benchmarks later.
> I could also create this as a new class instead of changing the existing one.
> 
> I would like to hear your overall thoughts before I spend more time on the change though.

Hello @spkrka 

Have you ever read about [WriterReaderPhaser](http://stuff-gil-says.blogspot.com/2014/11/writerreaderphaser-story-about-new.html)? This concept is specially designed for that you are trying to do in this request. I think it is better to upgrade this request to eliminate the probability of loss of any update operation, the license of [HdrHistogram](https://github.com/HdrHistogram/HdrHistogram) allows copying of [WriterReaderPhaser](https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/WriterReaderPhaser.java) if adding the direct dependency to HdrHistogram is not desired.

spkrka:
I had not seen that but it seems that could be an alternative solution.
But I think since this reservoir is lossy by design, it doesn't matter if have a fully phased flip, especially when the rescale only happens once per hour

vladimir-bukhtoyarov:
Yep, I forgot that ExponentiallyDecayingReservoir has a sampling nature, so events can be missed anyway in the case of a high event rate.

spkrka:
Here are the options as I see them:
1. Do nothing, keep things as they are - people can create their own Reservoir implementations in their own repos (Kind of what I have right now).
2. Merge this as it is.
3. Keep original implementation, add this with the name LossyLockFreeExponentiallyDecayingReservoir instead.
4. Someone (not me, not sure how to do it) applies the WriterReaderPhaser pattern to this.
5. Someone (not me, not sure how to do it) applies the WriterReaderPhaser pattern to the original class.

I'll let you decide what to do.

joschi:
@spkrka Thanks for the regular updates and sorry for the late response.

Would you say #1656 could replace this PR?
Would you be able to review #1656 to evaluate this? 😃

PR #1650, opened 2020-09-08
Add supportive Loggers for HealthCheckServlet

Thopap:
The current HealthCheckServlet provide details about the current state in the HTTP body to the requesting client. 
For administrative i would like to propose some additionals log statements to provide details in the logfile of the system itself.

### DIFF
--- a/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
+++ b/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
@@ -21,8 +21,12 @@ import com.codahale.metrics.health.HealthCheckRegistry;
 import com.codahale.metrics.json.HealthCheckModule;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HealthCheckServlet extends HttpServlet {
+    private static final Logger LOG = LoggerFactory.getLogger(HealthCheckServlet.class);
+    
     public static abstract class ContextListener implements ServletContextListener {
         /**
          * @return the {@link HealthCheckRegistry} to inject into the servlet context.
@@ -55,7 +59,7 @@ public class HealthCheckServlet extends HttpServlet {
 
         @Override
         public void contextDestroyed(ServletContextEvent event) {
-            // no-op
+         // no-op
         }
     }
 
@@ -114,7 +118,7 @@ public class HealthCheckServlet extends HttpServlet {
         super.destroy();
         registry.shutdown();
     }
-
+    
     @Override
     protected void doGet(HttpServletRequest req,
                          HttpServletResponse resp) throws ServletException, IOException {
@@ -122,11 +126,14 @@ public class HealthCheckServlet extends HttpServlet {
         resp.setContentType(CONTENT_TYPE);
         resp.setHeader("Cache-Control", "must-revalidate,no-cache,no-store");
         if (results.isEmpty()) {
+            LOG.error("No healthchecks implemented, return HTTP {}", HttpServletResponse.SC_NOT_IMPLEMENTED);
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
         } else {
             if (isAllHealthy(results)) {
+                LOG.debug("All health checks report state healthy: {}", results);
                 resp.setStatus(HttpServletResponse.SC_OK);
             } else {
+                LOG.warn("Healthcheck is unhealthy: {}", results);
                 resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
             }
         }
@@ -152,11 +159,6 @@ public class HealthCheckServlet extends HttpServlet {
     }
 
     private static boolean isAllHealthy(Map<String, HealthCheck.Result> results) {
-        for (HealthCheck.Result result : results.values()) {
-            if (!result.isHealthy()) {
-                return false;
-            }
-        }
-        return true;
+        return results.values().stream().allMatch(HealthCheck.Result::isHealthy);
     }
 }


joschi:
@Thopap Thanks for your contribution!

Could you please elaborate on your use case for adding the log messages?

These messages are only being logged when the health checks are being requested through `HealthCheckServlet`. But in this case, the client (part of the monitoring system?) already knows the health state of the application and should propagate it accordingly.

If the application should monitor its own health status, it probably won't do it through `HealthCheckServlet`.

Thopap:
Not all clients have good capabilities in providing the monitoring details. E.g. when using the healthcheck as kubernetes liveness probe, getting details from the log make sense, since kubernetes do not provide details from the http body. Especially if the healthcheck is unhealthy.

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -55,7 +59,7 @@ public void contextInitialized(ServletContextEvent event) {
 
         @Override
         public void contextDestroyed(ServletContextEvent event) {
-            // no-op
+         // no-op
```
pstackle, line 62: unintended change?

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -114,19 +118,22 @@ public void destroy() {
         super.destroy();
         registry.shutdown();
     }
-
+
```
pstackle, line 121: unintended change?

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -114,19 +118,22 @@ public void destroy() {
...
-
+    
     @Override
     protected void doGet(HttpServletRequest req,
                          HttpServletResponse resp) throws ServletException, IOException {
         final SortedMap<String, HealthCheck.Result> results = runHealthChecks();
         resp.setContentType(CONTENT_TYPE);
         resp.setHeader("Cache-Control", "must-revalidate,no-cache,no-store");
         if (results.isEmpty()) {
+            LOG.error("No healthchecks implemented, return HTTP {}", HttpServletResponse.SC_NOT_IMPLEMENTED);
```
pstackle, line 129: the HTTP status code of a response would show up in the app's requests log, so seems a little odd to include it here as well. 

Why is the level for this `ERROR` and not `WARN` like the other log for when a check is unhealthy?

Also, missing a space between "health" and "check".
"No healthchecks implemented" -> "No health checks implemented"

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -114,19 +118,22 @@ public void destroy() {
...
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
         } else {
             if (isAllHealthy(results)) {
+                LOG.debug("All health checks report state healthy: {}", results);
```
pstackle, line 133: I'm concerned about the noise that this log statement would cause. In an environment where the app is running with debug logging. 

My vote would be to not log here at all.

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -21,8 +21,12 @@
 import com.codahale.metrics.json.HealthCheckModule;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HealthCheckServlet extends HttpServlet {
+    private static final Logger LOG = LoggerFactory.getLogger(HealthCheckServlet.class);
```
pstackle, line 28: everywhere else in the project, `LOGGER` is used

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -114,19 +118,22 @@ public void destroy() {
...
         resp.setHeader("Cache-Control", "must-revalidate,no-cache,no-store");
         if (results.isEmpty()) {
+            LOG.error("No healthchecks implemented, return HTTP {}", HttpServletResponse.SC_NOT_IMPLEMENTED);
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
         } else {
             if (isAllHealthy(results)) {
+                LOG.debug("All health checks report state healthy: {}", results);
                 resp.setStatus(HttpServletResponse.SC_OK);
             } else {
+                LOG.warn("Healthcheck is unhealthy: {}", results);
```
pstackle, line 136: Suggested wording change:
"One or more health checks is unhealthy"

pstackle (commented):
I added a few review comments, but I tend to agree with @joschi that this doesn't quite seem like the right way to do this. 

I understand your use case, that you want to be able to see _what_ is actually unhealthy when a probe fails, but I actually think that this HealthCheckServlet isn't really well suited for use as a Kubernetes health probe. Depending on the frequency you've configured your liveness probe (as well as your readiness probe), and the kinds of health checks you configure, you may not get the best behavior. Because _all_ the health checks are run on _every_ request from one of the probes, the checks may be executed much more frequently than they should. For example, if you configure a health check that checks whether your database is still available by performing a simple query against it, then you could end up making a lot more queries against the database than you really need to or you end up configuring your probes to not execute as frequently, which could mean your deployment is less resilient to failures as it won't detect the instances' unhealthy state as quickly. 

Assuming you are using the Dropwizard framework for your application, I recommend you check out the [dropwizard-health](https://github.com/dropwizard/dropwizard-health) project (disclosure: I'm one of the original authors of it). While it doesn't yet have a good distinction between liveness and readiness probes, it _does_ allow you to configure a separate schedule for each health check  so that the execution of each check can happen at whatever interval is appropriate for the check and allows you to identify which health checks are "critical", so that you can still have a health check that gets executed to check something that isn't critical to the functioning of your app, but if that check is unhealthy, it wouldn't have an impact on the ultimate response returned to a health probe.

PR #1665, opened 2020-10-06
Remove NOTICE file

joschi:
The `NOTICE` file contains a copyright notice which is also covered by the `LICENSE` file.

The paragraph about code from JSR-166 is obsolete since these classes have been replaced with their Java 8 native counterparts.


joschi:
CC @codahale, just in case you want to object this change.

arteam (approved):

codahale:
👍

PR #1687, opened 2020-10-21
Add parameter for customizing reservoir in timers in metrics-jersey2

### LINKED ISSUE #1653: Need an option to use SlidingTimeWindowArrayReservoir for timer in instrument jersey2
### Description
Currently metrics-jersey2 is using `ExponentiallyDecayingReservoir` for timer metrics. And there is no option to change the type to use `SlidingTimeWindowArrayReservoir`. Using `SlidingTimeWindowArrayReservoir` would help if we only want 1 minute window and have spikes as well.

### Feature Request
Can you add an option to `InstrumentedResourceMethodApplicationListener` to offer `SlidingTimeWindowArrayReservoir` usage for timer metric?

joschi:
Closes #1653


huizhilu:
@joschi Thanks for making it possible!

PR #1688, opened 2020-10-21
CachedGauge: Fix reloading when Clock's time is negative

stevenschlansker:
The UserTimeClock returns System.nanoTime(), which is allowed to be negative.
CachedGauge's initialization assumes 0 is before all time, which is not always true -
leading to a gauge that never updates, until your clock rolls past 0


joschi (approved):
LGTM. Thanks a lot!

PR #1694, opened 2020-10-30
Generify MetricRegistry.gauge() - #1693

jeff303:
Making gauge method() generic, update test


jeff303:
Opened against wrong target; will start over

PR #1695, opened 2020-10-30
Generify MetricRegistry.gauge() - #1693

### LINKED ISSUE #1693: Generify MetricRegistry.gauge()

jeff303:
Making gauge method() generic, update test

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/MetricRegistry.java
@@ -309,11 +309,10 @@ public class MetricRegistry implements MetricSet {
      * @param supplier a MetricSupplier that can be used to manufacture a Gauge
      * @return a new or pre-existing {@link Gauge}
      */
-    @SuppressWarnings("rawtypes")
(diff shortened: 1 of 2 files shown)


joschi:
@jeff303 I think this has been addressed in https://github.com/dropwizard/metrics/pull/1607.
Could you please check and either close this PR or adapt it to the latest `release/4.2.x` branch? Thanks. 😃

PR #1702, opened 2020-11-07
Allow configuration of the maxlength of names sent to collectd

### LINKED ISSUE #1701: Allow collectd max length of names be configurable
Newer versions of collectd have increased the default max length to 128 as well as allowing a compile time flag to increase it further.  I propose adding a new configuration option to set the max length at runtime.

See https://github.com/collectd/collectd/pull/1690 for allowing max length configuration.

jeremy-l-ford:
Fixes #1701


joschi (approved):
LGTM. Thanks!

PR #1721, opened 2020-12-07
Properly reconnect to Graphite via TCP

twillouer:
Hi,

when DNS is down, when I start my JVM, Metrics don't send to graphite. It's normal.
But when DNS is back, Metrics never come back and graphite has never the data


metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
```diff
@@ -124,6 +124,7 @@ public void connect() throws IllegalStateException, IOException {
         }
 
         if (address.getAddress() == null) {
+            address = null;
             throw new UnknownHostException(address.getHostName());
```
joschi: This will cause a `NullPointerException` in the following line. 😉

joschi (dismissed):

metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
twillouer: ho sorry

metrics-graphite/src/main/java/com/codahale/metrics/graphite/Graphite.java
```diff
@@ -119,7 +119,7 @@ public void connect() throws IllegalStateException, IOException {
         // this version of the simplified logic will always cause a dns request if hostname has been supplied.
         // InetAddress.getByName forces the dns lookup
         // if an InetSocketAddress was supplied at create time that will take precedence.
-        if (address == null || hostname != null) {
+        if (address == null || hostname != null || address.getHostName() == null) {
```
arteam: Can you please elaborate on your case? If I'm reading correctly: if `Graphite` is created with `hostname`, we will always re-initialize the address here. If it's created with `InetSocketAddress`, but the DNS lookup failed, the next line
```
address = new InetSocketAddress(InetAddress.getByName(hostname), port)
``` 
would fail, because `hostname` is null.

twillouer: My case is the following : I start my Java application with a DNS problem: the DNS server was not reachable. After few time, he come back and all my other stuff, who need dns too, restart to work properly. I have many framework, like elatic-apm, zabbix-send, akka-http, etc.

Graphite was the only one to throw Exception (UnknownHostException), even when the dns is good.
```
2020-11-14 00:02:27 [metrics-pool-4] WARN  c.c.m.graphite.GraphiteReporter [] [] - Unable to report to Graphite
java.net.UnknownHostException: gh-carbon.mycompangy.com
```

after few days, we check our logs and start to see the problem.
There is 1440 reports of UnknownHostException by day.

Perhaps my fix is bad, but, in my comprehension, when DNS is not reachable at start of the application, address.getHostName() will always be null and there is never again a lookup.

It's why I try to address this with a forcing lookup. I will try another way, but I hope my issue is clearer now

twillouer: ok, I understand my mystake, but I don't understand why I have this result:

```
zgrep -c "java.net.UnknownHostException: gh-carbon.mycompany.com" java-program-2020111*.log.gz
java-program-20201110.log.gz:0
java-program-20201111.log.gz:0
java-program-20201112.log.gz:465
java-program-20201113.log.gz:1440
java-program-20201114.log.gz:1440
java-program-20201115.log.gz:1440
java-program-20201116.log.gz:1440
java-program-20201117.log.gz:924
java-program-20201118.log.gz:0
java-program-20201119.log.gz:0
```

I restart the service the 11/17 (with Java 11)

twillouer: FYI, ZabbixSender use the InetAddress constructor, who do the lookup but have a fallback:

```
    public InetSocketAddress(String hostname, int port) {
        checkHost(hostname);
        InetAddress addr = null;
        String host = null;
        try {
            addr = InetAddress.getByName(hostname);
        } catch(UnknownHostException e) {
            host = hostname;
        }
        holder = new InetSocketAddressHolder(host, addr, checkPort(port));
    }
```

joschi (approved):

PR #1724, opened 2020-12-08
Add support for Jetty 10.0.0 and 11.0.0

joschi:
* https://github.com/eclipse/jetty.project/releases/tag/jetty-10.0.0
* https://github.com/eclipse/jetty.project/releases/tag/jetty-11.0.0

Both, Jetty 10.0.0 and 11.0.0, are targeting Java 11 as their baseline, so that the Metrics project has to be built with JDK 11 or higher.

The target version for  `metrics-jetty10` and `metrics-jetty11` is Java 11 while keeping the other modules on Java 8.

### DIFF
--- a/metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
+++ b/metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
@@ -1,0 +1,379 @@
+package io.dropwizard.metrics.jetty10;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.RatioGauge;
+import com.codahale.metrics.Timer;
+import org.eclipse.jetty.http.HttpMethod;
+import org.eclipse.jetty.server.AsyncContextState;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.HttpChannelState;
+import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.server.handler.HandlerWrapper;
+
+import javax.servlet.AsyncEvent;
+import javax.servlet.AsyncListener;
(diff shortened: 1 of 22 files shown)


metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * @param registry the registry for the metrics
+     * @param prefix   the prefix to use for the metrics names
+     */
+    public InstrumentedHandler(MetricRegistry registry, String prefix) {
+        this.metricRegistry = registry;
+        this.prefix = prefix;
+
+        try {
+            DISPATCHED_HACK = HttpChannelState.State.valueOf("HANDLING");
```
arteam: I believe we don't need this hack in Jetty 10/11 as `HttpChannelState.State.HANDLING` will always be present there.

metrics-jetty11/src/main/java/io/dropwizard/metrics/jetty11/InstrumentedHandler.java
```diff
@@ -0,0 +1,379 @@
...
+    private Timer deleteRequests;
+    private Timer optionsRequests;
+    private Timer traceRequests;
+    private Timer connectRequests;
+    private Timer moveRequests;
+    private Timer otherRequests;
+
+    private AsyncListener listener;
+
+    private HttpChannelState.State DISPATCHED_HACK;
```
arteam: Ditto here. I believe we don't need the hack in Jetty11 either.

metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
joschi: @arteam Thank, you're absolutely right!

jplock (approved):

PR #1727, opened 2020-12-16
Backport to 4.1.x: Avoid queuing up report tasks in ScheduledReporter (#1590)

orenovadia:
This PR backports #1590 to `release/4.1.x`.

Would love to have this fix in version `4.1`.

Thank you

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -166,7 +166,7 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
             throw new IllegalArgumentException("Reporter already started");
         }
 
-        this.scheduledFuture = executor.scheduleAtFixedRate(runnable, initialDelay, period, unit);
+        this.scheduledFuture = executor.scheduleWithFixedDelay(runnable, initialDelay, period, unit);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


orenovadia:
By the way, I couldn't find when `4.2` is expected to be released.

joschi:
@orenovadia Thanks for your contribution!

The changes in #1590 were scheduled for Dropwizard Metrics 4.2.x because they change the behavior of `ScheduledReporter` and potentially break some implicit assumptions of our users.

So I'm wary of merging it into Dropwizard Metrics 4.1.x.

@dropwizard/metrics @dropwizard/committers Any opinions?

orenovadia:
I am okay with declining this PR for now if there is no interest. 

Will be interested to know when `4.2` is expected to be released though.

Thanks @joschi !

joschi:
@orenovadia I'd prefer putting the change only in Dropwizard Metrics 4.2.x so that people aren't surprised by the different behavior (even if it could be considered a bug).

I've just triggered the release of Dropwizard Metrics 4.2.0-beta.3 and it would be great if you could give it a test drive.

> Will be interested to know when `4.2` is expected to be released though.

I'd like to get https://github.com/dropwizard/metrics/pull/1806 into Dropwizard Metrics 4.2.0, but then I think we're good to go.

orenovadia:
@joschi, thanks for all the help!
Going to decline this PR

PR #1729, opened 2020-12-22
ci: add JDK 15 to build matrix

sullis:


joschi:
@sullis Thanks for your contribution! Unfortunately this won't work for now.

See 10110e7daf5bd6315e012d6963c66fb1581322f8 and 4527d40573077b3ab5cfea907cb7bc4317a57d56.

arteam:
I believe this should be possible if we apply this change https://github.com/dropwizard/metrics/commit/5e3795465357d139e503778915320f60bbea0280 from the 5.0 branch. I've been running tests in the 5.0 branch on JDK15 for a couple of weeks and it seems to be working without issues.

joschi:
> I believe this should be possible if we apply this change 5e3795465357d139e503778915320f60bbea0280 from the 5.0 branch. I've been running tests in the 5.0 branch on JDK15 for a couple of weeks and it seems to be working without issues.

@arteam @sullis Would you be able to backport the mentioned change to the `release/4.2.x` branch?
➡️ https://github.com/dropwizard/metrics/issues/1734

PR #1730, opened 2021-01-02
Add support for jakarta.servlet

joschi:


pom.xml
```diff
@@ -26,6 +26,8 @@
         <module>metrics-httpclient</module>
         <module>metrics-httpclient5</module>
         <module>metrics-httpasyncclient</module>
+        <module>metrics-jakarta-servlet</module>
+        <module>metrics-jakarta-servlets</module>
```
joschi, line 30: I thought about integrating support for the Jakarta namespace in the existing modules but couldn't come up with a robust and non-confusing way to do so.

@dropwizard/metrics @dropwizard/committers Any opinions?

jplock (approved):

PR #1759, opened 2021-02-22
Add support for Caffeine 3.x

joschi:
https://github.com/ben-manes/caffeine/releases/tag/v3.0.0

Closes #1758

### DIFF
--- a/metrics-caffeine3/pom.xml
+++ b/metrics-caffeine3/pom.xml
@@ -8,16 +8,20 @@
         <version>4.2.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>metrics-caffeine</artifactId>
-    <name>Metrics Integration for Caffeine 2.x</name>
+    <artifactId>metrics-caffeine3</artifactId>
+    <name>Metrics Integration for Caffeine 3.x</name>
     <packaging>bundle</packaging>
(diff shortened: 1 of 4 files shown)


pkwarren (approved):

ben-manes:
I believe you can support both versions in one dependency, if you prefer. The v3 change was to remove the deprecated methods,  where the newer versions were defaults that delegated to pass additional parameters. Instead simply implement those methods directly, which would make it compatible with either library version. If you implement `recordEviction()` then it will be unused but binary compatible.

```java
@Deprecated
void recordEviction();

@Deprecated
default void recordEviction(@NonNegative int weight) {
  recordEviction();
}

default void recordEviction(@NonNegative int weight, RemovalCause cause) {
  // This method will be abstract in version 3.0.0
  recordEviction(weight);
}
```

joschi:
@ben-manes Thanks for your input!

Given that Caffeine 3.x also raises the required JVM version to Java 11, I think having two separate modules for Caffeine 2.x (which can then be deprecated in a future version of Dropwizard Metrics) and Caffeine 3.x makes sense.

This being said, I'll check if we can make `metrics-caffeine` compatible with both but still depend on Caffeine 2.x by default.

PR #1765, opened 2021-02-24
Make metrics-caffeine compatible with Caffeine 2.x and 3.x

joschi:
Refs #1759

### DIFF
--- a/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
+++ b/metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
@@ -93,14 +93,17 @@ public final class MetricsStatsCounter implements StatsCounter {
     totalLoadTime.add(loadTime);
   }
 
-  @Override
+  // @Override -- Caffeine 2.x
+  @Deprecated
   @SuppressWarnings("deprecation")
   public void recordEviction() {
     // This method is scheduled for removal in version 3.0 in favor of recordEviction(weight)
     recordEviction(1);
   }
 
-  @Override
+  // @Override -- Caffeine 2.x
+  @Deprecated
+  @SuppressWarnings("deprecation")
   public void recordEviction(int weight) {
     evictions.update(weight);
     evictionWeight.inc(weight);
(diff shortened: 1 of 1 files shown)


joschi:
@ben-manes Thanks for your hints in https://github.com/dropwizard/metrics/pull/1759! ❤️

metrics-caffeine/src/main/java/com/codahale/metrics/caffeine/MetricsStatsCounter.java
```diff
@@ -114,7 +117,7 @@ public void recordEviction(@NonNegative int weight, RemovalCause cause) {
 
   @Override
   public CacheStats snapshot() {
-    return new CacheStats(
+    return CacheStats.of(
```
joschi, line 120: This factory method has been introduced in Caffeine 2.9.0 and is not available in older Caffeine 2.x releases:
[`CacheStats#of(...)`](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/2.9.0/com/github/benmanes/caffeine/cache/stats/CacheStats.html#of-long-long-long-long-long-long-long-)

At the same time, the constructors of `CacheStats` have been made private in Caffeine 3.x, so that there's currently not way to support Caffeine 3.x and Caffeine <2.9.0 with the same code.

@dropwizard/committers @dropwizard/metrics Given that there's a dedicated module for Caffeine 3.x in Dropwizard Metrics 4.2.x, should we pass on this PR or do we accept that `dropwizard-metrics` will only work with Caffeine 3.x and 2.9.x?

ben-manes: Just fyi that I did try to be a good citizen by following semver strictly. My hope is that upgrading minor releases is safe and desirable as it includes bug fixes.

joschi: > Just fyi that I did try to be a good citizen by following semver strictly.

Yes, much appreciated! 🤗

arteam: I'm in favor in supporting Caffeine from 2.9.x and 3.x!

arteam (approved):

PR #1821, opened 2021-03-23
Override healthcheck HTTP status

maffe:
Some monitoring tools only parse responses with status 200. With this change status 500 can be disabled.

### DIFF
--- a/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
+++ b/metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
@@ -124,7 +124,8 @@ public class HealthCheckServlet extends HttpServlet {
         if (results.isEmpty()) {
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
         } else {
-            if (isAllHealthy(results)) {
+            final boolean alwaysOk = Boolean.parseBoolean(req.getParameter("alwaysOk"));
+            if (alwaysOk || isAllHealthy(results)) {
                 resp.setStatus(HttpServletResponse.SC_OK);
             } else {
                 resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);


metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -124,7 +124,8 @@ protected void doGet(HttpServletRequest req,
         if (results.isEmpty()) {
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
         } else {
-            if (isAllHealthy(results)) {
+            final boolean alwaysOk = Boolean.parseBoolean(req.getParameter("alwaysOk"));
```
joschi, line 127: Could we rename the parameter to `overrideStatusCode`?

The name `alwaysOk` implies that the health check result would be always healthy. 😉

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
```diff
@@ -124,7 +124,8 @@ protected void doGet(HttpServletRequest req,
         if (results.isEmpty()) {
             resp.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
```
joschi, line 125: Shouldn't this also be 200 (OK) if the flag was enabled?

joschi (commented):
@maffe Thanks for your contribution! 🙂

I've added a few remarks about your PR.

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
maffe: Initially I planned to name it similar to your proposal, but then I thought this would suggest the parameter value should be the desired status code (`overrideStatusCode=200`) instead of a boolean. I included `OK` because this is the official name for HTTP status 200. Also, I think `OK` differs from `healthy` enough to not suggest it would affect those entries.

Maybe `httpStatusOk=true`?

metrics-servlets/src/main/java/com/codahale/metrics/servlets/HealthCheckServlet.java
maffe: I don’t think so. The point is to make it possible to parse the content. If there are no health checks at all, there’s nothing to parse. In that case it is valid (and useful) for the monitoring tool to report a failed request.

pstackle:
This approach is a bit different than what I'd expect. It seems a little strange to have the client be able to dynamically influence the response status code of the `HealthCheckServlet` via a query string parameter. One of the main points of the `HealthCheckServlet` is to indicate the overall health of the running instance by having that servlet return a 200 status code if everything is healthy but return a server error (5xx) if it's not healthy.

If the behavior _was_ to be configurable, I would expect it to be configured through the `ServletConfig` via the `init` method instead of via a query string parameter. Supporting config from the calling client might be setting a bad precedent. 

@maffe Can you elaborate a bit on your use case and which monitoring tools you are having trouble with? 

Alternatively, a workaround could be to use a servlet filter to swallow the 500 status code from the `HealthCheckServlet` response and turn it into a 200 status code instead.

maffe:
Currently a [PRTG HTTP Sensor](https://www.paessler.com/manuals/prtg/http_sensor) is used to retrieve the overall status. The goal is to monitor each healthcheck individually, but the [REST Custom Sensor](https://www.paessler.com/manuals/prtg/rest_custom_sensor) just shows gaps for everything when the status is 500.

Disabling the status 500 via query parameter would make it possible to use both types of sensors with minimal code change in Dropwizard. Using static configuration would also work when the basic HTTP Sensor is not needed.

joschi:
@maffe @pstackle I've included this change in https://github.com/dropwizard/metrics/pull/1871 and made it possible to set the general behavior with a Servlet attribute.

joschi:
Resolved in #1871.

PR #1872, opened 2021-04-15
Build with Java 16

joschi:


arteam:
It looks like the build passes now! I've opened the Javac internals to error-prone as recommended in https://errorprone.info/docs/installation, and had to change some tests that spied on JDK classes, hence depending on their internal implementations.

jplock (approved):

PR #2031, opened 2021-06-16
Fix for #2030, where JMXReporter may create invalid ObjectNames

### LINKED ISSUE #2030: Certain metric names cause MalformedObjectNameException in JMX clients
The default managed object naming strategy used by JMX reporter does not handle metric names with quotable characters appropriately, causing various JMX clients like jmxtrans to fail RMI object deserialization with `MalformedObjectNameException`. Here's a reproducer code snippet:

```java
DefaultObjectNameFactory f = new DefaultObjectNameFactory();
ObjectName on = f.createName("type", "com.domain", "something.with.quotes(\"ABcd\")");
assertThatCode(() -> new ObjectName(on.toString())).doesNotThrowAnyException();
```

The object `on` is created without raising exceptions, however parsing its serialized form (which is what happens in clients like jmxtrans) causes `MalformedObjectNameException` to be thrown.

It is not clear to me whether the JDK should be checking for validly quoted values for object name properties, or whether the onus is on clients of the javax.management APIs to appropriately shape the values.

Any metric name containing the following characters result in this behaviour: `" , = :`

amrith92:
Fixes #2030.

JMX managed ObjectNames cannot contain properties with values that are unquoted if the aforementioned values contain any of comma, equals, colon or double-quote characters. This fixes the default object naming strategy used by JMX reporter.
Note that construction of the `ObjectName` does not fail, but JMX clients break on deserialization.

### DIFF
--- a/metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
+++ b/metrics-jmx/src/main/java/com/codahale/metrics/jmx/DefaultObjectNameFactory.java
@@ -10,6 +10,7 @@ import org.slf4j.LoggerFactory;
 
 public class DefaultObjectNameFactory implements ObjectNameFactory {
 
+    private static final char[] QUOTABLE_CHARS = new char[] {',', '=', ':', '"'};
     private static final Logger LOGGER = LoggerFactory.getLogger(JmxReporter.class);
 
     @Override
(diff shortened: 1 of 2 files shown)


joschi (approved):
@amrith92 Thank you for your contribution!

The changes look good to me. 👍

PR #2182, opened 2021-10-02
Add missing modules to `metrics-bom`

zUniQueX:
In the `metrics-bom` module there are no specifications for `metrics-httpclient5` and `metrics-caffeine3`. As Jetty and Jersey have different versions in the bom too, I think these modules can be added as well.


joschi (approved):
@zUniQueX Thanks for your contribution!

PR #2193, opened 2021-10-12
Migrate tests to JUnit 5

arteam:


joschi:
@arteam Could you please rebase this pull request on the `release/5.0.x` branch?

I'm hesitant applying such a large change set to an already released version of Dropwizard Metrics even if it shouldn't change the release artifacts.

PR #2413, opened 2022-01-28
Ensure that start times aren't overwritten by concurrent asynchronous…

### LINKED ISSUE #2121: Jetty InstrumentedHttpChannelListener races to track async request times
A single version of the `InstrumentedHttpChannelListener` is used to track all requests. The `InstrumentedHttpChannelListener` creates a single [AsyncListener](https://github.com/dropwizard/metrics/blob/release/4.2.x/metrics-jetty9/src/main/java/com/codahale/metrics/jetty9/InstrumentedHttpChannelListener.java#L172-L173) and stores the request time for async requests in a field on this class. This is overwritten everytime a new async request is started, in high throughput scenarios this results in the latency of async requests being drastically underreported. Since on completion, the time since the last async request started is reported as the latency, rather than the time the request took. This issue persists across all versions of the `InstrumentedHttpChannelListener`. The fix here is straightforward, a new listener must be constructed for each async request. Or the start time must otherwise be stored uniquely for each async request.

jschlather:
… requests.

Resolves https://github.com/dropwizard/metrics/issues/2121

To limit increasing the memory footprint, I added an `AsyncAttachingListener` to replace the old anonymous `AsyncListener`. Which creates a new listener if an async request is actually processed, which has a final `startTime` that cannot be overwritten. I made the change in `InstrumentedHandler` and `InstrumentedHttpChannelListener` for each of the jetty versions, since the race condition exists in each one. The tests seem to be passing, but I can probably add something if the current tests aren't viewed as sufficient. I do notice the async tests are marked with ignore.

### DIFF
--- a/metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
+++ b/metrics-jetty10/src/main/java/io/dropwizard/metrics/jetty10/InstrumentedHandler.java
@@ -209,36 +209,7 @@ public class InstrumentedHandler extends HandlerWrapper {
         });
 
 
-        this.listener = new AsyncListener() {
-            private long startTime;
-
-            @Override
-            public void onTimeout(AsyncEvent event) throws IOException {
-                asyncTimeouts.mark();
-            }
-
-            @Override
-            public void onStartAsync(AsyncEvent event) throws IOException {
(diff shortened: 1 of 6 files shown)


joschi (approved):
@jschlather LGTM, thanks for your contribution!

PR #2639, opened 2022-06-01
Use Gauge<String> values as metric names

baunz:
A lot of metric use gauges of type String to report the status of a component (e.g. failed / running). Currently all those values are discared since they are not numeric. This commit appends the value of the gauge to the metric name and sends it with a constant value of 1.

<a href="https://gitpod.io/#https://github.com/dropwizard/metrics/pull/2639"></a>

### DIFF
--- a/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
+++ b/metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java
@@ -474,9 +474,18 @@ public class GraphiteReporter extends ScheduledReporter {
     }
 
     private void reportGauge(String name, Gauge<?> gauge, long timestamp) throws IOException {
-        final String value = format(gauge.getValue());
+        Object value = gauge.getValue();
+        String finalMetricName;
+        String finalValue;
+        if(value instanceof String) {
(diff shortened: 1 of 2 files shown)


baunz:
I am sceptic if this change is too rigourous as it would produce a lot of metrics suddenly for people with Gauge<String> usage, e.g. Jvm Version. Maybe it would be better to have an explicit StringGauge? This way clients could decide if they want to report Strings? 

But I am not too convinced of this either. Maybe it's a problem not solvable...

joschi:
@baunz Thanks for reporting this and providing a pull request for your suggested change.

I'm a bit hesitant to merge this in Metrics 4.2.x because, as you've said yourself, it would be a breaking change and people would suddenly see a lot of new metrics in their Graphite-compatible metrics stores.

If you depend on this change I'd suggest creating a custom Metrics reporter, basically taking the same code you've contributed in this PR.

@dropwizard/metrics Any thoughts on making this change in Metrics 4.3.x?

baunz:
@joschi Thanks for your feedback, the more I think about this I feel that it really is better fitted as a customization place in a separate reporter. My main target was the Kafka Connect framework where there is an important status metric I would have loved to consume via the standard reporter. I am closing this now and think if I can come up with a solution external to metrics..BR Johannes

PR #2642, opened 2022-06-02
Add a more type safe way of registering gauges

arteam:
Add the `registerGauge` method to `MetricRegistry` that accepts `Gauge<T>` instead of `Metric` which makes it possible to register a gauge as a lambda expression without declaring its type explicetly.


joschi:
@arteam Anything you want to polish/improve in this pull request?

Happy to merge it otherwise. 😃

arteam:
@joschi I think it should be good to go :+1:

joschi (approved):
LGTM. Thanks! 👍

Issue #2664, opened 2022-06-09
ScheduledReporter may have missed reports


dkaukov:
### Problem Statement

I think, fix for https://github.com/dropwizard/metrics/issues/1524 has undesired side-effects in our environment. Because `executor.scheduleWithFixedDelay` accumulating errors (due network latency, for example) when it goes over reporting period, we have "holes" in graphite metrics:
![image](https://user-images.githubusercontent.com/4616540/172738427-0dc9f4d6-b655-4fbe-a62e-91dd8bcd9d99.png)

### Possible solution
In our case, report bursts are probably lesser evil than holes, so maybe we can make it configurable.

mihalyr:
We are seeing the same problem which is a result of #1524 and the switch to `scheduleWithFixedDelay`

We are sending `counter` metrics every minute to Graphite, Carbon Aggregator and we use `nonNegativeDerivative` to show only the changes per minute and we are seeing "fake" metrics.

![image](https://user-images.githubusercontent.com/26523902/182618689-162c7639-f07b-41cd-a963-83b111344727.png)

The `derivative` function is sensitive to missing metrics, if one of the hosts aggregated is not sending and then sends again, it will show up as an increase in the metric for that minute. We are aggregating to hours in the above graph and the results are completely wrong, there is nothing happening on the server and all those bars are caused by missing metrics.

Here is an example of missing metrics showing up from two hosts:

![image](https://user-images.githubusercontent.com/26523902/182619399-58a23fd8-beff-4505-a8e0-300352bfb462.png)

I've captured the metrics on the Graphite server with tcp dump and saw the following reports from a single source server:

* arrived `08:08:59.087611` metric payload timestamp `1659514138 (08:08:58)`
* arrived `08:09:59.343252` metric payload timestamp `1659514199 (08:09:59)`
* arrived `08:10:59.597794` metric payload timestamp `1659514259 (08:10:59)`
* arrived `08:11:59.856127` metric payload timestamp `1659514319 (08:11:59)`
* arrived `08:13:00.118483` metric payload timestamp `1659514380 (08:13:00)`

You can see that we never receive data for `08:12`, that's 1-min gap on the charts and one spike in derivatives.
First I suspected network latency issues and missing aggregation buckets on Graphite, but from the timestamps added by GraphiteReporter it is clear that the reporter is sending them late and missing the buckets. The increasing delay was an immediate sign of `scheduleWithFixedDelay` which was confirmed in the codebase.

[...]

PR #2931, opened 2022-10-15
Allow override of scheduler

mikebell90:
https://github.com/dropwizard/metrics/issues/1524 introduced a fix to queuing issues and ends up doing atFixedDelay rather than atFixedRate. The logic is sound enough, but leads to frustrating unanticipated issues.

Issues with binning exist in Graphite (and probably others). These accumulate "drift", and so metric gaps occur. See #2664 for two examples.

Hence allowing this to be configurable is desirable.

This is the lamest but easiest form of this - moving the method to an overridable method.

Usage would be (example graphite reporter)
* Create your own subclass of GraphiteReporter, override getScheduledFuture.
* Said class will also need to explicitly set the constructor instead of using the builder.

Pseudo code:
```
public class MyReporter extends GraphiteReporter {
     public MyReporter(....) {
         super(...)
    }

    @Override 
    public ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable) {
        return executor.scheduleWithFixedRate(runnable, initialDelay, period, unit);
   }
}
```

I looked at other options but they all involved breaking or overloading constructors too much:
* making it an abstract method, and implementing at concrete classes. 
    * There are about 5 , and their constructors are hidden by builders. One could add to builders (with a lot of tedious copy/paste) but the constructors are heavily overloaded and would break other subclassers or add lots of jargon
    * Move the logic to base class. That would involve changing another heavily overloadd constructor (ScheduledReporter)

There's a lot of unfortunate outdated api design in DropWizard
* Not enough interfaces.
* Too many overloaded constructors instead of config objects.
* Lack of centralized configuration options.
* Often using concrete classes instead of interfaces, which imposes many constraints.

Those are not fixable without a major breaking version change. But we can at least offer something here.

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -166,7 +166,19 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
             throw new IllegalArgumentException("Reporter already started");
         }
 
-        this.scheduledFuture = executor.scheduleWithFixedDelay(runnable, initialDelay, period, unit);
+        this.scheduledFuture = getScheduledFuture(initialDelay, period, unit, runnable);
+    }
+
+
+    /**
+     * Schedule the task, and return a future.
+     * The current implementation uses scheduleWithFixedDelay, replacing scheduleWithFixedRate. This avoids queueing issues, but may
+     * cause some reporters to skip metrics, as scheduleWithFixedDelay introduces a growing delta from the original start point.
+     *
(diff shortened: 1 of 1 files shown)


joschi (approved):
@mikebell90 Thanks for your contribution!

mikebell90:
@joschi unfortunately (in the middle of an incident) but I think

https://github.com/mikebell90/metrics/blob/9049dd3bd402d8668c1b6a41ebc50bba59aa2340/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java#L59

needs to be corrected. That needs to be protected or overriding class can't access. Thought that was submitted.

PR #3043, opened 2022-12-19
Add response code meters for ResponseMetered annotation

dennyac:
Currently the ResponseMetered annotation gives us top level 1xx/2xx/3xx/4xx/5xx meters. We have many use cases where we care about specific response codes like 401/503 and end up manually instrumenting code to create these meters. This PR adds meters for individual response codes in addition to the 1xx/2xx/3xx/4xx/5xx meters in the ResponseMetered annotation.

Hoping others find this useful and I'm open to suggestions/feedback. 

Happy to add this functionality to metrics-jersey3 and metrics-jersey31 as well if this approach is acceptable. Can also create a PR to add individual response code meters to the jetty InstrumentedHandler

### DIFF
--- a/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
+++ b/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
@@ -135,7 +135,7 @@ public class InstrumentedResourceMethodApplicationListener implements Applicatio
     private static class ResponseMeterMetric {
         private static final Set<ResponseMeteredLevel> COARSE_METER_LEVELS = EnumSet.of(COARSE, ALL);
         private static final Set<ResponseMeteredLevel> DETAILED_METER_LEVELS = EnumSet.of(DETAILED, ALL);
-        public final List<Meter> meters;
+        private final List<Meter> meters;
         private final Map<Integer, Meter> responseCodeMeters;
         private final MetricRegistry metricRegistry;
         private final String metricName;
@@ -171,7 +171,7 @@ public class InstrumentedResourceMethodApplicationListener implements Applicatio
             }
         }
 
-        public Meter getResponseCodeMeter(int statusCode) {
+        private Meter getResponseCodeMeter(int statusCode) {
             return responseCodeMeters
                     .computeIfAbsent(statusCode, sc -> metricRegistry
                             .meter(name(metricName, String.format("%d-responses", sc))));
--- a/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
+++ b/metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
@@ -27,6 +27,7 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
@@ -127,11 +128,14 @@ public class InstrumentedResourceMethodApplicationListener implements Applicatio
      */
     private static class ResponseMeterMetric {
         public final List<Meter> meters;
+        private final ConcurrentMap<Integer, Meter> responseCodeMeters = new ConcurrentHashMap<>();
+        private final MetricRegistry metricRegistry;
+        private final String metricName;
 
         public ResponseMeterMetric(final MetricRegistry registry,
                                    final ResourceMethod method,
                                    final ResponseMetered responseMetered) {
-            final String metricName = chooseName(responseMetered.name(), responseMetered.absolute(), method);
+            this.metricName = chooseName(responseMetered.name(), responseMetered.absolute(), method);
             this.meters = Collections.unmodifiableList(Arrays.asList(
                     registry.meter(name(metricName, "1xx-responses")), // 1xx
                     registry.meter(name(metricName, "2xx-responses")), // 2xx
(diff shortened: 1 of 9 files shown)


joschi (changes requested):
@dennyac Thanks for your contribution! It looks good in general.

Could you please make this change backward-compatible and optional by adding a flag to the `ResponseMetered` annotation which controls whether the detailed, the "coarse" (2xx, etc.), or both styles status codes are recorded?

Maybe an enum would make sense here: `COARSE` (default), `DETAILED`, and `ALL` (because `BOTH` might not be true anymore if another mode was added 😅).

dennyac:
@joschi - That makes sense. I'll incorporate that feedback. Thanks!

dennyac:
@joschi - I've pushed changes based on your feedback. Please review when you get time.

Also, let me know if I should add this functionality to metrics-jersey3 and metrics-jersey31 modules as well (in a separate PR or the same PR)

dennyac:
Happy New Year @joschi !!

Bumping this PR for review when you have time.

metrics-jersey2/src/test/java/com/codahale/metrics/jersey2/SingletonMetricsResponseMeteredPerClassJerseyTest.java
```diff
@@ -126,14 +126,21 @@ public void responseMeteredUnmappedExceptionPerClassMethodsAreMetered() {
 
     @Test
     public void subresourcesFromLocatorsRegisterMetrics() {
+        final Meter meter2xx = registry.meter(name(InstrumentedSubResourceResponseMeteredPerClass.class,
+                "responseMeteredPerClass",
+                "2xx-responses"));
+        final Meter meter201 = registry.meter(name(InstrumentedSubResourceResponseMeteredPerClass.class,
```
joschi: question: Should the variable name be `meter200` or the name suffix be `201-responses`?

Given the test class, I would expect the variable name to be different.

metrics-jersey2/src/test/java/com/codahale/metrics/jersey2/SingletonMetricsResponseMeteredPerClassJerseyTest.java
```diff
@@ -126,14 +126,21 @@ public void responseMeteredUnmappedExceptionPerClassMethodsAreMetered() {
...
+                "responseMeteredPerClass",
+                "200-responses"));
+
+        assertThat(meter2xx.getCount()).isZero();
+        assertThat(meter2xx.getCount()).isZero();
```
joschi: question: Should this be `meter200.getCount()`?

metrics-jersey2/src/test/java/com/codahale/metrics/jersey2/SingletonMetricsResponseMeteredPerClassJerseyTest.java
```diff
@@ -126,14 +126,21 @@ public void responseMeteredUnmappedExceptionPerClassMethodsAreMetered() {
...
                 .request()
                 .get().getStatus())
                 .isEqualTo(200);
 
-        final Meter meter = registry.meter(name(InstrumentedSubResourceResponseMeteredPerClass.class,
-                "responseMeteredPerClass",
-                "2xx-responses"));
-        assertThat(meter.getCount()).isEqualTo(1);
+        assertThat(meter2xx.getCount()).isOne();
+        assertThat(meter2xx.getCount()).isOne();
```
joschi: question: Should this be `meter200.getCount()`?

metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
```diff
@@ -126,19 +133,48 @@ public ExceptionMeterMetric(final MetricRegistry registry,
...
+
+            if (COARSE_METER_LEVELS.contains(level)) {
+                final int responseStatus = statusCode / 100;
+                if (responseStatus >= 1 && responseStatus <= 5) {
+                    meters.get(responseStatus - 1).mark();
+                }
+            }
+        }
+
+        public Meter getResponseCodeMeter(int statusCode) {
```
joschi: question: Does this method need to be `public` or would a more restrictive visibility scope also work?

joschi (changes requested):

metrics-jersey2/src/main/java/com/codahale/metrics/jersey2/InstrumentedResourceMethodApplicationListener.java
dennyac: Changed to private

metrics-jersey2/src/test/java/com/codahale/metrics/jersey2/SingletonMetricsResponseMeteredPerClassJerseyTest.java
dennyac: Good catch! Fixed!

metrics-jersey2/src/test/java/com/codahale/metrics/jersey2/SingletonMetricsResponseMeteredPerClassJerseyTest.java
dennyac: Good catch! Fixed!

dennyac:
Thanks for the review @joschi !! Addressed your comments.

joschi (approved):

joschi:
@dennyac Thanks a lot for your contribution and the changes you've made to this PR! 😄 

> Happy to add this functionality to metrics-jersey3 and metrics-jersey31 as well if this approach is acceptable. Can also create a PR to add individual response code meters to the jetty InstrumentedHandler

If you want to invest the time into creating PRs for these I'll happily review and merge them. 😉

dennyac:
@joschi - Thanks for taking the time to review the PRs. I'll create PRs for metrics-jersey3 and metrics-jersey31

PR #3125, opened 2023-01-24
[JVM] Fix maximum/total memory calculation

### LINKED ISSUE #2609: jvm.memory.total.max calculated incorrectly
`jvm.memory.total.max` is calculated by adding the heap maximum and the non-heap maximum:

https://github.com/dropwizard/metrics/blob/7a755288e2290cbcb0caab9d10a07994b4988012/metrics-jvm/src/main/java/com/codahale/metrics/jvm/MemoryUsageGaugeSet.java#L50-L51

This is fine when both the heap maximum and the non-heap maximum are set. But when either is undefined (has value `-1`) I think this is incorrect. You can end up with a situation where `total.use` is larger than `total.max`, see the table below:

| type | .use  | .max |
| --- | ---: | ---: |
| heap | 210.47M | 259.52M |
| non-heap | 256.97M | -1 |
| total | 467.44M | 259.52M |

My suggestion would be letting `jvm.memory.total.max` return `-1` when either the heap maximum or the non-heap maximum is `-1`.

the-thing:
Fixes https://github.com/dropwizard/metrics/issues/2609

**Changes**
- when max non-heap memory is undefined then total.max is equal to -1
- when max non-heap memory is undefined then "non-heap.usage" ratio is non-heap.used / non-heap.commited (instead of non-heap.used / non-heap.max - similar to memory pools usage calculation)


joschi (approved):
@the-thing Thank you for your contribution!

PR #3132, opened 2023-01-30
Allow override of scheduler

dkaukov:
This is addition to the https://github.com/dropwizard/metrics/pull/2931 Basically, to override `getScheduledFuture`, we need access to the `executor`

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -167,9 +167,20 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
             throw new IllegalArgumentException("Reporter already started");
         }
 
-        this.scheduledFuture = getScheduledFuture(initialDelay, period, unit, runnable, executor);
+        this.scheduledFuture = getScheduledFuture(initialDelay, period, unit, runnable);
     }
 
+
+    /**
+     * Schedule the task, and return a future.
+     *
+     * @deprecated Use {@link #getScheduledFuture(long, long, TimeUnit, Runnable, ScheduledExecutorService)} instead.
+     */
+    @SuppressWarnings("DeprecatedIsStillUsed")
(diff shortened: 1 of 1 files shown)


metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -177,7 +178,7 @@ synchronized void start(long initialDelay, long period, TimeUnit unit, Runnable
      *
      * Overriding this in a subclass to revert to the old behavior is permitted.
      */
-    protected ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable) {
+    protected ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable, ScheduledExecutorService executor) {
         return executor.scheduleWithFixedDelay(runnable, initialDelay, period, unit);
     }
```
joschi, line 194: issue(blocking): This is a breaking change for anybody extending `ScheduledReporter` and overriding `ScheduledReporter#getScheduledFuture(...)`.

Please add a new method instead, delegate this method to the new one, and mark the old method as deprecated.

```java
    /**
     * [...]
     * @deprecated [...]
     */
    @Deprecated
    protected ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable) {
        return getScheduledFuture(initialDelay, period, unit, runnable, this.executor);
    }

    protected ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable, ScheduledExecutorService executor) {
        return executor.scheduleWithFixedDelay(runnable, initialDelay, period, unit);
    }

joschi (changes requested):
@dkaukov Thanks for your contribution! Since we want to avoid breaking changes in minor or patch versions, please address the requested change. 😄

joschi (approved):

PR #3133, opened 2023-01-30
Add response code meters for jetty9

dennyac:
Similar to https://github.com/dropwizard/metrics/pull/3043, we have use cases where we want to instrument specific server response codes instead of the aggregated  1xx/2xx/3xx/4xx/5xx meters.

This PR adds configurable individual response code meters in jetty. If this approach is acceptable, I'll create an accompanying PR in dropwizard to allow this to be specified in config via [AbstractServerFactory](https://github.com/dropwizard/dropwizard/blob/release/2.1.x/dropwizard-core/src/main/java/io/dropwizard/server/AbstractServerFactory.java)

This change is backwards compatible and the default behavior will only include the existing 1xx/2xx/3xx/4xx/5xx meters. Can port this over to metrics-jetty10 and metrics-jetty11 modules as well.

### DIFF
--- a/metrics-jetty9/pom.xml
+++ b/metrics-jetty9/pom.xml
@@ -43,7 +43,10 @@
         <dependency>
             <groupId>io.dropwizard.metrics</groupId>
             <artifactId>metrics-core</artifactId>
-            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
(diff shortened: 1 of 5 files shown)


dennyac:
FYI - @joschi

joschi (approved):
@dennyac Thanks a lot for your numerous contributions!

It would be great if you would also port this feature to the modules for Jetty 10.x and 11.x. 😄

PR #3149, opened 2023-02-07
feat: add sum aggregation to timers and histograms

kjschnei001:
The purpose of this proposed change is to facilitate the simple calculation of averages for reporters that poll at arbitrary intervals. By including a sum aggregation alongside the the existing count aggregation, we can calculate the average between any two datapoints by calculating diff(sum)/diff(mean).


joschi (changes requested):
@kjschnei001 Thanks for your contribution! Could you please revert the non-functional changes (whitespace/linebreaks) in your PR?

kjschnei001:
Thanks @joschi. I've reverted the formatting changes.

PR #3175, opened 2023-02-12
Add detailed response code meters for Jetty 11.x

joschi:
Refs #3133


dennyac:
@joschi - I was about to port the changes over to the other jetty modules, but came across this PR. Thanks! 😊

PR #3202, opened 2023-02-22
InstrumentedExecutorService registered metrics cleanup after shutdown

### LINKED ISSUE #2920: InstrumentedExecutorService does not remove registered gauges after shutdown
Similar to https://github.com/dropwizard/metrics/issues/1684.

As the gauges are not removed, once we shutdown the executor service and try to start again with the same name, the executor service can no longer be started.

the-thing:
Fixes https://github.com/dropwizard/metrics/issues/2920

Additional metrics registed in `InstrumentedExecutorService` constructors are cleaned up after shutdown.


joschi (approved):
@the-thing Thanks a lot for your contribution! ❤️

PR #3448, opened 2023-07-20
Fix formatting in servlets.rst

maffe:
avoids linebreak


joschi (approved):
@maffe Thanks a lot! ❤️

PR #3578, opened 2023-09-21
Fix Jetty 12/Jakarta EE 10 InstrumentedEE10Handler

joschi:


metrics-jetty12-ee10/src/test/java/io/dropwizard/metrics/jetty12/ee10/InstrumentedEE10HandlerTest.java
```diff
@@ -118,6 +118,7 @@ public void createsAndRemovesMetricsForTheHandler() throws Exception {
     }
 
     @Test
+    @Ignore("flaky on virtual machines")
```
joschi, line 121: Would be great if we eventually found out why this is flaky. 🙈

PR #4181, opened 2024-05-23
Correction for the Jetty-12 QTP metrics

dkaukov:
This uses Jetty-12 recommended way to calculate thread pool utilization. Taking into account number of threads that are leased to internal components, and therefore cannot be used to execute transient jobs.


joschi (approved):
@dkaukov Awesome, thanks for your contribution!

PR #4188, opened 2024-05-23
feature request: make executor rescheduling configurable - no lambda implementation proposal

github-louis-fruleux:
Hello :wave:
First of all, thanks for maintaining the project, we appreciate it.

We had an issue when using Grafana, actually, the exact same as described here: https://github.com/dropwizard/metrics/issues/2664

While having the possibility to override it is nice, we lose a lot of things by doing so, including the nice API builder on the graphite report.
The PR aims to make it configurable with the constructor.

Would that make sense to move forward with this PR?
Feel free to make any comments, long time I have not done any Java. I might not be aware of the good practices

PS: I made another PR proposition here, https://github.com/dropwizard/metrics/pull/4192. Which does not use lambdas and is maybe cleaner to read / maintain. What do you think?

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
@@ -67,6 +67,8 @@ public abstract class ScheduledReporter implements Closeable, Reporter {
     private final long rateFactor;
     private final String rateUnit;
 
+    private final GetScheduledFuture getScheduledFuture;
+
     /**
      * Creates a new {@link ScheduledReporter} instance.
      *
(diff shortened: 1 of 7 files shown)


metrics-core/src/main/java/com/codahale/metrics/ScheduledReporter.java
```diff
@@ -167,20 +172,10 @@ synchronized void start(long initialDelay, long period, TimeUnit unit, Runnable
...
-    /**
-     * Schedule the task, and return a future.
-     *
-     * @deprecated Use {@link #getScheduledFuture(long, long, TimeUnit, Runnable, ScheduledExecutorService)} instead.
-     */
-    @SuppressWarnings("DeprecatedIsStillUsed")
-    @Deprecated
-    protected ScheduledFuture<?> getScheduledFuture(long initialDelay, long period, TimeUnit unit, Runnable runnable) {
-        return getScheduledFuture(initialDelay, period, unit, runnable, this.executor);
-    }
```
github-louis-fruleux, line 183: I am aware this is a breaking change. Unfortunately, I did not find a way to make this function still relevant while passing the functional interface
Additionally, I think it is unlikely someone is overriding this method because it can make sense only with an executor.

What do you think?
Something obvious I missed in keeping this method for retro compatibility, while still making it relevant?

PR #4246, opened 2024-06-16
Add check in ScheduledReporter's stop method

### LINKED ISSUE #4135: ScheduledReporter reports metrics when stopped/closed even when not started
`ScheduledReporter`'s `close` method calls its `stop` method, which calls `report` to report the metrics "one last time". This also happens if `start` has never been called.

In a SpringBoot application that registers a `ScheduledReporter` as a bean, this can lead to strange behavior since the `stop` method is automatically called by Spring Boot.

ummels:
Check whether reporter is started before reporting metrics one last time.

Closes https://github.com/dropwizard/metrics/issues/4135


joschi (approved):
@ummels Thanks for your contribution! ❤️

PR #4272, opened 2024-07-04
instrument rejected tasks in ThreadPoolExecutor

gal-leib:
This adds an instrumentation for the number of rejected tasks in the ThreadPoolExecutor.


joschi (approved):
@gal-leib Thanks for your contribution! ❤️

gal-leib:
Happy to! Now we can upgrade the dependency and use the built in class 🙂

PR #4313, opened 2024-07-30
Graphite: implement connection & socket timeout

dolfinus:
`Graphite` class methods does not have any explicit timeout. So when Graphite instance is not available, experiencing network issues or is overloaded with too many requests, reporter may hang for some long time.

In my case (Spark metrics sink sending reports to Graphite) the lack of timeouts prevents Spark session from being stopped if Graphite instance is overloaded. This can last 5-10 minutes, sometimes even more.

What's changed:
* Add `connectionTimeoutMs` argument to `Graphite` constructor (default 500ms)
* Add `socketTimeoutMs` argument to `Graphite` constructor (default 5s)


dolfinus:
Closed because `socket.setSoTimeout(...)` can be used only to set timeout of read operations, but not write operations which are used by this sender. Java sockets don't support timeout for write operations at all. The only way to implement this is using some kind of non-blocking IO.

PR #4817, opened 2025-06-04
Implement MetricName#toString()/parse() with an InfluxDB-like format

### LINKED ISSUE #2167: MetricName well-defined toString + valueOf(String)
To simplify migration from the plain `dotted.name` syntax (in 4.x versions), usually represented as a Java String, the new tagged MetricName should have a well-defined string syntax. 

For example, [dropwizard-metrics-graphite](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-metrics-graphite/src/main/java/io/dropwizard/metrics/graphite/GraphiteReporterFactory.java) uses a String for its' `prefix` in the configuration.

Currently the MetricName have the toString() representation of `dotted.name{tag1=value, tag2=value}`, i.e. dottedName + Map.toString(). However, without escaping that representation is not parsable. 

As a comparison:
* InfluxDB use the syntax ([line protocol syntax](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/)): `weather,location=us-midwest`
* [Prometheus](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) use the syntax `bicycle_distance_meters_total{brand="monark",gears="7"}`

**Discussion:**

Should we stick to the existing string representation, or use e.g. the InfluxDB syntax, which is well-defined ([explicit escaping rules](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/#special-characters)), slimmer than Prometheus' and more liberate in allowed characters? The InfluxDB line protocol syntax is also backwards compatible when not using tags, e.g. `dotted.name` vs `dotted.name,with=tags`

**Update:** 

**Note:** A common "prefix" is to add tags for an application, e.g. `environment=production,region=eu-west` without the "dotted.name" part. 

What do you think? Should we go for a InfluxDB-inspired syntax?

joschi:
This change set has been created using GitHub Copilot as an experiment.

Closes #2167

### DIFF
--- a/metrics-core/src/main/java/io/dropwizard/metrics5/MetricName.java
+++ b/metrics-core/src/main/java/io/dropwizard/metrics5/MetricName.java
@@ -1,12 +1,11 @@
 package io.dropwizard.metrics5;
 
 import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -17,7 +16,7 @@ import java.util.stream.Stream;
 public class MetricName implements Comparable<MetricName> {
 
     private static final String SEPARATOR = ".";
-    private static final Map<String, String> EMPTY_TAGS = Collections.emptyMap();
+    private static final SortedMap<String, String> EMPTY_TAGS = Collections.emptySortedMap();
     static final MetricName EMPTY = new MetricName("", EMPTY_TAGS);
 
     /**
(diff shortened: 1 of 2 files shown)


joschi:
@mbrannstrom @baharclerode Do you want to weigh in on this?

mbrannstrom:
> @mbrannstrom Do you want to weigh in on this?

The primary purpose of #2167 was to come to an agreement on the desired behavior (in 2021). Once we knew WHAT to build, then the HOW would be trivial. I could have provided an implementation back in 2021...

@joschi What is the reason to revive this issue now? Do we have a planned release of Metric 5.0 soon?

joschi:
> What is the reason to revive this issue now? Do we have a planned release of Metric 5.0 soon?

We have the following options:
- Release Dropwizard Metrics 5.0.0 without this. Then changing the format will be (strictly speaking) a breaking change.
- Release Dropwizard Metrics 5.0.0 with this change for peace of mind.
- Scrap Dropwizard Metrics 5.0.0 altogether because the demand has frankly been quite low. On the other hand most of the work has already been done.

I wanted to play with GitHub Copilot's agent mode and this issue seemed innocent enough, so here we are. 😅

PR #4936, opened 2025-09-03
Optimize SlidingTimeWindowMovingAverages sumBuckets

schlosna:
`SlidingTimeWindowMovingAverages` `sumBuckets()` method could be optimized to perform indexed list access and remove allocations as it currently allocates a `LongAdder` as well as one or two streams. If one is using the 1, 5, or 15 minute rates on hot paths, these can be unnecessary expensive allocations as well as less optimized computations.

We can avoid the allocations completely and accumulate the sum directly to a `long` via optimized direct indexed list access.

[MovingAverageBenchmarks](https://github.com/palantir/tritium/blob/davids/OptimizedSlidingTimeWindowMovingAverages/tritium-jmh/src/jmh/java/com/palantir/tritium/microbenchmarks/MovingAverageBenchmarks.java) demonstrates the difference in implementations in both execution time (22x faster) and allocations:

```
Benchmark                                             (recordings)                                    (type)  Mode  Cnt     Score     Error   Units
MovingAverageBenchmarks.getM1Rate                               10           SlidingTimeWindowMovingAverages  avgt   20  1364.969 ±  12.040   ns/op
MovingAverageBenchmarks.getM1Rate:gc.alloc.rate.norm            10           SlidingTimeWindowMovingAverages  avgt   20   688.037 ±   0.001    B/op
MovingAverageBenchmarks.getM1Rate                               10  OptimizedSlidingTimeWindowMovingAverages  avgt   20    59.182 ±   0.343   ns/op
MovingAverageBenchmarks.getM1Rate:gc.alloc.rate.norm            10  OptimizedSlidingTimeWindowMovingAverages  avgt   20     0.002 ±   0.001    B/op
MovingAverageBenchmarks.getM1Rate                             1000           SlidingTimeWindowMovingAverages  avgt   20  1401.864 ± 107.134   ns/op
MovingAverageBenchmarks.getM1Rate:gc.alloc.rate.norm          1000           SlidingTimeWindowMovingAverages  avgt   20   688.038 ±   0.003    B/op
MovingAverageBenchmarks.getM1Rate                             1000  OptimizedSlidingTimeWindowMovingAverages  avgt   20    61.157 ±   1.393   ns/op
MovingAverageBenchmarks.getM1Rate:gc.alloc.rate.norm          1000  OptimizedSlidingTimeWindowMovingAverages  avgt   20     0.002 ±   0.001    B/op
```

### DIFF
--- a/metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowMovingAverages.java
+++ b/metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowMovingAverages.java
@@ -4,7 +4,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicLongArray;
+import java.util.concurrent.atomic.LongAdder;
 
 /**
  * A triple of simple moving average rates (one, five and fifteen minutes rates) as needed by {@link Meter}.
@@ -38,7 +38,7 @@ public class SlidingTimeWindowMovingAverages implements MovingAverages {
      * One counter per time bucket/slot (i.e. per second, see TICK_INTERVAL) for the entire
      * time window (i.e. 15 minutes, see TIME_WINDOW_DURATION_MINUTES)
      */
-    private final AtomicLongArray buckets;
+    private final LongAdder[] buckets;
 
     /**
      * Index into buckets, pointing at the bucket containing the oldest counts
@@ -78,7 +78,10 @@ public class SlidingTimeWindowMovingAverages implements MovingAverages {
         final long startTime = clock.getTick();
         lastTick = new AtomicLong(startTime);
 
-        buckets = new AtomicLongArray(NUMBER_OF_BUCKETS);
+        buckets = new LongAdder[NUMBER_OF_BUCKETS];
+        for (int i = 0; i < NUMBER_OF_BUCKETS; i++) {
+            buckets[i] = new LongAdder();
+        }
         bucketBaseTime = Instant.ofEpochSecond(0L, startTime);
         oldestBucketTime = bucketBaseTime;
         oldestBucketIndex = 0;
@@ -87,7 +90,7 @@ public class SlidingTimeWindowMovingAverages implements MovingAverages {
 
     @Override
     public void update(long n) {
-        buckets.addAndGet(currentBucketIndex, n);
+        buckets[currentBucketIndex].add(n);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowMovingAverages.java
```diff
@@ -39,7 +40,7 @@ public class SlidingTimeWindowMovingAverages implements MovingAverages {
      * One counter per time bucket/slot (i.e. per second, see TICK_INTERVAL) for the entire
      * time window (i.e. 15 minutes, see TIME_WINDOW_DURATION_MINUTES)
      */
-    private ArrayList<LongAdder> buckets;
+    private final List<LongAdder> buckets;
```
pkoenig10: This is a fixed size and could just be an array.

schlosna: good call, updated

metrics-core/src/main/java/com/codahale/metrics/SlidingTimeWindowMovingAverages.java
```diff
@@ -39,7 +38,7 @@ public class SlidingTimeWindowMovingAverages implements MovingAverages {
      * One counter per time bucket/slot (i.e. per second, see TICK_INTERVAL) for the entire
      * time window (i.e. 15 minutes, see TIME_WINDOW_DURATION_MINUTES)
      */
-    private ArrayList<LongAdder> buckets;
+    private final LongAdder[] buckets;
```
schlosna, line 41: a separate optimization would be to avoid the `LongAdder`s and use `AtomicLongArray`

```suggestion
    private final AtomicLongArray buckets;
```

schlosna: using `AtomicLongArray` cuts optimized `getM1Rate` from ~60ns to ~45ns, but there's additional overhead on the concurrent `update` side so going to hold off on that change.

```
Benchmark                          (recordings)                                    (type)  Mode  Cnt     Score    Error  Units
MovingAverageBenchmarks.getM1Rate            10           SlidingTimeWindowMovingAverages  avgt   20  1366.241 ± 24.341  ns/op
MovingAverageBenchmarks.getM1Rate            10  OptimizedSlidingTimeWindowMovingAverages  avgt   20    43.180 ±  0.153  ns/op
MovingAverageBenchmarks.getM1Rate          1000           SlidingTimeWindowMovingAverages  avgt   20  1377.545 ± 65.159  ns/op
MovingAverageBenchmarks.getM1Rate          1000  OptimizedSlidingTimeWindowMovingAverages  avgt   20    43.967 ±  2.058  ns/op
```

ash211:
Thanks for opening this PR @schlosna !

The reasoning about doing fewer allocations and more performant array indexing makes sense, and the benchmarks show a clear performance improvement. It does not seem like a reduction in code readability either. I also checked if LongAdder does anything special for long overflow and didn't see anything, so I believe this is behavior-equivalent.

Requesting review from maintainer @joschi

joschi (approved):
@schlosna Thank you for your contribution and providing benchmarks for your change!

PR #5101, opened 2026-01-14
Add a metrics-jackson3 module - metrics-json but dependent on jackson 3

### LINKED ISSUE #5100: Upgrade to Jackson 3?
What are the plans on upgrading metrics-json to jackson 3?

https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md

cunningt:
https://github.com/dropwizard/metrics/issues/5100

https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md

Add a metrics-jackson3 module (metrics-json but dependent on jackson 3 and with the API changes required by jackson 3 migration).       PR builds cleanly on my machine, the two tests in metrics-jackson3 pass.      

I see two untracked files after building (metrics-benchmarks/dependency-reduced-pom.xml, metrics-jcstress/dependency-reduced-pom.xml) but I don't see them checked into 4.2.x branch.    Should I add them?     Should dependency-reduced-pom.xml be added to .gitignore?

### DIFF
--- a/metrics-jackson3/pom.xml
+++ b/metrics-jackson3/pom.xml
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.dropwizard.metrics</groupId>
+        <artifactId>metrics-parent</artifactId>
+        <version>4.2.38-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>metrics-jackson3</artifactId>
+    <name>Jackson 3 Integration for Metrics</name>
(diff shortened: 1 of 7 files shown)


cunningt:
@joschi I think the JDK11 build is failing because the baseline JDK for Jackson 3 is 17 - should I add a profile here to enable/disable the module based on JDK building it?

metrics-jackson3/pom.xml
joschi, line 1: issue: Jackson 3.x is using JDK 17 as its baseline, so the module needs to be excluded from builds with lower JDK versions.

- https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0#jdk

Take a look at the metrics-jetty12 module for an example how to do this.

- https://github.com/dropwizard/metrics/blob/v4.2.37/metrics-jetty12/pom.xml#L22
- https://github.com/dropwizard/metrics/blob/v4.2.37/pom.xml#L149-L191

joschi:
> I think the JDK11 build is failing because the baseline JDK for Jackson 3 is 17 - should I add a profile here to enable/disable the module based on JDK building it?

@cunningt Yes, please. See my comment in the code. 😄

joschi (approved):

joschi:
@cunningt Thanks for your contribution!

